PR #1, opened 2015-08-04
Add 'systemd-tmpfiles-create'

roidelapluie:


raphink:
Looks good. If you're going to provide the tmpfiles exec though, I wouldn't mind a defined type to wrap around Augeas modifications, and even shipping the lens (since it's currently unreleased). Maybe an Augeas provider could be a good idea here.

roidelapluie:
You mean that we could ship a custom type with this module?

raphink:
We sure could. We do that in our openldap and varnish modules for example.

PR #8, opened 2016-07-25
Simplify the systemd version detection

bastelfreak:


petems:
@bastelfreak Looks like this has been fixed by https://github.com/camptocamp/puppet-systemd/pull/9 👍

PR #10, opened 2016-07-31
Add target param for the unit file

tampakrap:
This is useful in case the Unit file is a symlink to another one


raphink:
This seems OK. Could you rebase on master to check if the tests pass please?

tampakrap:
done

PR #12, opened 2016-08-23
Refactor systemd facts

petems:
* Use `Facter::Util::Resolution.exec`, which is backwards compatible with older Facter, calls the same method in modern facter
* Adds basic unit testing for facts

### DIFF
--- a/lib/facter/systemd.rb
+++ b/lib/facter/systemd.rb
@@ -23,13 +23,13 @@
 Facter.add(:systemd) do
   confine :kernel => :linux
   setcode do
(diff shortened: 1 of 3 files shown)


raphink:
Looks good, but the tests don't pass.

petems:
Bah, forgot to mock out the native call from the machine, didn't notice on my machine because it's a mac!

Fixing now

raphink:
OK great.

petems:
@raphink Ok to merge? 👍

raphink:
Yes, thank you.

PR #13, opened 2016-09-30
Manage resource limits of services

ruriky:
Hi,

Are you planning to improve the module further or is the idea to keep it simple and use it as a "global sync points for reloading systemd"?

Here is a PR for managing resource limits for services started by systemd.  
The module creates _servicename.service_.d/limits.conf file for storing the extra configuration for the service, refreshes systemd and restarts the service.  
Limits are configurable also by using hiera.


crayfishx:
@raphink Any chance of some feedback on this one?  We're wanting to use this patch internally.... Cheers

PR #14, opened 2016-10-11
Add Journald support

duritong:
Make it possible to manage journald settings

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -91,6 +91,7 @@ class systemd (
   Boolean                                                $manage_accounting,
   Hash[String,String]                                    $accounting,
   Boolean                                                $purge_dropin_dirs,
+  Hash[String,Variant[String,Hash[String, String]]]      $journald_settings,
 ){
 
   contain systemd::systemctl::daemon_reload
@@ -112,4 +113,6 @@ class systemd (
   if $manage_accounting {
     contain systemd::system
   }
+
+  contain systemd::journald
 }
--- a/manifests/journald.pp
+++ b/manifests/journald.pp
@@ -1,0 +1,31 @@
+# **NOTE: THIS IS A [PRIVATE](https://github.com/puppetlabs/puppetlabs-stdlib#assert_private) CLASS**
+#
+# This class provides a solution to enable accounting
+#
+# https://www.freedesktop.org/software/systemd/man/journald.conf.html
+class systemd::journald {
+
+  assert_private()
+
+  service{'systemd-journald':
+    ensure => running,
+  }
+  $systemd::journald_settings.each |$option, $value| {
+    ini_setting{
+      $option:
+        path    => '/etc/systemd/journald.conf',
+        section => 'Journal',
+        setting => $option,
+        notify  => Service['systemd-journald'],
+    }
+    if $value =~ Hash {
+      Ini_setting[$option]{
+        * => $value,
+      }
+    } else {
+      Ini_setting[$option]{
+        value   => $value,
+      }
+    }
+  }
+}
--- a/metadata.json
+++ b/metadata.json
@@ -11,6 +11,10 @@
     {
       "name": "puppetlabs/stdlib",
       "version_requirement": ">= 4.13.1 < 6.0.0"
+    },
+    {
+      "name": "puppetlabs/inifile",
+      "version_requirement": ">= 1.6.0 <3.0.0"
     }
   ],
   "requirements": [
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -9,6 +9,9 @@ describe 'systemd' do
         it { is_expected.to compile.with_all_deps }
         it { is_expected.to create_class('systemd') }
         it { is_expected.to create_class('systemd::systemctl::daemon_reload') }
+        it { is_expected.to contain_class('systemd::journald')}
+        it { is_expected.to create_service('systemd-journald') }
+        it { is_expected.to have_ini_setting_resource_count(0) }
         it { is_expected.to_not create_service('systemd-resolved') }
         it { is_expected.to_not create_service('systemd-networkd') }
         it { is_expected.to_not create_service('systemd-timesyncd') }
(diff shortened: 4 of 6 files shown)


manifests/init.pp
```diff
@@ -15,4 +17,20 @@
       command => 'systemd-tmpfiles --create',
   }
 
+  # https://www.freedesktop.org/software/systemd/man/journald.conf.html
+  if !empty($journald_settings) {
+    service{'systemd-journald':
+      ensure => running,
+    }
```
raphink: Do we only want to manage journald if settings are passed, or always?

duritong: I'm totally open, I just did an implementation that was the most non-intrusive.

raphink: I'd say we might aswell manage it if we're going to.

raphink: @duritong can you make the change to always manage it then?

duritong:
Ok, I pushed a rebased version that manages the journald service all the time.

thaiphv:
When can we get this merged?

trevor-vaughan:
I understand wanting to put all of the systemd subsystems under the `systemd` module, but the various subsystems really are probably complex enough, with enough non-linear changes, to warrant their own modules.

We forked https://github.com/cristifalcas/puppet-journald over at https://github.com/simp/pupmod-simp-journald to continue on with it, get it updated, and add our tests so I would recommend not adding `journald` support to the main `systemd` module.

ekohl:
Given the size of those modules I wonder if a ```system::journald``` class would also work.

trevor-vaughan:
@ekohl The issue isn't the size, it's the fact that the API for journald and systemd settings change independently and a change to either would cause a (potentially breaking) change to the API. You would basically end up with two forks running under this project to keep track of minute changes in either subsystem.

ekohl:
@trevor-vaughan I'm not that familiar with it so I'll let you judge it then.

raphink:
@mcanevet what do you think? Should we manage journald (and all its friends eventually) or let other modules do that (like @trevor-vaughan suggested)?

locknut:
@raphink Our team is okay with (and prefers) one module for both systemd and journald,

locknut:
We've been running @duritong 's journald patch for quite a while. Seeing a strange issue on CentOS 7.4 - when puppet refreshes the systemd-journald service, the agent run silently dies.

trevor-vaughan:
@raphink and @mcanevet I would like to point out that both `systemd` and `journald` are different subsystems and that an update to the API of one may introduce breaking changes in the module that do not affect the other. As such, I would like to petition again for keeping the two subsystems as separate modules.

raphink:
@mcanevet what's your take on this?

bastelfreak:
Personally I would be fine to have this PR in this module. Journald is, besides timesyncd, resolved and the actual init systemd, part of the systemd project. I see this module as something that managed the project, not only the init system.

@trevor-vaughan systemd itself (the init system) and systemd-journald have different options  In a well designed code, breaking changes in one of those components shouldn't effect the other components, so I don't see an issue here. The module may need a major release to support those changes, but that should only affect this specific component (e.g. journald) and not the rest.

bastelfreak:
This module already has support for managing networkd, resolved, timesyncd, so I think managing journald is totally fine as long as the default behavior is to not manage it. @duritong are you still interested in this PR? Can you rebase it?

@trevor-vaughan would you be okay with this?

manifests/init.pp
```diff
@@ -1,7 +1,8 @@
 # -- Class systemd  
 # This module allows triggering systemd commands once for all modules 
 class systemd (
-  $service_limits = {}
+  $service_limits    = {},
```
bastelfreak: Those parameters need datatypes

spec/classes/init_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'spec_helper'
+
+describe 'systemd' do
+
+  let(:facts) { {
```
bastelfreak: could you use rspec-puppet-facts instead?

duritong:
Sure, if it gives the chance to get it merged...

manifests/journald.pp
```diff
@@ -0,0 +1,31 @@
...
+class systemd::journald {
+
+  assert_private()
+
+  service{'systemd-journald':
+    ensure => running,
+  }
+  $systemd::journald_settings.each |$option, $value| {
+    ini_setting{
+      $option:
```
bastelfreak: interesting format, I haven't seen it this way in the past.

metadata.json
```diff
@@ -11,6 +11,10 @@
     {
       "name": "puppetlabs/stdlib",
       "version_requirement": ">= 4.13.1 < 5.0.0"
+    },
+    {
+      "name": "puppetlabs/inifile",
+      "version_requirement": ">= 1.6.0 <2.0.0"
```
bastelfreak: can you bump the version requirement to < 3.0.0?  2.2.0 is the latest release.

bastelfreak:
Hi @duritong, thanks for rebasing it. I found one little issue that needs to be fixed. I'm going to merge this tomorrow unless somebody raises concerns.

elfchief:
Any chance of actually getting this merged?

trevor-vaughan:
@bastelfreak Arg...sorry that I missed this until now.

You said previously:

> The module may need a major release to support those changes, but that should only affect this specific component (e.g. journald) and not the rest.

But this is exactly my issue. In this case, I can't reliably pin on an upper bound because features in two different subsystems can change independently and in potentially breaking changes. This increases the likelihood of forks if someone needs the fixes to the `systemd` part but can't use the breaking changes in the `journald` part.

Note that several of the major OS vendors take in subsystem patches at different rates so there is a relatively good chance that the module will end up in this scenario.

It's irritating, but `systemd` isn't a single thing, it's a big pluggable system that needs to be treated as such.

bastelfreak:
@trevor-vaughan so we are testing against a fixed list operating systems, and it is pretty unlikely that for example RHEL 7.6 gets an update to 7.7 that pulls in a breaking change within the journald config. We do not even care about breaking changes within journald, right? As long as it still eats the config we provide. We have support for timesyncd/resolved and it is pretty generic and opt in. I think this pattern would be fine for journald as well. And a breaking change within this module for either the init part or any of the other components would result in another major release, so pinning to an upper version boundary works. Overall I think this would be fine?

----

@elfchief it cannot be merged because it has merge conflicts (and there is still a little debate going on). If you are interested in this PR, you can checkout the feature branch, rebase it, add a switch to enable/disable the management, like we do it here:
https://github.com/camptocamp/puppet-systemd/blob/10cdeb617478262caf05245f131652fa823fef22/manifests/init.pp#L83

Afterwards we can review it again.

trevor-vaughan:
@bastelfreak I have no  concerns about EL7 updates but I do worry about the same module trying to take care of EL and Ubuntu since they tend to pull in different changes over time.

I suppose that it can always be split in the future if necessary.

duritong:
I rebased it once again.

raphink:
@duritong 	I'm sorry, I merged https://github.com/camptocamp/puppet-systemd/commit/db56678399126502ecfc9ddea7b1829e2607503b 2 days ago and it created a conflict again :cry:

duritong:
So I rebased again.

PR #16, opened 2016-11-21
Create simplified unit file defined type

blajos:
Hi,

We needed a way to include multiple unit files from one of our profile modules and we thought this was the right way of doing it.

What do you think?

Thanks,
Lajos


raphink:
I can see how this is useful to you, but I'm not sure this kind of wrapper really belongs in the module.

blajos:
We'll move the wrapper to another place then.

Thanks for your feedback.

PR #18, opened 2017-01-16
Update to support Puppet 4

### LINKED ISSUE #17: Module should be updated to use the new Puppet 4 goodness
I'll be submitting a PR as soon as I update the tests

trevor-vaughan:
* Optimized the limits files
* Added a custom data type for service limits
* Created a class for restarting the daemon so that other modules can
  use it cleanly
* Updated tests

Closes #17

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,8 @@ file { '/usr/lib/systemd/system/foo.service':
   group  => 'root',
   mode   => '0644',
   source => "puppet:///modules/${module_name}/foo.service",
-} ~>
-Class['systemd::systemctl::daemon_reload']
+}
+~> Class['systemd::systemctl::daemon_reload']
 ```
 
 ### tmpfiles
@@ -57,8 +57,8 @@ file { '/etc/tmpfiles.d/foo.conf':
   group  => 'root',
   mode   => '0644',
   source => "puppet:///modules/${module_name}/foo.conf",
-} ~>
-Class['systemd::tmpfiles']
+}
+~> Class['systemd::tmpfiles']
 ```
 
 ### service limits
@@ -69,7 +69,7 @@ Manage soft and hard limits on various resources for executed processes.
 ::systemd::service_limits { 'foo.service':
   limits => {
     LimitNOFILE => 8192,
-    LimitNPROC  => 16384
+    LimitNPROC  => 16384,
   }
 }
 ```
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ Let this module handle file creation and systemd reloading.
 Or handle file creation yourself and trigger systemd.
 
 ```puppet
-include ::systemd
+include ::systemd::systemctl::daemon_reload
+
 file { '/usr/lib/systemd/system/foo.service':
   ensure => file,
   owner  => 'root',
@@ -32,7 +33,7 @@ file { '/usr/lib/systemd/system/foo.service':
   mode   => '0644',
   source => "puppet:///modules/${module_name}/foo.service",
 } ~>
-Exec['systemctl-daemon-reload']
+Class['systemd::systemctl::daemon_reload']
 ```
 
 ### tmpfiles
@@ -48,7 +49,8 @@ Let this module handle file creation and systemd reloading
 Or handle file creation yourself and trigger systemd.
 
 ```puppet
-include ::systemd
+include ::systemd::tmpfiles
+
 file { '/etc/tmpfiles.d/foo.conf':
   ensure => file,
   owner  => 'root',
@@ -56,7 +58,7 @@ file { '/etc/tmpfiles.d/foo.conf':
   mode   => '0644',
   source => "puppet:///modules/${module_name}/foo.conf",
 } ~>
-Exec['systemd-tmpfiles-create']
+Class['systemd::tmpfiles']
 ```
 
 ### service limits
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -19,11 +19,11 @@
 #   The state that the ``networkd`` service should be in
 #
 class systemd (
-  Optional[Hash]                             $service_limits  = undef,
-  Boolean                                    $manage_resolved = false,
-  Variant[Enum['stopped','running'],Boolean] $resolved_ensure  = 'running',
-  Boolean                                    $manage_networkd = false,
-  Variant[Enum['stopped','running'],Boolean] $networkd_ensure = 'running',
+  Optional[Hash]            $service_limits  = undef,
+  Boolean                   $manage_resolved = false,
+  Enum['stopped','running'] $resolved_ensure = 'running',
+  Boolean                   $manage_networkd = false,
+  Enum['stopped','running'] $networkd_ensure = 'running',
 ){
 
   contain ::systemd::systemctl::daemon_reload
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -1,27 +1,23 @@
-# -- Class systemd
 # This module allows triggering systemd commands once for all modules
+#
+# @api public
+#
+# @param service_limits
+#   May be passed a resource hash suitable for passing directly into the
+#   ``create_resources()`` function as called on ``systemd::service_limits``
+#
 class systemd (
-  $service_limits          = {},
+  Optional[Hash] $service_limits = undef
   Boolean $manage_resolved = true,
   Boolean $manage_networkd = true,
 ){
 
-  Exec {
-    refreshonly => true,
-    path        => $::path,
-  }
+  include '::systemd::systemctl::daemon_reload'
+  include '::systemd::tmpfiles'
 
-  exec {
-    'systemctl-daemon-reload':
-      command => 'systemctl daemon-reload',
+  if $service_limits {
+    create_resources('systemd::service_limits', $service_limits)
   }
-
-  exec {
-    'systemd-tmpfiles-create':
-      command => 'systemd-tmpfiles --create',
-  }
-
-  create_resources('systemd::service_limits', $service_limits, {})
 
   if $manage_resolved {
     service{'systemd-resolved':
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -19,11 +19,11 @@
 #   The state that the ``networkd`` service should be in
 #
 class systemd (
-  Optional[Hash]            $service_limits  = undef,
-  Boolean                   $manage_resolved = false,
-  Enum['stopped','running'] $resolved_ensure = 'running',
-  Boolean                   $manage_networkd = false,
-  Enum['stopped','running'] $networkd_ensure = 'running',
+  Optional[Systemd::ServiceLimits] $service_limits  = undef,
+  Boolean                          $manage_resolved = false,
+  Enum['stopped','running']        $resolved_ensure = 'running',
+  Boolean                          $manage_networkd = false,
+  Enum['stopped','running']        $networkd_ensure = 'running',
 ){
 
   contain ::systemd::systemctl::daemon_reload
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -12,8 +12,7 @@ class systemd (
   Boolean $manage_networkd = true,
 ){
 
-  include '::systemd::systemctl::daemon_reload'
-  include '::systemd::tmpfiles'
+  contain systemd::systemctl::daemon_reload
 
   if $service_limits {
     create_resources('systemd::service_limits', $service_limits)
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -12,7 +12,7 @@ class systemd (
   Boolean $manage_networkd = true,
 ){
 
-  contain systemd::systemctl::daemon_reload
+  contain ::systemd::systemctl::daemon_reload
 
   if $service_limits {
     create_resources('systemd::service_limits', $service_limits)
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -20,9 +20,9 @@
 #
 class systemd (
   Optional[Hash]                             $service_limits  = undef,
-  Boolean                                    $manage_resolved = true,
+  Boolean                                    $manage_resolved = false,
   Variant[Enum['stopped','running'],Boolean] $resolved_ensure  = 'running',
-  Boolean                                    $manage_networkd = true,
+  Boolean                                    $manage_networkd = false,
   Variant[Enum['stopped','running'],Boolean] $networkd_ensure = 'running',
 ){
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -9,8 +9,8 @@
 # @param manage_resolved
 #   Manage the systemd resolver
 #
-# @param resolvd_ensure
-#   The state that the ``resolvd`` service should be in
+# @param resolved_ensure
+#   The state that the ``resolved`` service should be in
 #
 # @param manage_networkd
 #   Manage the systemd network daemon
@@ -21,7 +21,7 @@
 class systemd (
   Optional[Hash]                             $service_limits  = undef,
   Boolean                                    $manage_resolved = true,
-  Variant[Enum['stopped','running'],Boolean] $resolvd_ensure  = 'running',
+  Variant[Enum['stopped','running'],Boolean] $resolved_ensure  = 'running',
   Boolean                                    $manage_networkd = true,
   Variant[Enum['stopped','running'],Boolean] $networkd_ensure = 'running',
 ){
@@ -33,32 +33,10 @@ class systemd (
   }
 
   if $manage_resolved {
-    $_enable_resolvd = $resolvd_ensure ? {
-      /stopped/ => false,
-      /running/ => true,
-      default   => $resolvd_ensure
-    }
-
-    service{ 'systemd-resolved':
-      ensure => $resolvd_ensure,
-      enable => $_enable_resolvd
-    }
-    -> file{ '/etc/resolv.conf':
-      ensure => 'symlink',
-      target => '/run/systemd/resolve/resolv.conf'
-    }
+    contain ::systemd::resolved
   }
 
   if $manage_networkd {
-    $_enable_networkd = $networkd_ensure ? {
-      /stopped/ => false,
-      /running/ => true,
-      default   => $networkd_ensure
-    }
-
-    service{ 'systemd-networkd':
-      ensure => $networkd_ensure,
-      enable => $_enable_networkd
-    }
+    contain ::systemd::networkd
   }
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -6,10 +6,24 @@
 #   May be passed a resource hash suitable for passing directly into the
 #   ``create_resources()`` function as called on ``systemd::service_limits``
 #
+# @param manage_resolved
+#   Manage the systemd resolver
+#
+# @param resolvd_ensure
+#   The state that the ``resolvd`` service should be in
+#
+# @param manage_networkd
+#   Manage the systemd network daemon
+#
+# @param networkd_ensure
+#   The state that the ``networkd`` service should be in
+#
 class systemd (
-  Optional[Hash] $service_limits = undef
-  Boolean $manage_resolved = true,
-  Boolean $manage_networkd = true,
+  Optional[Hash]                             $service_limits  = undef,
+  Boolean                                    $manage_resolved = true,
+  Variant[Enum['stopped','running'],Boolean] $resolvd_ensure  = 'running',
+  Boolean                                    $manage_networkd = true,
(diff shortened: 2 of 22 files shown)


zbikmarc:
Any chance for merging it?

Gemfile
```diff
@@ -25,7 +25,8 @@ group :development, :unit_tests do
 end
 
 group :system_tests do
-  gem 'beaker',               :require => false
+  gem 'beaker', :git => 'https://github.com/trevor-vaughan/beaker.git', :branch => 'BKR-978-2.51.0'
```
raphink: We're managing this with modulesync, so we might need to use this in all of our modules...

raphink:
That's quite a PR. I'll have to take time to review it.

oranenj:
I took a look, and did not find anything objectionable.

Perhaps define types for the patterns used for limits? those seem like they might be generally useful when dealing with systemd. Also, don't some values accept 'infinity' as well?

Something like Systemd::Interval and Systemd::Size?

raphink:
That LGTM too, I'll just have to resolve that modulesync bit. @trevor-vaughan could you rebase please?

Gemfile
trevor-vaughan: @raphink This is now fixed in Beaker. I'll put in the correct updates.

trevor-vaughan:
@oranenj You are correct about the 'infinity' option. I'll double check those. The systemd documentation is...fragmented at best.

I hope to have this done by the end of the week.

Gemfile
raphink: Great!

trevor-vaughan:
@oranenj and @raphink Sorry that this took so long, the actual items supported by the configuration were a bit more involved than expected but it should all be properly updated now.

README.md
```diff
@@ -24,15 +24,16 @@ Let this module handle file creation and systemd reloading.
 Or handle file creation yourself and trigger systemd.
 
 ```puppet
-include ::systemd
+include ::systemd::systemctl::daemon_reload
```
raphink: I think I'd rather keep the interface in the `systemd` class and contain the subclasses (`systemd::systemctl::daemon_reload` and `systemd::tmpfiles`) for dependencies instead. Is there something that prevents from doing that?

manifests/init.pp
```diff
@@ -1,24 +1,19 @@
...
+#
+# @api public
+#
+# @param service_limits
+#   May be passed a resource hash suitable for passing directly into the
+#   ``create_resources()`` function as called on ``systemd::service_limits``
+#
 class systemd (
-  $service_limits = {}
+  Optional[Hash] $service_limits = undef
```
raphink: Why not make the default `{}` and check for the size of the hash? I find this a bit more logical than allowing for `undef` as a default value.

manifests/service_limits.pp
```diff
@@ -1,51 +1,87 @@
...
 
-  if $limits {
-    validate_hash($limits)
-    $content = template('systemd/limits.erb')
+  if $title !~ Pattern['^.+\.(service|socket|mount|swap)$'] {
+    fail('$name must match Pattern["^.+\.(service|socket|mount|swap)$"]')
+  }
+
+  if $limits and !empty($limits) {
+    $_content = template("${module_name}/limits.erb")
```
raphink: I don't see a `$content` variable. What is the need for `$_content`?

manifests/service_limits.pp
```diff
@@ -1,51 +1,87 @@
...
-  }
+  ensure_resource('file', "${path}/${title}.d/",
+    {
+      ensure => 'directory',
+      owner  => 'root',
+      group  => 'root',
+      mode   => '0644',
+      seltype => 'systemd_unit_file_t',
+    }
+  )
```
raphink: I'm not a big fan of using `ensure_resource`. I'd rather put this in a separate class and include it.

manifests/service_limits.pp
```diff
@@ -1,51 +1,87 @@
...
+    content => $_content,
     source  => $source,
     owner   => 'root',
     group   => 'root',
-    mode    => '0444',
-    notify  => Exec['systemctl-daemon-reload'],
+    mode    => '0644',
   }
 
+  File["${path}/${title}.d/90-limits.conf"] ~> Class['systemd::systemctl::daemon_reload']
```
raphink: Why not specify this directly above in the file declaration?

manifests/service_limits.pp
```diff
@@ -1,51 +1,87 @@
...
+    exec { "restart ${title} because limits":
+      command     => "systemctl restart ${title}",
       path        => $::path,
       refreshonly => true,
-      subscribe   => File["${path}/${title}.d/limits.conf"],
-      require     => Exec['systemctl-daemon-reload'],
     }
+
+    File["${path}/${title}.d/90-limits.conf"] ~> Exec["restart ${title} because limits"]
+    Class['systemd::systemctl::daemon_reload'] -> Exec["restart ${title} because limits"]
```
raphink: This could also be specified in the exec declaration.

manifests/tmpfile.pp
```diff
@@ -1,20 +1,49 @@
...
+
   file { "${path}/${title}":
     ensure  => $ensure,
     content => $content,
     source  => $source,
     owner   => 'root',
     group   => 'root',
-    mode    => '0444',
-    notify  => Exec['systemd-tmpfiles-create'],
+    mode    => '0644',
```
raphink: I'm not sure why this was 0444 and why this needs to be 0644 now.

manifests/unit_file.pp
```diff
@@ -1,22 +1,60 @@
...
+  Enum['file', 'absent']         $ensure  = 'file',
+  Stdlib::Absolutepath           $path    = '/etc/systemd/system',
+  Optional[String]               $content = undef,
+  Optional[String]               $source  = undef,
+  Optional[Stdlib::Absolutepath] $target  = undef,
 ) {
   include ::systemd
 
+  if $title !~ Pattern['^.+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope)$'] {
+    fail('$name must match Pattern["^.+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope)$"]')
```
raphink: Or maybe ship a data type for this and use `assert_type`?

manifests/unit_file.pp
```diff
@@ -1,22 +1,60 @@
...
+  }
+
+  if $target {
+    $_ensure = 'link'
+  }
+  else {
```
raphink: style nitpick: I'd rather not have a NL here

manifests/unit_file.pp
```diff
@@ -1,22 +1,60 @@
...
-    ensure  => $ensure,
+    ensure  => $_ensure,
     content => $content,
     source  => $source,
     target  => $target,
     owner   => 'root',
     group   => 'root',
-    mode    => '0444',
-    notify  => Exec['systemctl-daemon-reload'],
+    mode    => '0644',
```
raphink: Why change the mode?

README.md
trevor-vaughan: We can contain `daemon_reload` but we can't contain `tmpfiles` since tmpfiles is triggered from the use of the `tmpfile` defined type.

A notify to `Class[systemd]` makes sense for reloading the whole thing though so I'll switch that one to a contain and hope that we don't get more creative in the future.

manifests/init.pp
trevor-vaughan: I'd like to leave this one if possible. The compiler is so bloated already that everything we can bind to a single symbol helps.

Also, the code is easier to read and not confusing if, for some reason, we decide to include data in modules later.

manifests/service_limits.pp
trevor-vaughan: It denotes it as a private variable that should not be used outside of the class. It's been convention for a while but hasn't made it into the style guide yet.

When actual private variables become a thing, this will be easier.

manifests/service_limits.pp
trevor-vaughan: I can't. The name is dynamic based on the `path` and `title` parameters. It must be created for each call to the define but there can be multiple calls to the define that target the same location.

If we break it out into another define, we'll still have to use `ensure_resource`

manifests/service_limits.pp
trevor-vaughan: I find it easier to read, but I've switched it.

manifests/service_limits.pp
trevor-vaughan: Done

manifests/tmpfile.pp
trevor-vaughan: Reverted

manifests/unit_file.pp
trevor-vaughan: OK

manifests/unit_file.pp
trevor-vaughan: It makes my C soul hurt, but OK.

manifests/unit_file.pp
trevor-vaughan: Reverted

trevor-vaughan:
@raphink Code updated per the discussion above

manifests/init.pp
bastelfreak: I'm :+1: for this. We use the same style in the Vox Pupuli modules.

trevor-vaughan:
@bastelfreak Would you mind looking this over one last time. I made a few additional changes and would like some feedback.

I've pinged @raphink to see if he has any objections to the latest batch of updates and will push this tomorrow evening if I haven't heard anything by then. We can then rebase and work on #32

manifests/init.pp
```diff
@@ -1,43 +1,42 @@
...
+#
 class systemd (
-  $service_limits          = {},
-  Boolean $manage_resolved = true,
-  Boolean $manage_networkd = true,
+  Optional[Hash]                             $service_limits  = undef,
+  Boolean                                    $manage_resolved = false,
+  Variant[Enum['stopped','running'],Boolean] $resolved_ensure  = 'running',
+  Boolean                                    $manage_networkd = false,
+  Variant[Enum['stopped','running'],Boolean] $networkd_ensure = 'running',
```
bastelfreak: on the vox pupuli modules we agreed to only allow 'stopped' and 'running'. we dropped the boolean during the migration from puppet3 to puppet4/5.

manifests/unit_file.pp
```diff
@@ -1,22 +1,57 @@
...
+#
+#   * Mutually exclusive with both ``$source`` and ``$content``
+#
 define systemd::unit_file(
-  $ensure = file,
-  $path = '/etc/systemd/system',
-  $content = undef,
-  $source = undef,
-  $target = undef,
+  Enum['file', 'absent']         $ensure  = 'file',
```
bastelfreak: compared to the service ensure above, we don't allow true/false here, which is good in my opinion.

README.md
```diff
@@ -24,15 +24,16 @@ Let this module handle file creation and systemd reloading.
...
+
 file { '/usr/lib/systemd/system/foo.service':
   ensure => file,
   owner  => 'root',
   group  => 'root',
   mode   => '0644',
   source => "puppet:///modules/${module_name}/foo.service",
 } ~>
-Exec['systemctl-daemon-reload']
+Class['systemd::systemctl::daemon_reload']
```
bastelfreak: the ~> needs to be in front of Class[]

README.md
```diff
@@ -48,15 +49,16 @@ Let this module handle file creation and systemd reloading
...
+
 file { '/etc/tmpfiles.d/foo.conf':
   ensure => file,
   owner  => 'root',
   group  => 'root',
   mode   => '0644',
   source => "puppet:///modules/${module_name}/foo.conf",
 } ~>
-Exec['systemd-tmpfiles-create']
+Class['systemd::tmpfiles']
```
bastelfreak: the ~> needs to be in front of Class[]

bastelfreak:
@trevor-vaughan done

manifests/init.pp
trevor-vaughan: Fixed

README.md
trevor-vaughan: I honestly like to break these out explicitly, but I've been trying to follow the class.

README.md
trevor-vaughan: Fixed these and the one missing trailing comma (which I also don't agree with) :-D

trevor-vaughan:
@bastelfreak Mischief Managed. @raphink indicated that he could review this week, but not this weekend. Will pester mercilessly until we get things through ;-).

README.md
raphink: I see you contained the class in `init.pp`, so I'm guessing the README is not up-to-date?

manifests/init.pp
```diff
@@ -1,43 +1,42 @@
...
+#   Manage the systemd network daemon
+#
+# @param networkd_ensure
+#   The state that the ``networkd`` service should be in
+#
 class systemd (
-  $service_limits          = {},
-  Boolean $manage_resolved = true,
-  Boolean $manage_networkd = true,
+  Optional[Hash]            $service_limits  = undef,
```
raphink: We might want to create a Data Type for this eventually to have finer validation.

manifests/networkd.pp
```diff
@@ -0,0 +1,24 @@
+# **NOTE: THIS IS A [PRIVATE](https://github.com/puppetlabs/puppetlabs-stdlib#assert_private) CLASS**
+#
+# This class provides an abstract way to trigger resolved
+#
+# @param ensure
+#   The state that the ``networkd`` service should be in
+#
+class systemd::networkd (
+  Variant[Enum['stopped','running'],Boolean] $ensure = 'running',
```
raphink: If it's a private class and the values are limited to `Enum['stopped','running']` in `init.pp`, I'm not sure how it could ever be a Boolean?

manifests/resolved.pp
```diff
@@ -0,0 +1,30 @@
+# **NOTE: THIS IS A [PRIVATE](https://github.com/puppetlabs/puppetlabs-stdlib#assert_private) CLASS**
+#
+# This class provides an abstract way to trigger resolved
+#
+# @param ensure
+#   The state that the ``resolved`` service should be in
+#
+class systemd::resolved (
+  Variant[Enum['stopped','running'],Boolean] $ensure  = $::systemd::resolved_ensure,
```
raphink: Same Enum/Boolean comment here

manifests/networkd.pp
```diff
@@ -0,0 +1,24 @@
...
+){
+
+  assert_private()
+
+  $_enable_networkd = $ensure ? {
+    /stopped/ => false,
```
raphink: Why use regexp here?

manifests/resolved.pp
```diff
@@ -0,0 +1,30 @@
...
+){
+
+  assert_private()
+
+  $_enable_resolved = $ensure ? {
+    /stopped/ => false,
```
raphink: and regexp here

manifests/service_limits.pp
```diff
@@ -1,51 +1,86 @@
...
+# @param restart_service
+#   Restart the managed service after setting the limits
+#
 define systemd::service_limits(
-  $ensure          = file,
-  $path            = '/etc/systemd/system',
-  $limits          = undef,
-  $source          = undef,
-  $restart_service = true
+  Enum['file','absent']            $ensure          = 'file',
```
raphink: While I agree files shouldn't be set to `ensure => present`, I'd think `ensure => present` would make more sense for a higher absraction such as `systemd::service_limits`.

manifests/service_limits.pp
raphink: OK, fine for me.

manifests/service_limits.pp
raphink: OK, but then that means that file is unique, so what's the point of using `ensure_resource`?

manifests/tmpfile.pp
```diff
@@ -1,12 +1,41 @@
...
+#   A ``File`` resource compatible ``source``
+#
+#  * Mutually exclusive with ``$limits``
+#
 define systemd::tmpfile(
-  $ensure = file,
-  $path = '/etc/tmpfiles.d',
-  $content = undef,
-  $source = undef,
+  Enum['file','absent'] $ensure  = 'file',
```
raphink: same comment on file/present

manifests/unit_file.pp
raphink: same comment on file/present though

manifests/unit_file.pp
```diff
@@ -1,22 +1,57 @@
...
+  Stdlib::Absolutepath           $path    = '/etc/systemd/system',
+  Optional[String]               $content = undef,
+  Optional[String]               $source  = undef,
+  Optional[Stdlib::Absolutepath] $target  = undef,
 ) {
   include ::systemd
 
+  assert_type(Systemd::Unit, $title)
```
raphink: I would rather assert `$name` and use a `path` attribute in the file below, to allow users to choose their resource title as they want.

manifests/init.pp
trevor-vaughan: @raphink To the first comment: No, this is correct. You can do either of those cases in a valid manner.

`systemd::unit_file`` includes `systemd` so that it is independently feature complete and since this is the 90% use case of this module.

trevor-vaughan: @raphink To the second comment....I already made one of those :-|

trevor-vaughan:
@raphink image

raphink:
OK for me. @trevor-vaughan @bastelfreak I'll let you squash and merge.

types/servicelimits.pp
```diff
@@ -0,0 +1,21 @@
...
+    Optional['LimitNOFILE']     => Integer[0],
+    Optional['LimitAS']         => Pattern['^(infinity|((\d+(K|M|G|T|P|E)(:\d+(K|M|G|T|P|E))?)))$'],
+    Optional['LimitNPROC']      => Integer[1],
+    Optional['LimitMEMLOCK']    => Pattern['^(infinity|((\d+(K|M|G|T|P|E)(:\d+(K|M|G|T|P|E))?)))$'],
+    Optional['LimitLOCKS']      => Integer[1],
+    Optional['LimitSIGPENDING'] => Integer[1],
+    Optional['LimitMSGQUEUE']   => Pattern['^(infinity|((\d+(K|M|G|T|P|E)(:\d+(K|M|G|T|P|E))?)))$'],
+    Optional['LimitNICE']       => Variant[Integer[0,40], Pattern['^(-\+([0-1]?[0-9]|20))|([0-3]?[0-9]|40)$']],
+    Optional['LimitRTPRIO']     => Integer[0],
+    Optional['LimitRTTIME']     => Pattern['^\d+(ms|s|m|h|d|w|M|y)?(:\d+(ms|s|m|h|d|w|M|y)?)?$']
```
mwhahaha: FYI this is MemoryLimit so this broke us

mwhahaha: This is missing...

mwhahaha: See #23 for context

trevor-vaughan: @mwhahaha It would be lovely if the systemd man pages weren't a rats nest of crazy.

Would it make sense to add all of the things in this section to the `Systemd::ServiceLimits` type?

http://www.dsm.fordham.edu/cgi-bin/man-cgi.pl?topic=systemd.resource-control&ampsect=5

mwhahaha: Yes if you're going to validate them, then you need to support all of them. IMHO i wouldn't validate the keys/values because it'll be a constant battle to keep the module to supporting all the things added to systemd over time.  It would be better to do simple validation to just ensure that the key is a valid string (no special chars) and just accept whatever values.

trevor-vaughan: @mwhahaha At that point, you might as well just drop your own file in place. That's all this does anyway after some validation. Really, the only thing it adds is validation.

mwhahaha: Well we were using this to trigger the required restarts and systemd actions. So dropping the file is not the only thing that is needed.

mwhahaha: For reference, https://github.com/openstack/puppet-tripleo/blob/8ca7a1e390ce50ac66a6861dae59bb44fbf0324a/manifests/profile/base/database/mongodb.pp#L64-L69

I was just saying that trying to validate via puppet the underlying (and changing) systemd config file requirements in this puppet module seems to add more work than it's worth.

trevor-vaughan: @mwhahaha Quite possibly, yes. We'll see as time goes on. In theory, as people automate more around `systemd`, they'll need to get their act together and stop changing everything regularly.

Then again, when you want `systemd` to *be* the automation tool, I guess that can get difficult.

PR #19, opened 2017-02-06
Specify settings for a new unit file

traylenator:
Adds a new parameter `settings` to the
systemd::unit_file type.

```puppet
::systemd::unit_file{ 'foo.service':
   settings => { 'Description'     => 'the foo service',
                 'ExecStart'       => '/bin/foo',
                 'Wants'           => ['network.target', 'bar.service'],
                 'Type'            => 'oneshot',
                 'EnvironmentFile' => ['-/etc/sysconfig/foo','/etc/sysconfig/myfoo'],
               }
}
```

Allows a systemd file to be specified with a setting hash
rather than a source file.

### DIFF
--- a/templates/unit_file.erb
+++ b/templates/unit_file.erb
@@ -1,0 +1,85 @@
+#puppet managed file.
+
+[Unit]
+<%
+[
+  'Description',
+  'Requires',
+  'Conflicts',
+  'Wants',
+  'After',
+].each do |s| 
+if @settings[s]
+  if @settings[s].kind_of?(Array)
+    @settings[s].each do |v| 
+-%>
+<%= s %>=<%= v %>
+<%
+    end
+  elsif @settings[s].kind_of?(String) 
+-%>
+<%= s %>=<%= @settings[s] %>
+<%
+  else
+   raise Puppet::Error, "settings hash contains a key '#{s}' whose value which is not an array or a string"
+  end
+end
+end
+-%> 
+
+[Service]
+<%
+[
+  'Type',
+  'ExecStart',
+  'ExecStop',
+  'ExecReload',
+  'Environment',
+  'EnvironmentFile',
+  'WorkingDirectory',
+  'Restart',
+  'RestartSec',
+  'RestartAfterExec',
+].each do |s| 
+if @settings[s]
+  if @settings[s].kind_of?(Array)
+    @settings[s].each do |v| 
+-%>
+<%= s %>=<%= v %>
+<%
+    end
+  elsif @settings[s].kind_of?(String) 
+-%>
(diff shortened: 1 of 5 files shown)


antaflos:
To me this looks like a very useful addition, but it seems to be limited to only a handful of settings. Wouldn't it be more practical to have the `settings` hash look something like this:

```puppet
settings => {
  'Unit' => {
    'Description'           => 'the foo service',
    'StartLimitBurst'       => '3',
    'StartLimitIntervalSec' => '120',
  },
  'Service' => {
    'Type'            => 'oneshot',
    'ExecStart'       => '/bin/foo',
    'EnvironmentFile' => ['-/etc/sysconfig/foo','/etc/sysconfig/myfoo'],
  },
  'Install' => {
    'WantedBy' => 'multi-user.target',
  },
}
```

The implementation in the ERB template could basically look like this (without any type validation):

```erb
<% @settings.sort.map do |section, params| -%>
[<%= section %>]
<% params.sort.map do |key, val| -%>
<% Array(val).each do |item| -%>
<%= key %>=<%= item %>
<%- end -%>
<%- end -%>

<%- end -%>

```

This way the user can specify any of the hundreds (thousands?) of possible settings for whichever section supports them. This is also quite future-proof and wouldn't require changes to this module and its templates whenever a new version of systemd is released.

traylenator:
Hi @antaflos 
A more structured hash probably is better, I was copying the existing logic in 
https://github.com/camptocamp/puppet-systemd/blob/master/manifests/service_limits.pp#L20
which is like this despite the sections in the file.

Even with the new structure I would add validation. With out the validation there is little point
in not just using a raw file type.

I'm happy to change it to

bastelfreak:
Hey @traylenator, sorry for the slow response here :(
Could you rebase this against master? Systemd adds more options frequently, so we should support a hash. Validation should happen for all needed bits.

templates/unit_file.erb
```diff
@@ -0,0 +1,85 @@
+#puppet managed file.
```
bastelfreak: can you use a epp template instead?

reenberg:
> With out the validation there is little point in not just using a raw file type.

Couldn't one just as well run the resulting file through `systemd-analyze verify` before actually using it, potentially in a tmp file?  
This way the module doesn't have to play "catch the mouse" with features implemented in systemd.
 
I'm currently using [pk/systemd](https://github.com/DesideriusE/puppet-systemd), and I find it useful to be able to specify simple things just as a hash, instead of having to add a template file (though it has its limits DesideriusE/puppet-systemd#3).  

My biggest fear is that I would end up having to wait for a new feature being added to the validation before I can use it.

Issue #15, opened 2016-10-24
Manage drop-in files


jarro2783:
Can support for drop-in files to override package units be added to this module?

raphink:
Could you give a concrete example of what you need?

jarro2783:
If I want to override some parameters for a systemd unit, then I would put a drop in file in `/etc/systemd/system/unit.service.d`, with the parameters that I want to override. So perhaps I could do something like:
```
systemd::unit_dropin { '99environment':
  unit => 'unitname',
  content => "[Service]\nStandardOutput=syslog\n",
}
```
and the contents would get placed in `/etc/systemd/system/unitname/99environment.conf`.

gauthier-delacroix:
That would be a nice addition to customize systemd services without overriding the entire original file.

TheMeier:
can this be closed now? and is PR #26 still needed at all?

jarro2783:
Yeah I think this is resolved by #39.

PR #26, opened 2017-05-15
support for drop-in unit files

rj667:
Hi, this should fix issue #15. We run this in production on 300+ systemd servers.

### DIFF
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -12,10 +12,10 @@ define systemd::unit_file(
   if $title =~ /^([^\/]+)\// {
     $subdir = $1
     if $subdir !~ /\.d$/ {
-      fail("drop-in unit dirs must have suffix .d")
+      fail('drop-in unit dirs must have suffix .d')
     }
     if $title !~ /\.conf$/ {
-      fail("drop-in unit files must have suffix .conf")
+      fail('drop-in unit files must have suffix .conf')
     }
     file { "${path}/${subdir}":
       ensure  => directory,
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -9,6 +9,25 @@ define systemd::unit_file(
 ) {
   include ::systemd
 
+  if $title =~ /^([^\/]+)\// {
+    $subdir = $1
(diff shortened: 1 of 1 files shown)


manifests/unit_file.pp
```diff
@@ -9,6 +9,25 @@
...
 
+  if $title =~ /^([^\/]+)\// {
+    $subdir = $1
+    if $subdir !~ /\.d$/ {
+      fail('drop-in unit dirs must have suffix .d')
+    }
+    if $title !~ /\.conf$/ {
+      fail('drop-in unit files must have suffix .conf')
+    }
+    file { "${path}/${subdir}":
```
raphink: This will fail if you try to manage several files in the same subdir.

rj667: You are right. We don't do that, so I hadn't noticed. We do override multiple configuration options of a single service, but we put those in a single drop-in unit file. I wonder if there is an actual use case for multiple drop-in unit files for the same service but that is a different discussion.
That situation could be solved by replacing the file{} with an exec{} that does "mkdir -p". Would you like me to do that or would you rather reject the whole patch?

salekseev: +1

raphink: @rj667 I'd rather avoid an exec resource. If you have to take that road, I'd rather you use `ensure_directory()`

manifests/unit_file.pp
```diff
@@ -9,6 +9,25 @@
 ) {
   include ::systemd
 
+  if $title =~ /^([^\/]+)\// {
```
bastelfreak: I think this whole logic can be simplified by using a proper datatype for the parameter?

bastelfreak:
Hi @rj667, are you still interested in this PR? If so, can you please rebase it against our current master?

manifests/unit_file.pp
ekohl: AFAIK titles can't have data types.

bastelfreak: Ah right, not in the params section I think. But we can still use assert_type()?

kenyon:
This should be closed due to #39.

PR #30, opened 2017-07-12
Add a network defined resource

bastelfreak:


bastelfreak:
Hi all, I added this as a WIP and I'm currently testing this locally. Could you provide me with feedback? I will add tests later on.

manifests/network.pp
```diff
@@ -0,0 +1,38 @@
...
+  Optional[String] $source               = undef,
+  Optional[Stdlib::Absolutepath] $target = undef,
+  Boolean $restart_service               = true,
+  Boolean $manage_service                = true,
+){
+
+  include ::systemd
+
+  if $manage_service {
+    service{'systemd-networkd':
```
ekohl: I'd split this off in a separate class which you can include. That way you don't need to choose one network where you manage the service.

README.md
```diff
@@ -79,3 +79,18 @@ Or provide the configuration file yourself. Systemd reloading and restarting of
   source => "puppet:///modules/${module_name}/foo.conf",
 }
 ```
+
+### network
+
+systemd-networkd is able to manage your network configuration. We provide a
+defined resource which can write the interface configurations. systemd-networkd
+needs to be restarted to apply the configs. The defined resource can do this
+for you, besided managing the service itself.
```
ekohl: s/besided/besides/

manifests/network.pp
bastelfreak: should we but this into the init.pp?

README.md
bastelfreak: thx, fixed

manifests/network.pp
ekohl: I was thinking of a ```systemd::networkd``` class that's included here, but since configuration goes through init that could work as well.

manifests/network.pp
```diff
@@ -0,0 +1,30 @@
...
+
+  include ::systemd
+
+  if $restart_service {
+    $notify = Service['systemd-networkd']
+  } else {
+    $notify = undef
+  }
+
+  file { "${path}/${title}":
```
raphink: I'd rather use `$name` than `$title` to allow for a propre namevar mechanism.

Issue #37, opened 2017-07-25
Better test for systemd (and other init systems)


pillarsdotnet:
The recommended test for init=systemd fails on systems where the init binary is named "/sbin/init" rather than "systemd".

The following test works on every system I've tried so far:

```
/proc/1/exe --version | grep -q 'systemd'
```

pillarsdotnet:
Submitted Pull-Request #38 to fix the problem.

TheMeier:
Use https://github.com/puppetlabs/puppetlabs-stdlib#service_provider instead.

jfleury-at-ovh:
This issue is fixed by #42.

PR #38, opened 2017-07-25
Better test for systemd existence.

pillarsdotnet:
As noted in #37 

Testing for the name of the init process fails on systems where init is always /sbin/init regardless of source package.

A better test is to check the output of `/proc/1/exe --version` for substrings.  This has the added benefit of distinguishing among non-systemd init systems.

### DIFF
--- a/lib/facter/systemd.rb
+++ b/lib/facter/systemd.rb
@@ -1,10 +1,27 @@
+# Fact: initsystem
+#
+# Purpose:
+#   Determine the init system in use on the node
+#
+# Resolution:
+#   Check the output of `/proc/1/exe --version` for identifying substrings:
+#   * runit will include "runit" substring.
+#   * systemd will include "systemd" substring.
+#   * upstart will include "upstart" substring.
+#   * sysvinit will include "invalid option" substring.
+#   * If none of the above are present, sets :initsystem => :unknown
+#
+# Caveats:
+#
+
 # Fact: systemd
 #
-# Purpose: 
+# Purpose:
 #   Determine whether systemd is the init system on the node
 #
 # Resolution:
-#   Check the name of the process 1 (ps -p 1)
+#   Returns true if :initsystem => :systemd
+#   Returns false otherwise
 #
 # Caveats:
 #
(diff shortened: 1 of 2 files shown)


lib/facter/systemd.rb
```diff
@@ -20,10 +37,28 @@
 # Caveats:
 #
 
+Facter.add(:initsystem) do
+  confine :kernel => :linux
+  setcode do
+    output = Facter::Util::Resolution.exec('/proc/exe/1 --version')
```
ekohl: On my Fedora system this doesn't exist. Did you mean ```/proc/1/exe```?

pillarsdotnet: Typo corrected; thanks.

spec/unit/facter/systemd_spec.rb
```diff
@@ -12,7 +12,7 @@
       end
       let(:facts) { {:kernel => :linux} }
       it do
-        Facter::Util::Resolution.expects(:exec).with('ps -p 1 -o comm=').returns('systemd')
```
ekohl: You still need to mock an exec call so it works on non-systemd machines as well

pillarsdotnet: I confess that I've got a lot to learn about rspec.

I like your idea of testing for /run/systemd as a lighter-weight solution, though.

ekohl: To make the current test run you could use:
```ruby
Facter::Util::Resolution.expects(:exec).with('/proc/1/exe --version').returns('systemd 233')
```

Modify ```systemd 233``` for other platforms in the other tests.

raphink:
Thank you for this PR. One problem I have with it is that `/proc/1/exe` can only be run by root, whereas the older command could be run by any user.

TheMeier:
Why don't you use https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/facter/service_provider.rb ?

ekohl:
@TheMeier because I didn't know that one existed. Nice!

PR #39, opened 2017-08-16
Add support for drop-in files

countsudoku:
Hi,

This solves #15. I know, there is another pull request, which manage drop-in files. But my solution is able to mange multiple drop-in files, even when the service file isn't managed by the `::systemd` module. I've also updated the spec tests and the README.

Regards

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,8 @@ file { '/usr/lib/systemd/system/foo.service':
 
 ### drop-in files
 
-Drop-in files are used to add or alter settings of a unit without modify the
-unit it self. As for the unit files, the module can handle the file and
+Drop-in files are used to add or alter settings of a unit without modifying the
+unit itself. As for the unit files, the module can handle the file and
 directory creation and systemd reloading:
 
 ```puppet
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -36,6 +36,39 @@ file { '/usr/lib/systemd/system/foo.service':
 ~> Class['systemd::systemctl::daemon_reload']
 ```
 
+### drop-in files
+
+Drop-in files are used to add or alter settings of a unit without modify the
```
raphink: typo: modifying

README.md
```diff
@@ -36,6 +36,39 @@ file { '/usr/lib/systemd/system/foo.service':
...
+unit it self. As for the unit files, the module can handle the file and
```
raphink: "itself" is one word

trevor-vaughan:
@raphink Fixed the noted typos.

types/dropin.pp
```diff
@@ -0,0 +1 @@
+type Systemd::Dropin = Pattern['^.+\.conf$']
```
ekohl: I wonder if it should check if there's no slashes in there.

countsudoku: I can add a check if there is no slash in there. I tried to use the same kind of pattern definition as for `Systemd::Unit` and there isn't a check for slashes.

PR #41, opened 2017-09-11
unit.d directory should be purged of unmanaged dropin files

treydock:


ekohl:
Should this be configurable?

raphink:
Making it configurable would be good; however I wonder where the parameter should reside.

ekohl:
On the main systemd class and then add a parameter on this class that defaults to the value on the systemd class? That's the only thing I can think of because of the ```ensure_resource```.

raphink:
I :+1: the parameter on the systemd class, but the current code is not a class, it's a define, so it should just use the value from the systemd class and not expose anything.

ekohl:
@raphink not exposing a parameter might even be better since the ```include ::systemd``` is inside the class so you're only guaranteed to have the value after that.

treydock:
@ekohl @raphink Parameter added and updated unit tests too.  The default is to purge, can change if needed.

manifests/dropin_file.pp
```diff
@@ -50,9 +50,11 @@
 
   if $ensure != 'absent' {
     ensure_resource('file', "${path}/${unit}.d", {
-      ensure => directory,
-      owner  => 'root',
-      group  => 'root',
+      ensure  => directory,
+      owner   => 'root',
+      group   => 'root',
+      recurse => $::systemd::purge_dropin_dirs,
```
ekohl: Is recursing needed? I thought purge would be enough.

treydock: @ekohl I think recurse is needed, at least that's my read of docs for both purge and recurse.  I use that pattern in my profiles where I'm essentially doing what this define does, it works as expected though recurse does seem unnecessary at first.

https://docs.puppet.com/puppet/latest/type.html#file-attribute-purge

treydock:
PR rebased

manifests/dropin_file.pp
bastelfreak: this is maybe a good candidate for an acceptance test?

manifests/init.pp
```diff
@@ -42,6 +42,7 @@
   Enum['stopped','running']        $timesyncd_ensure = 'running',
   Optional[Variant[Array,String]] $ntp_server          = undef,
   Optional[Variant[Array,String]] $fallback_ntp_server = undef,
+  Boolean                         $purge_dropin_dirs = true,
```
bastelfreak: can you check that every parameter is aligned at `=`?

bastelfreak:
Hey @treydock, sorry for the slow response here :(
Can you rebase again please?

treydock:
@bastelfreak rebased.  This module has no acceptance tests, should I add some in this pull request?

PR #42, opened 2017-09-15
Reuse the service_provider fact from stdlib

ekohl:


bastelfreak:
fact got introduced in stdlib 4.10.0, we require 4.13.1 in this module. Thanks for the PR @ekohl.

PR #43, opened 2017-10-02
Add systemd-timesyncd support

bastelfreak:


spec/classes/init_spec.rb
```diff
@@ -23,6 +23,14 @@
           it { is_expected.to create_service('systemd-networkd').with_ensure('running') }
           it { is_expected.to create_service('systemd-networkd').with_enable(true) }
         end
+        context 'when enabling timesyncd' do
```
ekohl: Maybe extend the existing test where both resolved and networkd are enabled? Or at least add the ```is_expected.to_not``` to the plain test?

ekohl: I meant to add ```is_expected.not_to``` on line 13 as well.

bastelfreak: ups. added that as well

manifests/init.pp
```diff
@@ -18,12 +18,30 @@
...
-  Boolean                          $manage_networkd = false,
-  Enum['stopped','running']        $networkd_ensure = 'running',
+  Optional[Systemd::ServiceLimits] $service_limits   = undef,
+  Boolean                          $manage_resolved  = false,
+  Enum['stopped','running']        $resolved_ensure  = 'running',
+  Boolean                          $manage_networkd  = false,
+  Enum['stopped','running']        $networkd_ensure  = 'running',
+  Boolean                          $manage_timesyncd = false,
+  Enum['stopped','running']        $timesyncd_ensure = 'running',
+  Optional[String] $ntp_server                       = undef,
```
wyardley: Why not an array?

bastelfreak: I thougt about it. Systemd expects a comma separated list. What do you think about an array which I join()?

alexjfisher: yep, I think that's better.  The API you expose through puppet should make sense in puppet land.  Use join in the template or wherever.

wyardley: I think that would be a cleaner interface, personally (having it be structured data in Puppet, and join it later). Not sure whether it should require an array for a single element list, or whether it should allow you to pass in a string and force to list context.

alexjfisher: If you're being really friendly, I suppose you might also accept a string with no commas in instead of forcing people to use arrays of size 1.

wyardley: And if it's an array, the params should probably be `ntp_servers` and `fallback_ntp_server`?
This also makes it clear that the params can take multiple values

trevor-vaughan:
Not all systems have `timesyncd`. I think that enabling the `timesyncd` class should be predicated on `timesyncd` actually being present on the system.

The same goes for `networkd`

manifests/init.pp
alexjfisher: @wyardley +1 Agreed.  If you expect an array, make the pluralise the parameter name.

bastelfreak:
@trevor-vaughan it defaults to off/dont manage. Is that okay? Is there a way to check the presence without a fact?

manifests/init.pp
bastelfreak: I would like to keep ntp_server and fallback_ntp_serve because those are the keys in the ini file. I updated the code to accept a single string or arrays. What do you think about that?

trevor-vaughan:
@bastelfreak I'm just worried that someone will turn it on without knowing that it's not present on their system and "bad things"(tm) will happen.

For instance, my Fedora system has both of these, but my RHEL 7 system does **not**. If I have something in Hiera based on `osfamily=RedHat`, then it's not going to be pretty for my Fedora systems.

While, technically, this is the fault of the user, it would be best to fail safe and perhaps notify them that they're trying to do something that won't work.

Unfortunately, you can't base it off of existing facts from what I can tell, you're going to have to actually query `systemd` itself to see what capabilities it actually has.

manifests/init.pp
wyardley: I personally think it should be plural, will defer to @alexjfisher

bastelfreak:
@trevor-vaughan uff. This needs a huge amount of code. I agree that it would be nice. I don't know a  sane way to query systemd about it. One solution would be to check for /usr/lib/systemd/systemd-timesyncd, but the path isn't the same on all distributions. Should this improvement be moved into its own pull request?

trevor-vaughan:
@bastelfreak It's a `systemd` service. `systemctl list-unit-files systemd-timesyncd.service` appears to work just fine.

trevor-vaughan:
@bastelfreak Did a bit more poking and this seems to give a good list that would make a great `facter` hash: `systemctl list-unit-files --plain systemd-* | grep -v @ | grep -v static`

bastelfreak:
oh right, I totally forgot that :D

bastelfreak:
```bash
systemctl list-unit-files --no-pager | awk '!/static|@/ && $1 ~ /systemd-.*.service/ { print $1 " " $2}'
```

I can wrap that up tomorrow in a fact (or do it in pure ruby)

trevor-vaughan:
Pure ruby would probably be less fragile over time, but that would be ideal, thanks!

bastelfreak:
Jo @trevor-vaughan! The awesome @jreinert and I added a ruby implementation of the structured fact. Output is:
```
{
  systemd-fsck-root.service => "enabled-runtime",
  systemd-journal-gatewayd.service => "indirect",
  systemd-journal-remote.service => "indirect",
  systemd-journal-upload.service => "disabled",
  systemd-networkd-wait-online.service => "enabled",
  systemd-networkd.service => "enabled",
  systemd-nspawn@.service => "disabled",
  systemd-resolved.service => "enabled",
  systemd-timesyncd.service => "disabled"
}
```

What do you think about that?

lib/facter/systemd.rb
```diff
@@ -1,17 +1,17 @@
 # Fact: systemd
 #
-# Purpose: 
-#   Determine whether systemd is the init system on the node
+# Purpose:
+#   Determine whether SystemD is the init system on the node
```
ekohl: it's plain systemd, not SystemD according to https://www.freedesktop.org/wiki/Software/systemd/

lib/facter/systemd.rb
```diff
@@ -20,10 +20,20 @@
...
+# Resolution:
+#  Check the output of systemctl --version
+#
+# Caveats:
+#
 Facter.add(:systemd) do
   confine :kernel => :linux
   setcode do
-    Facter.value(:service_provider) == 'systemd'
+    Facter::Util::Resolution.exec('ps -p 1 -o comm=') == 'systemd'
```
ekohl: Please don't revert this

manifests/timesyncd.pp
```diff
@@ -0,0 +1,66 @@
...
+    'running' => true,
+    default   => $ensure,
+  }
+
+  service{ 'systemd-timesyncd':
+    ensure => $ensure,
+    enable => $_enable_timesyncd,
+  }
+
+  if $ntp_server {
```
ekohl: Should it otherwise be ensured to be absent?

lib/facter/systemd.rb
bastelfreak: oh dang. I did some copy and paste issues. my patch for this file isn't based on master. I will fix that

lib/facter/systemd.rb
bastelfreak: fixed.

lib/facter/systemd.rb
bastelfreak: fixed

manifests/timesyncd.pp
bastelfreak: the linux distribution may place default values in there. so IMO we should not touch it.

EmilienM:
:+1:

manifests/init.pp
```diff
@@ -32,11 +50,15 @@
     create_resources('systemd::service_limits', $service_limits)
   }
 
-  if $manage_resolved {
+  if $manage_resolved and $facts['systemd_internal_services']['systemd-resolved.service'] {
```
trevor-vaughan: This is going to fail hard if the fact could not be populated for some reason.

Should probably be: `if $mange_resolved and $facts['systemd_internal_services'] and $facts['systemd_internal_services']['systemd-resolved.service']

manifests/init.pp
```diff
@@ -32,11 +50,15 @@
     create_resources('systemd::service_limits', $service_limits)
   }
 
-  if $manage_resolved {
+  if $manage_resolved and $facts['systemd_internal_services'] and $facts['systemd_internal_services']['systemd-resolved.service'] {
```
ekohl: Maybe a bit late to the party but if ```manage_resolved``` is true and the system can't use it, should that be an error?

bastelfreak: this would add another layer, not sure if this is actually needed. Is it safe to assume that only people that know their systems try to enable it? (Or at least know if they run systemd or not)

bastelfreak: Also using the module would be a bit more complicated/requires more logic. Currently I apply it to _all_ servers in our org, no matter if its a freebsd, old centos or something with systemd. Fail safely vs fail hard

ekohl: On the other hand, if you don't manage this you might silently not manage something when you expect it to. Makes me wonder if it should be a tristate: off, auto and on where on has checks if it's actually enabled.

bastelfreak: I thought about this, but unsure of the needed complexity is worth it

PR #44, opened 2017-10-17
release 1.1.0

bastelfreak:
@raphink

PR #45, opened 2017-10-17
implement github changelog generator

bastelfreak:
This is the code used to generate the changelog from #44. I am not sure if you want to have this in the module, or if you've a modulesync which would drop it. The used commit was tested on Vox Pupuli. Sadly there wasn't a working release in a longer time for the gem (but they are working on it)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 language: ruby
 sudo: false
 cache: bundler
-bundler_args: --without system_tests
+bundler_args: --without system_tests --without release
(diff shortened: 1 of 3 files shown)


bastelfreak:
Ping @raphink @mcanevet

mcanevet:
@bastelfreak isn't this part of PDK? If so, maybe we should move all our modules to it.

bastelfreak:
I've no idea. I had no time yet to look into it :(

bastelfreak:
This isn't currently part of the pdk, Puppet is still working on it. I think we should include this now to create nice releases. We can still purge it in a few months/years if PDK will be able to handle it.

PR #46, opened 2017-11-14
Use the same systemd drop-in file for different units

countsudoku:
If you wan't to use the same systemd drop-in file for different systemd units, you have to come up with some artificial title for the ::systemd::dropin_file resource. This is because the $name is used as filename. Hence you get a duplicate definition if you want to use the same drop_in file for different units.

Now there is a filename parameter were you can set the filename explicit different from the resource title.

### DIFF
--- a/manifests/dropin_file.pp
+++ b/manifests/dropin_file.pp
@@ -29,15 +29,14 @@
 #
 define systemd::dropin_file(
   Systemd::Unit                     $unit,
-  Enum['present', 'absent', 'file'] $ensure  = 'present',
-  Stdlib::Absolutepath              $path    = '/etc/systemd/system',
-  Optional[String]                  $content = undef,
-  Optional[String]                  $source  = undef,
-  Optional[Stdlib::Absolutepath]    $target  = undef,
+  Systemd::Dropin                   $filename = $name,
+  Enum['present', 'absent', 'file'] $ensure   = 'present',
+  Stdlib::Absolutepath              $path     = '/etc/systemd/system',
+  Optional[String]                  $content  = undef,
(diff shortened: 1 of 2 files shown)


spec/defines/dropin_file_spec.rb
```diff
@@ -36,6 +36,21 @@
             }.to raise_error(/expects a match for Systemd::Dropin/)
           }
         end
+
+        context 'with explicit filename' do
+            let (:title) {'test'}
```
ekohl: Indenting is a bit odd here

countsudoku: you are absolutely right! 
I'll fix that and create a test for the same drop-in file for two services.

countsudoku: I've fixed the indenting and changed unit-test to the explicit test you suggested.

manifests/dropin_file.pp
```diff
@@ -29,16 +29,15 @@
 #
 define systemd::dropin_file(
   Systemd::Unit                     $unit,
-  Enum['present', 'absent', 'file'] $ensure  = 'present',
-  Stdlib::Absolutepath              $path    = '/etc/systemd/system',
-  Optional[String]                  $content = undef,
-  Optional[String]                  $source  = undef,
-  Optional[Stdlib::Absolutepath]    $target  = undef,
+  Systemd::Dropin                   $filename = $name,
```
bastelfreak: Datatypes \o/

ekohl: Technically ```$name``` was already validated below with a function :)

PR #50, opened 2017-11-22
Use a space-separated in timesyncd.conf

hfm:
According to timesyncd.conf(5) in man, NTP and FallbackNTP need a space-separated list: https://www.freedesktop.org/software/systemd/man/timesyncd.conf.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,8 +163,8 @@ It is possible to configure the default ntp servers in /etc/systemd/timesyncd.co
 ```puppet
 class{'::systemd':
   $manage_timesyncd => true,
-  $ntp_server          => '0.pool.ntp.org,1.pool.ntp.org',
-  $fallback_ntp_server => '2.pool.ntp.org,3.pool.ntp.org',
+  $ntp_server          => '0.pool.ntp.org 1.pool.ntp.org',
+  $fallback_ntp_server => '2.pool.ntp.org 3.pool.ntp.org',
 }
 ```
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -163,8 +163,8 @@ It is possible to configure the default ntp servers in /etc/systemd/timesyncd.co
 ```puppet
 class{'::systemd':
   $manage_timesyncd => true,
-  $ntp_server          => '0.pool.ntp.org,1.pool.ntp.org',
-  $fallback_ntp_server => '2.pool.ntp.org,3.pool.ntp.org',
+  $ntp_server          => '0.pool.ntp.org 1.pool.ntp.org',
+  $fallback_ntp_server => '2.pool.ntp.org 3.pool.ntp.org',
```
raphink: How about passing an array instead?

hfm: $ntp_server and $fallback_ntp_server in systemd class now don't allow Array: https://github.com/camptocamp/puppet-systemd/blob/1.1.0/manifests/init.pp#L43-L44

hfm: IMO, it's better that systemd class allow $ntp_server and $ntp_fallback_server to use both Array and String. If you're okay, I'll do.

raphink: Oh ok. That's weird, cause the `timesyncd` class accepts arrays. @bastelfreak you originally wrote that class, what do you think?

ekohl: Sounds like an oversight and ```init.pp``` should match the types of ```timesyncd.pp```

bastelfreak: spaces are correct, not sure how I missed it. @hfm can you also allow array as datatype in the init.pp?

hfm: okay i'll change it.

hfm: I fixed it by 2c38f1a

PR #51, opened 2017-11-22
Use Hash for $service_limits of systemd class

hfm:
`create_resource('systemd::service_limits', $service_limits)` of systemd class (init.pp) fails unless `$service_limits` is set to Hash.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -33,7 +33,7 @@
 #   as the fallback NTP servers. Any per-interface NTP servers obtained from
 #   systemd-networkd take precedence over this setting. requires puppetlabs-inifile
 class systemd (
-  Optional[Systemd::ServiceLimits] $service_limits   = undef,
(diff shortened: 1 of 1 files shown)


raphink:
Maybe you're using a key that is not known in the data type? Check in https://github.com/camptocamp/puppet-systemd/blob/master/types/servicelimits.pp and add if needed.

hfm:
> Maybe you're using a key that is not known in the data type?

No. This is not a Data type problem. `create_resource` function needs Hash: https://puppet.com/docs/puppet/latest/function.html#createresources I think that `$service_limits` in systemd class now don't work.

ekohl:
https://github.com/camptocamp/puppet-systemd/pull/52 does the same thing, but also adds a test. Didn't realize you already had a PR for this.

hfm:
Oh. #52 is better than my PR. I'll close this.

Issue #49, opened 2017-11-20
Not able to set limits via systemd class


redcricket:
I want to setup some systemd limits for a service so I have included ...

```
    include ::systemd
```
... in my manifest.  And I have this hiera data:

```
systemd::service_limits:
    'openstack-nova-compute.service':
        limits:
                LimitNOFILE: 32768
```

But when puppet on my node I get this error:

```
Error while evaluating a Function Call, Class[Systemd]: parameter 'service_limits' unrecognized key 'openstack-nova-compute.service' at ...
```

I have looked at the docs and they do not explain how to form the hash in hiera, they just give this example:

```
::systemd::service_limits { 'foo.service':
  limits => {
    'LimitNOFILE' => 8192,
    'LimitNPROC'  => 16384,
  }
}
```

So how would I do that in hiera/yaml?

redcricket:
I have posted a question about this on [stackoverflow](https://stackoverflow.com/questions/47397542/how-should-i-structure-my-hiera-for-systemd-limits)

ekohl:
The stackoverflow answer is right: it should be a different type. I think ```Hash[String, Hash[String, Any]]``` would make the most sense while changing the default to ```{}```.

PR #52, opened 2017-11-22
Use the correct type on $service_limits

ekohl:
Closes GH-49


raphink:
Why not adapt `Systemd::ServiceLimits` instead?

ekohl:
This calls create_resources on the ```systemd::service_limits``` defined type. That has a parameter ```limits``` with the ```Systemd::ServiceLimits``` which does the actual checks.

raphink:
Ah, that makes sense. LGTM then.

bastelfreak:
Thanks for the PR @ekohl

ekohl:
@bastelfreak I'd say this is a bugfix rather than an enhancement. Before the parameter was unusable.

PR #53, opened 2017-11-22
Correct parameter documentation

ekohl:


bastelfreak:
ups, thanks for fixing my copy and paste foo.

PR #55, opened 2017-11-23
Add enable parameter to enable the service with systemd

acurus-puppetmaster:


bastelfreak:
Hi @acurus-puppetmaster, thanks for the PR. Can you take a look at the failed travis jobs?

acurus-puppetmaster:
Failures corrected @bastelfreak

ekohl:
I wonder if it needs a 3 parameter solution: ```manage_service```, ```enable_service``` and ```ensure_service```.

acurus-puppetmaster:
Can you ellaborate what the difference would be between
`manage_service`, `enable_service` and `ensure_service`

This module already manages the service definition, but does not enable or disable it.

So I can't understand what `manage_service` and `ensure_service` are?

ekohl:
I was thinking of:
```puppet
if $manage_service {
  service { $title:
    ensure => $ensure_service,
    enable => $enable_service,
  }
}
```

Then manage would default to ```false``` and the other to ```true```/```running```.

acurus-puppetmaster:
Added additional logic.

But defaults are `$manage_service = false`, `$enable_service = false` and `$ensure_service = stopped` so the unit file just exists but the service is **not** enabled at boot and **not** set to running.

This continues the default behaviour of this module.

If you set it to `$manage_service = true`, it will still default to `$enable_service = false` and `$ensure_service = stopped`, the underlying [puppet type ](https://puppet.com/docs/puppet/5.3/types/service.html) doesn't specify defaults values for ensure and enable.

You can then ensure a service is running independent of whether a service is enabled at boot or not.

I think most people would end up setting
```
$manage_service = true
$enable_service = true
$ensure_service = running
```

But I am not sure that should be the default.

manifests/unit_file.pp
```diff
@@ -57,4 +66,11 @@
     mode    => '0444',
     notify  => Class['systemd::systemctl::daemon_reload'],
   }
+
+  if $manage_service {
+    service { $title:
```
ekohl: I think they're the same, but in the rest of the file ```$name``` is used.

manifests/unit_file.pp
```diff
@@ -27,12 +27,21 @@
...
-  Stdlib::Absolutepath              $path    = '/etc/systemd/system',
-  Optional[String]                  $content = undef,
-  Optional[String]                  $source  = undef,
-  Optional[Stdlib::Absolutepath]    $target  = undef,
+  Enum['present', 'absent', 'file'] $ensure         = 'present',
+  Stdlib::Absolutepath              $path           = '/etc/systemd/system',
+  Optional[String]                  $content        = undef,
+  Optional[String]                  $source         = undef,
+  Optional[Stdlib::Absolutepath]    $target         = undef,
+  Boolean                           $manage_service = false,
```
ekohl: This is undocumented.

acurus-puppetmaster:
When we can expect this to be merged?

bastelfreak:
@acurus-puppetmaster can you add unit tests for those changes?

acurus-puppetmaster:
Can you be more specific with what you would like tested @bastelfreak.

These changes don't modify the file produced by unit_file.

They simply pass additional options to the service resource.

We shouldn't write test cases for the service resource as its not in the scope of the module, its core puppet functionality.

ekohl:
@acurus-puppetmaster I'd expect a test that verifies the service is not managed by default (```it { is_expected not.to contain_service('something') }```) and another that sets ```$manage_service``` to true and verifies it is managed. That's within the scope of this module and helps ensuring the module continues working even if it's refactored.

johanfleury:
@acurus-puppetmaster any updates on this PR?

manifests/unit_file.pp
```diff
@@ -55,6 +69,17 @@
...
+  File["${path}/${name}"] -> Class['systemd::systemctl::daemon_reload']
+
+  if $manage_service {
+
+    service { $name:
+      ensure => $ensure_service,
+      enable => $enable_service,
+    }
+
+    File["${path}/${name}"] -> Service[$name]
```
ekohl: Shouldn't this be after the daemon reload?

manifests/unit_file.pp
```diff
@@ -55,6 +69,17 @@
     owner   => 'root',
     group   => 'root',
     mode    => '0444',
-    notify  => Class['systemd::systemctl::daemon_reload'],
+  }
+
+  File["${path}/${name}"] -> Class['systemd::systemctl::daemon_reload']
```
ekohl: I think the notify that used to be there was fine.

jcharaoui:
Unit files can be [many different things](https://github.com/camptocamp/puppet-systemd/blob/master/types/unit.pp#L1) other than services, so the module should at least check if its dealing with a `.service` unit before declaring a `service` resource. In my opinion, the ideal solution would be to implement a new defined type, such as `systemd::unit::service` to contain this logic (and manage the service by default).

jcharaoui:
Also, please keep in mind that not just service units may be enabled. For example, when deploying a mount unit, only those which are enabled (using `systemctl enable foo.mount`) are mounted at boot. This module should support enabling all types of units.

johanfleury:
I think most of (if not all) the systemd units can be managed through puppet services though.

jcharaoui:
@johanfleury I was skeptical at first, but you're correct... as long as the service provider is set to `systemd`. The details can be found in the [documentation](https://puppet.com/docs/puppet/5.4/types/service.html#service-provider-systemd) and [this bug report](https://tickets.puppetlabs.com/browse/PUP-5285).

I think this means that to ensure predictable behavior across all operating systems, this module should declare service resources with `provider => systemd`.

@cliff-wakefield Are you still willing to work on this?

manifests/unit_file.pp
```diff
@@ -27,12 +27,26 @@
 #
 #   * Mutually exclusive with both ``$source`` and ``$content``
 #
+# @attr manage_service
```
johanfleury: As @jcharaoui said, a unit can be anything beside a service thus I find the names of these parameters to be confusing.

Maybe `$ensure_service` could be computed from `$ensure` and `$manage_service` (that would be renamed `$manage_unit` or just `$manage` idk) and `$enable_service` could then be renamed `$enable`.

bastelfreak:
Hi @acurus-puppetmaster, can you please check the failing spec tests and the inline comments?

bastelfreak:
Closing this in favor of https://github.com/camptocamp/puppet-systemd/pull/69

PR #57, opened 2017-11-30
Purge another NTP daemon

hfm:
In debian stretch or ubuntu xenial, systemd-timesyncd don't run if host has installed another NTP daemon:

```
root@987337e94ce5:/# cat /lib/systemd/system/systemd-timesyncd.service.d/disable-with-time-daemon.conf
[Unit]
# don't run timesyncd if we have another NTP daemon installed
ConditionFileIsExecutable=!/usr/sbin/ntpd
ConditionFileIsExecutable=!/usr/sbin/openntpd
ConditionFileIsExecutable=!/usr/sbin/chronyd
ConditionFileIsExecutable=!/usr/sbin/VBoxService
```

This commit allows to purge NTPd packages (default false).

### DIFF
--- a/manifests/timesyncd.pp
+++ b/manifests/timesyncd.pp
@@ -13,13 +13,28 @@
 #   A space-separated list of NTP server host names or IP addresses to be used
 #   as the fallback NTP servers. Any per-interface NTP servers obtained from
 #   systemd-networkd take precedence over this setting. requires puppetlabs-inifile
+#
+# @param purge_another_ntp_daemon
+#   Determine to purge another NTP daemon
 class systemd::timesyncd (
   Enum['stopped','running'] $ensure                    = $systemd::timesyncd_ensure,
   Optional[Variant[Array,String]] $ntp_server          = $systemd::ntp_server,
   Optional[Variant[Array,String]] $fallback_ntp_server = $systemd::fallback_ntp_server,
+  Boolean $purge_another_ntp_daemon                    = $systemd::purge_another_ntp_daemon,
 ){
 
   assert_private()
+
+  if $purge_another_ntp_daemon {
+    package { [
+      'ntp',
+      'chrony',
+      'openntpd',
+    ]:
+      ensure  => absent,
+      require => Service['systemd-timesyncd'],
+    }
+  }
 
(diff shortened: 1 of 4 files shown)


manifests/timesyncd.pp
```diff
@@ -13,14 +13,29 @@
...
   assert_private()
 
+  if $purge_another_ntp_daemon {
+    package { [
+      'ntp',
+      'chrony',
+      'openntpd',
+    ]:
+      ensure  => absent,
+      require => Service['systemd-timesyncd'],
```
ekohl: Shouldn't the purging happen before the service is started to prevent conflicts?

manifests/timesyncd.pp
```diff
@@ -13,14 +13,29 @@
...
   Enum['stopped','running'] $ensure                    = $systemd::timesyncd_ensure,
   Optional[Variant[Array,String]] $ntp_server          = $systemd::ntp_server,
   Optional[Variant[Array,String]] $fallback_ntp_server = $systemd::fallback_ntp_server,
+  Boolean $purge_another_ntp_daemon                    = $systemd::purge_another_ntp_daemon,
 ){
 
   assert_private()
 
+  if $purge_another_ntp_daemon {
+    package { [
```
ekohl: Maybe this list should at least be a class parameter so users can override it with hiera. It might also be an idea to use ```ensure_package``` though that can be weird if another one ensures it with present and it's not overridden.

hfm: Sorry for late reply. I changed it with cfcc3bc

bastelfreak:
Hi @hfm, thanks for this PR. The code has a good quality, but I'm unsure if this fits into this module. As far as I know, many people have puppetlabs-ntp installed. A module should only manage its own packages. This is kind of out of scope, but would be useful. @raphink what do you think about this?

raphink:
Yes indeed, I think the systemd module should not manage resources outside of the systemd scope.

bastelfreak:
I will close this PR for now. @hfm thanks for all the work you put into this. Feel free to reopen if we should discuss this further.

PR #58, opened 2017-12-08
Replace iterator with stdlib function

jfleury-at-ovh:
If I understand correctly, this line does what `stdlib`'s `prefix` does so I replaced it.


bastelfreak:
Thanks for the PR @jfleury-at-ovh!

Issue #62, opened 2018-01-15
LimitNOFILE restricted to integer type


AlbertoGarciaFdez:
Hi, why is LimitNOFILE an integer type? AFAIK infinity can be used with LimitNOFILE on systemd.
Should not be this more appropriate?

```
    Optional['LimitNOFILE']         => Pattern['^(infinity|(\d+)$'],
```

raphink:
LGTM. Can you turn this into a PR please?

AlbertoGarciaFdez:
Sure, thanks!

traylenator:
Also 'soft:hard' is also permitted.

raphink:
Leaving this open for now for the sake of `soft:hard`

PR #63, opened 2018-01-16
Use a pattern to allow infinity values on limitNOFILE

AlbertoGarciaFdez:
Switch from integer to a pattern type to allow infinity value on LimitNOFILE for service limit provider. #62


types/servicelimits.pp
```diff
@@ -7,7 +7,7 @@
     Optional['LimitSTACK']          => Pattern['^(infinity|((\d+(K|M|G|T|P|E)(:\d+(K|M|G|T|P|E))?)))$'],
     Optional['LimitCORE']           => Pattern['^(infinity|((\d+(K|M|G|T|P|E)(:\d+(K|M|G|T|P|E))?)))$'],
     Optional['LimitRSS']            => Pattern['^(infinity|((\d+(K|M|G|T|P|E)(:\d+(K|M|G|T|P|E))?)))$'],
-    Optional['LimitNOFILE']         => Integer[0],
+    Optional['LimitNOFILE']         => Pattern['^(infinity|(\d+))$'],
```
ekohl: Is ```Variant[Integer[0], Enum['infinity']]``` cleaner? This changes the datatype from a real integer to a string representation of integers.

AlbertoGarciaFdez: Agree, my commit would break current usage, my bad. Let me change it to your proposal, is exactly what im looking for, just add for infinity. Thanks!

traylenator:
I suggest:
```diff
-    Optional['LimitNOFILE']         => Integer[0],
+    Optional['LimitNOFILE']         => Variant[Integer[0],Pattern['^(infinity|((\d+(K|M|G|T|P|E)(:\d+(K|M|G|T|P|E))?)))$']],
```

ekohl:
That doesn't sound like it matches https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Process%20Properties:
> The multiplicative suffixes K, M, G, T, P and E (to the base 1024) may be used for resource limits measured in bytes (e.g. LimitAS=16G).

However, the table below lists it as a number, not bytes. That means it should be ```Variant[Integer[0],Pattern['^(infinity|\d+(:(infinity|\d+))?)$']]``` instead.

AlbertoGarciaFdez:
Agree, this has to match either an int or infinity, if using soft:hard, it has to match either int:'infinity' or int:int

ekohl:
Because I suggested the regex I also know that a test might be a good thing :)

bastelfreak:
Maybe this would be a good candidate for an acceptance test?

wyardley:
Is unlimited still valid as well, or is it deprecated in favor of 'infinity'?

I'm not sure that all this strict type checking of limits really benefits the module, but can we get this merged, with / without perfect tests? Feel free to cherry-pick the test case in #76 if it helps, but let's not let the perfect be the enemy of the good, especially Voxpupuli wants to encourage people to use the systemd module.

bastelfreak:
Should we just switch to `Variant[Integer,String]`?

trevor-vaughan:
@bastelfreak Please don't do this. The *only* thing that is valid is `infinity` and switching would open room for typos and/or adding things that simply won't work.

In addition to this change, there are others that need to be updated. Additionallly, the tests should test all three cases `infinity', `INT`, and `INT:INT`.

Per the `man` page the other items that need to be updated are:

* `LimitCPU`
* `LimitNPROC`
* `LimitLOCKS`
* `LimitSIGPENDING`
* `LimitNICE`
* `LimitRTPRIO`
* `LimitRTTIME`

This can be done relatively quickly. Negative numbers are valid for `LimitNICE` and should already be handled.

bastelfreak:
@trevor-vaughan did you by any chance check if the allowed values vary based on the systemd version? That happened in the past for other values :(

trevor-vaughan:
It's very probable. That said, so far they've been expansive instead of
restrictive.

Technically, we should probably check the besides and OS, but days types
don't allow for logic.

wyardley:
> The only thing that is valid is infinity and switching would open room for typos and/or adding things that simply won't work.

I guess at some point, I disagree that doing such careful type checking is helping end users of the module. Properly checking all possible values starts to get pretty nasty.

That said, if we _are_ doing type checking, maybe consider making reusable generic types if there is enough overlap between the syntax for certain types of checks, rather than having a ton of types using essentially 2 or 3 different patterns? And maybe force the hard / soft values to be an array? Otherwise, what's the point of forcing integers to be integers, but then allowing the string `500:800` and using regex to check validity? That's kind of confusing and inconsistent, esp. just a single `\d+` is not allowed to be a string. At the point where we're being that strict, maybe it would be better to require an array of ints and handle conversion to `N:N` in template?

I don't really care what solution gets implemented, but when the module doesn't support valid parameters, it's kind of hard to suggest that people switch to using it to manage things.

I agree that some more spec tests would be a good idea, though in the end, there are other values that allow different types of values we're not currently testing for. Invalid values raising an error I think isn't tested either?

trevor-vaughan:
@wyardley There will always be bugs and updates to new systems. We could, of course, strip out all type checking besides `String` but that won't help anyone who isn't a `systemd` expert or who thinks something "should work this way" based on what they remember (I've been there) and ends up screwing up their `systemd` config.

The Integer/Pattern items is there because a lot of these contain string additions that make it impossible to build with the current Type system using Integers but adding on a random string value.

In terms of `hard` and `soft`, I think we might want to use a Hash if you want users to be explicit.

```
limits => {
  'hard' => 12345,
  'soft' => 1234
}
```

wyardley:
The hash pattern would look pretty clean, I think. For settings that take units, would it then validate each element of the hash according to the regex?

Can we use `Struct` yet?
I guess you'd end up with something like (completely untested and probably with errors):

`Variant[Enum['infinity'], Struct[{hard => Variant[Integer[0], Enum['infinity'], Pattern[(\d+(K|M|G|T|P|E)], Optional[soft] => Variant[Integer[0], Enum['infinity'], Pattern[(\d+(K|M|G|T|P|E)]}]`
or for the NOFILES type ones
`Variant[Enum['infinity'], Struct[{hard => Variant[Integer[0], Enum['infinity']], Optional[soft] => Variant[Integer[0], Enum['infinity']]}]`

trevor-vaughan:
@wyardley Sure, `Struct` would be fine.

That said, we need to keep the existing material for legacy usage but we can probably consolidate a lot of it into a couple of sub-types and order them based on performance.

bastelfreak:
Implemented a basic version in https://github.com/camptocamp/puppet-systemd/pull/81

bastelfreak:
I am going to close this since an update got implemented. Please create a new PR if you would like to do more changes to the type validation.

PR #66, opened 2018-02-28
purge legacy puppet-lint checks

bastelfreak:
this check isn't needed anymore for puppet4 and newer


ekohl:
It'd be good to have the reverse lint check plugin to clean up existing code

bastelfreak:
ah right, I wanted to cleanup the existing code and lost track...

PR #65, opened 2018-02-28
Add support for Resource Accounting via systemd

bastelfreak:

PR #67, opened 2018-03-03
move params to data-in-modules

bastelfreak:
This got pulled out of https://github.com/camptocamp/puppet-systemd/pull/65. The lowest recommended puppet version for this is 4.10.10. In the Vox Pupuli namespace we had a lot of argumentation if this requires a major version bump or not. From a point of defensive programming this should be a major bump.

### DIFF
--- a/data/common.yaml
+++ b/data/common.yaml
@@ -1,0 +1,10 @@
+---
+systemd::service_limits: {}
+systemd::manage_resolved: false
+systemd::resolved_ensure: 'running'
+systemd::manage_networkd: false
+systemd::networkd_ensure: 'running'
+systemd::manage_timesyncd: false
+systemd::timesyncd_ensure: 'running'
+systemd::ntp_server: ~
+systemd::fallback_ntp_server: ~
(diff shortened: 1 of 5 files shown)


data/common.yaml
```diff
@@ -0,0 +1,10 @@
+---
+systemd::service_limits: {}
+systemd::manage_resolved: false
+systemd::resolved_ensure: 'running'
+systemd::manage_networkd: false
+systemd::networkd_ensure: 'running'
+systemd::manage_timesyncd: false
+systemd::timesyncd_ensure: 'running'
+systemd::ntp_server: ~
```
ekohl: I'm not sure what the best practice is, but I think it's cleaner to have any ```Optional``` type in ```init.pp``` default to ```undef```.

bastelfreak: AFAIK best practice is ~, which is also used in the puppetlabs modules.

ekohl: I guess it would be more consistent to have all parameters in hiera.

raphink:
LGTM, but I think there should be a warning in the README that you need to use the Hiera 5 format to use the module from then on.

PR #69, opened 2018-03-08
Add enable and active parameters to unit_file

jcharaoui:
Allows the deployed units to be enabled and/or started by the module. By default it will leave the units unmanaged.

This is an attempt to supersede PR #55 for which the original submitter appears to have lost interest.

### DIFF
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -27,12 +27,20 @@
 #
 #   * Mutually exclusive with both ``$source`` and ``$content``
 #
+# @attr enable
+#   If set, will manage the unit enablement status.
+#
+# @attr active
+#   If set, will manage the state of the unit.
+#
 define systemd::unit_file(
-  Enum['present', 'absent', 'file'] $ensure  = 'present',
-  Stdlib::Absolutepath              $path    = '/etc/systemd/system',
-  Optional[String]                  $content = undef,
-  Optional[String]                  $source  = undef,
-  Optional[Stdlib::Absolutepath]    $target  = undef,
+  Enum['present', 'absent', 'file']        $ensure  = 'present',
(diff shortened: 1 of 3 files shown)


manifests/unit_file.pp
```diff
@@ -57,4 +65,14 @@
     mode    => '0444',
     notify  => Class['systemd::systemctl::daemon_reload'],
   }
+
+  if $enable != undef or $active != undef {
```
ekohl: Previously it was mentioned there are other unit files like ```mount``` or ```timer```. Can ```service``` handle those as well or should we use a regex like ```$name =~ /^.+\.service$/``` in this if as well?

ekohl:
It'd be good if there was a testcase for this.

manifests/unit_file.pp
jcharaoui: The `service` native resource can definitely handle other types of unit files, as long as the provider is set as `systemd`.

jcharaoui: It should be noted, however, that unless a unit type suffix is used as the resource name (eg. `.mount` or `.timer`, it will default to a service unit type.

jcharaoui:
I wish I could write this testcase myself but I don't know enough about puppet module tests to do so myself, at his time.

bastelfreak:
@jcharaoui have you checked the existing tests in https://github.com/camptocamp/puppet-systemd/blob/master/spec/classes/init_spec.rb ? Is that something that helps you? If you've any questions about writing tests, Puppet, or rspec-puppet (the unit testing framework) feel ree to join the IRC channel #voxpupuli on freenode from the Puppet Community. We're happy to help out.

bastelfreak:
Thanks for the tests @jcharaoui !

PR #71, opened 2018-03-13
bump facter to latest 2.x version

bastelfreak:
It isn't necessary to pin it, facterdb now has all needed facts for 2.5


raphink:
LGTM

PR #73, opened 2018-03-15
Remain compatible with older Puppet 4.x releases

jcharaoui:
Puppet supports data-in-modules since 4.3, although the metadata syntax has changed a little bit. With these changes, the module can keep evolving with the current data-in-modules scheme while staying compatible with older Puppet 4 releases.

Once it's decided to completely drop Puppet 4 support, only small changes will be required.

### DIFF
--- a/hiera.yaml
+++ b/hiera.yaml
@@ -1,12 +1,13 @@
 ---
 version: 5
-defaults:
-  datadir: 'data'
-  data_hash: 'yaml_data'
+datadir: 'data'
 hierarchy:
   - name: 'Major Version'
+    backend: 'yaml'
     path: '%{facts.os.name}-%{facts.os.release.major}.yaml'
   - name: 'Distribution Name'
+    backend: 'yaml'
     path: '%{facts.os.name}.yaml'
   - name: 'common'
+    backend: 'yaml'
     path: 'common.yaml'
--- a/hiera.yaml
+++ b/hiera.yaml
@@ -1,5 +1,5 @@
 ---
-version: 5
+version: 4
 datadir: 'data'
 hierarchy:
   - name: 'Major Version'
--- a/metadata.json
+++ b/metadata.json
@@ -7,6 +7,7 @@
   "source": "https://github.com/camptocamp/puppet-systemd",
   "project_page": "https://github.com/camptocamp/puppet-systemd",
   "issues_url": "https://github.com/camptocamp/puppet-systemd/issues",
+  "data_provider": "hiera",
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
@@ -16,7 +17,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 4.10.10 < 6.0.0"
+      "version_requirement": ">= 4.7.0 < 6.0.0"
     }
   ],
   "operatingsystem_support": [


trevor-vaughan:
I would highly recommend this change.

bastelfreak:
Hi,
Puppet 4.9 and older are deprecated by Puppet, I am not sure if it is a good idea to still support it. As a module author we act as a software "vendor" to or users, we're responsible for them. We should not encourage them to run outdated software. We should also point out that people are not required to update their local module version, just because we released a newer version. The current releases of camptocamp/systemd are well tested and work well.

bastelfreak:
On Vox Pupuli we agreed on bumping everything to at least Puppet 4.10.0, and to newer versions if required. I suggest the same for this module. What are your objections?

jcharaoui:
In our case, we're simply not moving to 4.10 anytime soon. I can definitely understand that Vox Pupuli as a group may not want guarantee support for every Puppet version out there, but in this case, the diff is very small, and it doesn't hurt compatibility with current and new versions of Puppet (as the unit tests show). I simply wouldn't see the point of maintaining a version requirement "just for the sake of it".

baldurmen:
I agree with @jcharaoui on this one: I'm not upgrading my Puppet version until the new Debian stable is released (likely with Puppet 5).

Debian Buster should be released in a year or two and I would really like to be able to continue using this module

ge-fa:
Debian stable currently ships 4.8.2, the same version is backported to oldstable. I agree with others here: This should still be supported (even if that's is "deprecated"). Release schedules of puppetlabs inc. doesn't align well with people running production setups on stable releases.

bastelfreak:
@ge-fa this depends a bit on the point of view. We could also argue that some people don't align their update process with upstream releases. Also the Puppetlabs repos make it pretty easy to run on supported versions (but there are valid reasons against them, in some areas).

I don't want to be responsible for any issues that people encounter in modules I provide because they run outdated dependencies (that we don't even test). Vox Pupuli and camptocamp provide a wide range of modules. To make maintainability easier the support matrix should be as equal as possible in all modules. At least on Vox Pupuli this is very important to us because the list of modules is quite huge but the manpower is low (also CI resources are limited).

----

@baldurmen but you're not forced to upgrade your local module version just because there  is a new upstream one. The current releases are all working and have a good code coverage. You're able to continue to use this module.

lelutin:
@bastelfreak not garanteeing support for older versions is ok, and only aiming for 4.10 because it's an LTS is perfectly sane.

but in this case a small change makes it possible for everyone from "old" 4.x versions to the newest of the new in 5.x to run the same thing. you'll still be testing things for the puppet versions and os releases that you want to support and thus making sure that it works well for these (e.g. hiera 4 is still supported in puppet 5), but it opens up the possibility for ppl staying behind to take their chance running this module.

and we can reckon that we can still use this module if we stay behind on versions but that means we'll be missing out on all of the nice work that's being done to fix bugs and add features.

bastelfreak:
I agree with @binford2k here. In my opinion we should not add support for hiera 4. @mcanevet @raphink your opinions?

trevor-vaughan:
@bastelfreak The oldest LTS version of Puppet Enterprise 2016.4 does *NOT* support Hiera 5. It does support Hiera 4 (quite well in fact) and we have had no issues with this at all during testing. Unfortunately, the website documentation is unclear as to whether or not users *must* upgrade to retain LTS support (and generally that's not something a vendor can dictate). To that point, I've been testing against `4.7.0` until such time that the release is completely EOL.

It will be removed from support after October so I would suggest accepting this PR for now and then Hiera 5 after the Hiera 4 supporting versions are EOL since there is no breaking functionality (or system changes required) for this.

ahpook:
#include std_disclaimer.h

@trevor-vaughan the LTS has had puppet 4.10 agents since May 2017 (PE2016.4.5) and only the latest patch release is really supported, which was part of the rationale for Vox not extending back further.

that said, if authors want to be more generous on compatibility on modules they maintain, it's generally a Good Thing™ since the variance in puppet versions in the world is so wide. as @ge-fa says the upstream release/support lifecycle is more bleeding-edge than many people want to run.

trevor-vaughan:
@ahpook The "latest patch release is really supported" is what is the sticking point for me. I would *love* to be able to play that line and just ditch `4.7.0` forever but that's not stated nor implied by the website or documentation as far as I can tell.

ahpook:
@trevor-vaughan hm, you're right that it's not called out explicitly on the lifecycle page.  you could  *technically* get support to help with an issue on (e.g.) PE2016.4.3, but from a CI testing and engineering maintenance standpoint, only the latest release of a given series is "alive". Otherwise the test matrix explodes with every point release.

dhollinger:
@trevor-vaughan @ahpook I think maybe a definition on what _supported_ implies would be good.

My understanding is that while you could get support for PE2016.4.3 (or as a separate example Gitlab Enterprise 10.5.x), that version is not getting any more updates or active development. Meaning no security or bug fixes (which, IMO, is a bad thing to encourage with the amount of data leaks, breaches, and hacks that seem to happen every month).

I guess, what constitutes _supported_ in this context?

trevor-vaughan:
@dhollinger From a module point of view, my impression is that `supported` means "we tested this on this version and it worked as expected without regressions". It does *not* mean we encourage you to use that version, just that we won't destroy your system if you do (according to our testing).

dhollinger:
I'm not sure that "without regressions" can be guaranteed with this PR, though. As @binford2k mentioned, Hiera 4 was and is still experimental. While I did not use it, I do know a few that did have some issues crop up with Hiera 4. I would strongly recommend against using Hiera 4 as the compatibility layer for older Puppet 4.x releases.

Defaults for Parameters and using Hiera 5 module data can both be defined in a module. Perhaps either the removal of Module Data or adding defaults to required parameters would be best for Puppet < 4.10 support.

trevor-vaughan:
The only issues that we found were with knockout prefixes and some array/hash munging. I would suggest moving everything in `common.yaml` back up into `init.pp` so that `puppet strings` works properly.

Since the rest is a Hash, you could easily move it into `params.pp`. That said, you're no better, or worse, off than you are in `data` since overrides will work just fine but merging and knockouts won't work (which is the same as `params.pp` and broken in some versions of 5.X anyway).

This means that the functionality from `4.7.0` through `5.X` will be the same by default.

Honestly, this is a bit 'data in modules is the new hotness' and doesn't really add functional benefits unless I'm missing something.

I really do wish that the decision had been made to have two different paths in modules for Hiera 4 and 5 but there you go.

trevor-vaughan:
@bastelfreak That's my point. Per the conversation with @ahpook above, `4.7` is alive and kicking until October of this year.

ahpook:
trevor - i think 4.10 is the right target for this, sorry if my statement about what's technically supported by puppet inc muddied that particular puddle of water.

trevor-vaughan:
@ahpook Fundamentally, this whole thread is about technicalities. It's 100% OK to say "we're not going to support 4.7 because we think it's awful". But it is *not* correct to say that it is unsupported (until October anyway).

ekohl:
I can see the case for supporting Debian Stretch's 4.8.2 version but as I understand it your master needs to be 4.10.x for hiera, not the client. That means you only need to upgrade your master(s). Is that correct?

Either way, it's up to the repository maintainers to set the minimum version and having experience maintaining a decent number of modules I can tell you that it's a decent chunk of work. Keeping the supported matrix small does help.

trevor-vaughan:
@jcharaoui Since this doesn't seem to be going much of anywhere, the SIMP project (https://github.com/simp) will be maintaining a 4.7+ compatible fork until the full EOL of Puppet 4 https://github.com/simp/puppet-systemd/tree/simp-master

trevor-vaughan:
@dhollinger We don't have much of a choice at this time. We moved to data in modules as soon as we could and got caught in the Hiera 4 vs 5 rollover fun. Honestly, we don't **want** to maintain this version and will be switching back as soon as we can but we can't just force everyone to upgrade that uses the framework.

bastelfreak:
I'm going to close this since @trevor-vaughan provides a fork that uses the older hiera.

PR #76, opened 2018-05-08
Allow 'unlimited' and 'infinity' for LimitNOFILE

wyardley:
I think this is an improvement (thought this had already been fixed). However, there may still be some valid values which can't be set this way.


wyardley:
cc: @bastelfreak: could you review this?

wyardley:
#63 is probably a better fix.

PR #78, opened 2018-07-18
Fix CHANGELOG.md duplicate footer

alexjfisher:


alexjfisher:
By fixing HISTORY.md, CHANGELOG.md will no longer be generated with a repeated footer.

I'm not entirely sure why HISTORY.md exists at all though (or at least why it's been added to since it was originally created).  I assume that generating CHANGELOG.md without a HISTORY.md resulted in a lot of messy entries for older releases.  Cleaning these up manually and moving them into HISTORY.md is a lot less work (and spam) than massaging PRs and Issues that were closed years ago.

PR #80, opened 2018-07-24
Handle ensuring service_limits to be absent

ekohl:
When ensuring it to be absent, the limits can be undefined since they
don't matter.


bastelfreak:
Thanks @ekohl !

PR #82, opened 2018-08-20
Lint: include classes by absolute name

raphink:


manifests/dropin_file.pp
```diff
@@ -36,7 +36,7 @@
   Optional[String]                  $source   = undef,
   Optional[Stdlib::Absolutepath]    $target   = undef,
 ) {
-  include systemd
+  include ::systemd
```
bastelfreak: I think that is legacy syntax at should be avoided?

bastelfreak: We spent a huge amount of time on Vox Pupuli (and Puppet Inc modules) to get rid of the trailing colons

raphink: Hehe ok… I guess it's safe to deactivate that puppet-lint module then. @mcanevet your opinion?

mcanevet: I'm OK with that

ekohl: I have noticed that my puppet-lint in vim (called through syntastic) often does pick up the puppet-lint module even though that particular puppet module doesn't include it. Probably because it uses the system wide puppet-lint rather than through bundler. That means I get false positives. It's not possible to ignore a check for a module that isn't installed.

```
$ grep absolute Gemfile
  #gem 'puppet-lint-absolute_classname-check',              :require => false
$ grep no-relative .puppet-lint.rc 
--no-relative_classname_inclusion-check
$ bundle exec rake lint
invalid option: --no-relative_classname_inclusion-check
```

That makes me wonder what the cleanest path forward is. Include the lint module just to disable it is also very ugly.

bastelfreak: My workaround was to uninstall the system-wide puppet-lint. But that is ugly as well

ekohl: Then it's a good thing it doesn't parse project settings :) It just calls puppet-lint if the filetype is puppet. Because puppet-lint finds the gem installed, it uses it. Perhaps it should first try bundle exec puppet-lint. Note that puppet-lint exists globally because bundler or rubygems creates binstubs in ~/bin.

bastelfreak:
Also a quote from the plugins README.md:
> Including a class by a relative name might lead to unexpected results in Puppet 3. This plugin is not recommended for use with Puppet code that has dropped support for Puppet 3 (EOL 20161231).

https://github.com/voxpupuli/puppet-lint-absolute_classname-check#relative-class-name-inclusion

manifests/dropin_file.pp
bastelfreak: so in my opinion everything is fine when we remove the gem from the Gemfile? This is really legacy and should not be installed in the global ruby env, nor in any project.

bastelfreak:
Hi people, I had to merge https://github.com/camptocamp/puppet-systemd/pull/83 to fix modules that pull in this one as a dependency. I had to remove the gem dependency to get travis green.

manifests/dropin_file.pp
raphink: @ekohl I've often used `bundle exec vim` to use the `puppet-lint` from the bundle.

ekohl: @raphink that's a good suggestion

PR #87, opened 2018-09-20
Use _cache variable as intended

aquister:


manifests/resolved.pp
```diff
@@ -181,7 +181,7 @@
     false   => 'no',
   }
 
-  if $cache {
+  if $_cache {
```
bastelfreak: Does this change actually make sense? `$cache` is a boolean, `$_cache` is a string.

aquister: Well. If you set `$cache = false`. The ini_setting is not executed and `cache = no` is not set.

bastelfreak:
I think the original pattern here was to enable/disable the management of the resource. Now it would be used to set a value for the resource. Maybe an additional parameter like `$manage_cache`? Or should it allow more datatypes? Boolean false means we don't manage the resource, if it's a string the resource will be managed with the string as value.

PR #89, opened 2018-10-17
Manage journald service and configuration

treydock:


manifests/init.pp
```diff
@@ -91,6 +98,8 @@
   Boolean                                                $manage_accounting,
   Hash[String,String]                                    $accounting,
   Boolean                                                $purge_dropin_dirs,
+  Boolean                                                $manage_journald,
+  Hash[String, String]                                   $journald_config,
```
bastelfreak:
```suggestion
  Hash[String[1], String[1]]                                   $journald_config,
```

treydock: Updated.

trevor-vaughan:
No type validation? Are the `journald` options just too varied?

bastelfreak:
Hey @treydock, thanks for the work! I did an inline suggestion. Can you also document it in the README.md?

treydock:
@trevor-vaughan I suppose a custom type where the validation is done on a per-key basis would make sense, just not sure how it would survive for varied versions of systemd.  This would require subtypes to handle the varied units that are allowed for different keys.

@bastelfreak Updated README.

trevor-vaughan:
@treydock Yeah, I get it. The config hashes just always feel like a 'slightly better than just flopping down raw content' approach to life without some sort of validation :-|

raphink:
Agreed @trevor-vaughan. I usually like to provide a Data Type for the hash options to limit the scope of input data.

treydock:
I'll add a custom data type.

treydock:
Added custom data type but looks like #14 did same thing as this pull request.  Should I just make this pull request the addition of data type?

raphink:
Ah right, I'm very sorry about that @treydock 	I hadn't noticed… Sure, the data type would still be a nice addition.

ekohl:
https://rspec-puppet.com/documentation/type_aliases/ can be used to test type aliases for various input.

raphink:
Very nice!

treydock:
New data type added

PR #93, opened 2018-10-23
Add acceptance tests

raphink:


.travis.yml
```diff
@@ -17,6 +17,34 @@ matrix:
     env: PUPPET_GEM_VERSION="~> 4.0"
   - rvm: 2.4.1
     env: PUPPET_GEM_VERSION="~> 5.0"
+  - rvm: default
+    sudo: required
+    dist: trusty
+    services: docker
+    env: BEAKER_set="debian-8"
+    bundler_args:
+    script: sudo service docker restart ; sleep 10 && bundle exec rspec spec/acceptance/*_spec.rb
```
ekohl: Any particular reason to use this over `bundle exec rake beaker`? Also wondering why the docker restart and sleep is needed. We run docker tests in many repos without it.

raphink: We've been running things this way for years on other modules. We had problems with docker in Travis a few years ago so we added the restart + sleep and it fixed it, but it might not be required anymore. As for `rake beaker`, I was not aware of it; maybe it didn't exist when we started doing that.

spec/acceptance/nodesets/debian-9.yml
```diff
@@ -0,0 +1,15 @@
+HOSTS:
```
ekohl: In Voxpupuli and Theforeman we generate this on the fly: `PUPPET_INSTALL_TYPE=agent BEAKER_IS_PE=no BEAKER_PUPPET_COLLECTION=puppet5 BEAKER_debug=true BEAKER_setfile=centos7-64{hypervisor=docker\,hostname=centos7-64.example.com} bundle exec rake beaker`

You can use `debian9-64` instead of `centos7-64` or `ubuntu1604-64`. The hostname isn't needed but some applications prefer a FQDN instead of a short name.

.travis.yml
ekohl: That's very likely :)

spec/acceptance/nodesets/debian-9.yml
raphink: Ah, I wasn't aware of that possibility, that's very nice.

ekohl: I added it to beaker-hostgenerator in https://github.com/puppetlabs/beaker-hostgenerator/commit/5adf569c1a0d8cf1071c53ccda7bea0fbecde30a so the various projects could easily reuse work from others.

mcanevet: @ekohl nice work, I'm testing it right now.

PR #94, opened 2018-12-02
Allow specifying owner/group/mode/show_diff

simondeziel:
Some systemd files could contain sensitive information like .netdev
files associated with wireguard devices that contain private keys
and potentially pre-shared keys.


raphink:
LGTM (except for "wether", a detail)

PR #95, opened 2018-12-13
added function to returns a file system path as escaped by systemd

KoenDierckx:
While creating a systemd mount and automount unit file, I needed a way to convert a mount point into a systemd unit name, as this requires some very specific conventions
https://www.freedesktop.org/software/systemd/man/systemd.unit.html#String%20Escaping%20for%20Inclusion%20in%20Unit%20Names

Instead of recreating the logic in puppet, I've created a custom function that calls the native systemd-escape binary.

This can be used like this:
`$mount_name = systemd_escape($mountpoint)`
`systemd::unit_file { "${mount_name}.mount": }`

### DIFF
--- a/lib/puppet/parser/functions/systemd_escape.rb
+++ b/lib/puppet/parser/functions/systemd_escape.rb
@@ -1,0 +1,15 @@
+require 'puppet/util/execution'
+
+module Puppet::Parser::Functions
+  newfunction(:systemd_escape, :type => :rvalue, :doc => <<-EOS
+  Returns a file system path as escaped by systemd.
+  https://www.freedesktop.org/software/systemd/man/systemd.unit.html#String%20Escaping%20for%20Inclusion%20in%20Unit%20Names
+  EOS
+  ) do |args|
+    raise Puppet::ParseError, ("validate_cmd(): wrong number of arguments (#{args.length}; must be 1)") if args.length != 1
+    path = args[0]
+    cmd = "/bin/systemd-escape --path #{path}"
+    escaped = Puppet::Util::Execution.execute(cmd)
+    escaped.strip
+  end
+end


bastelfreak:
Hi @KoenDierckx, thanks for providing this PR. Can you please convert his to the modern functions API? Some docs are available at https://puppet.com/docs/puppet/4.9/functions_basics.html

alexjfisher:
Use of the modern API is needed for correct ‘environment isolation’.  With the new API, you can also namespace functions. This is recommended for all functions that aren’t built into puppet.
https://github.com/voxpupuli/puppet-extlib has lots of examples, including name spacing, testing and use of puppet-strings.

lib/puppet/parser/functions/systemd_escape.rb
```diff
@@ -0,0 +1,15 @@
...
+
+module Puppet::Parser::Functions
+  newfunction(:systemd_escape, :type => :rvalue, :doc => <<-EOS
+  Returns a file system path as escaped by systemd.
+  https://www.freedesktop.org/software/systemd/man/systemd.unit.html#String%20Escaping%20for%20Inclusion%20in%20Unit%20Names
+  EOS
+  ) do |args|
+    raise Puppet::ParseError, ("validate_cmd(): wrong number of arguments (#{args.length}; must be 1)") if args.length != 1
+    path = args[0]
+    cmd = "/bin/systemd-escape --path #{path}"
```
alexjfisher: The function will be run on the master, not the agents.  Could this be a problem?  (Perhaps the server isn't a systemd based system and doesn't have this command available).

raphink: Arguably, Linux OSes without systemd are getting more and more rare these days…

lib/puppet/parser/functions/systemd_escape.rb
```diff
@@ -0,0 +1,15 @@
+require 'puppet/util/execution'
+
+module Puppet::Parser::Functions
+  newfunction(:systemd_escape, :type => :rvalue, :doc => <<-EOS
+  Returns a file system path as escaped by systemd.
+  https://www.freedesktop.org/software/systemd/man/systemd.unit.html#String%20Escaping%20for%20Inclusion%20in%20Unit%20Names
+  EOS
+  ) do |args|
+    raise Puppet::ParseError, ("validate_cmd(): wrong number of arguments (#{args.length}; must be 1)") if args.length != 1
```
raphink: I'm guessing this was copied from the `validate_cmd` function code. It should be adapted to reflect the name of the function.

lib/puppet/parser/functions/systemd_escape.rb
alexjfisher: Probably ok then (so long as they all put that binary under `/bin`)

bastelfreak:
```
$ which systemd-escape
/usr/bin/systemd-escape
$ facter  os
{
  architecture => "x86_64",
  distro => {
    codename => "n/a",
    description => "Arch Linux",
    id => "Arch",
    release => {
      full => "rolling",
      major => "rolling"
    },
    specification => "1.4"
  },
  family => "Archlinux",
  hardware => "x86_64",
  name => "Archlinux",
  release => {
    full => "4.19.10-arch1-1-ARCH",
    major => "4",
    minor => "19"
  },
  selinux => {
    enabled => false
  }
}
```

raphink: Systemd: the unique layer to standardize everything on Linux OSes :smile:

lib/puppet/parser/functions/systemd_escape.rb
bastelfreak: nitpick: It isn't the fault from the systemd team:
* https://www.archlinux.org/news/binaries-move-to-usrbin-requiring-update-intervention/
* https://lists.archlinux.org/pipermail/arch-dev-public/2012-March/022625.html

lib/puppet/parser/functions/systemd_escape.rb
bastelfreak: nitpick: It isn't the fault from the systemd team:
* https://www.archlinux.org/news/binaries-move-to-usrbin-requiring-update-intervention/
* https://lists.archlinux.org/pipermail/arch-dev-public/2012-March/022625.html

lib/puppet/parser/functions/systemd_escape.rb
```diff
@@ -0,0 +1,15 @@
...
+    path = args[0]
+    cmd = "/bin/systemd-escape --path #{path}"
+    escaped = Puppet::Util::Execution.execute(cmd)
+    escaped.strip
+  end
+end
```
bastelfreak: please add the missing newline here.

lib/puppet/parser/functions/systemd_escape.rb
bastelfreak: Do we actually need to provide an absolute path? I thought puppet is able to figure this out by itself?

raphink: We could use `/usr/bin/env`?

towo:
@KoenDierckx Hey there, could you finish up this PR?

traylenator:
New attempt #220

PR #98, opened 2019-03-12
add missing systemd fact check for systemd::journald

Nachtmerrie:
Version 2.2.0 is no longer compatible with non-systemd systems. The addition of journald which I look forward to implementing for our farm does not have a check if the system has systemd.

This PR adds that check and makes the puppet-systemd module compatible again with non-systemd systems.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -117,6 +117,7 @@ class systemd (
   if $manage_accounting {
     contain systemd::system
   }
-
-  contain systemd::journald
+  if $facts['systemd_internal_services'] {
+    contain systemd::journald
+  }
 }


ekohl:
It looks correct, but when do you include the systemd class on non-systemd systems?

Nachtmerrie:
We include the module globally. 5% of our servers do not have systemd.

The systemd_timesyncd, systemd_networkd and systemd_resolved also have this check:
https://github.com/camptocamp/puppet-systemd/blob/master/manifests/init.pp#L105

It seems it was forgotten for the new journald option.

--- 
Daniël Vos

Nachtmerrie:
The check should probably also be there for the ` $manage_accounting` option. I can include this in my PR if you want.

ekohl:
That's because those services are not always present. journald is always present with systemd. I'm wondering if this module was designed but I'll leave that up to the module maintainers.

rj667:
> It looks correct, but when do you include the systemd class on non-systemd systems?

How would I know if a system has systemd or not, without the facts from this class to tell me?

ekohl:
Facts are not created by the class but by facter. This module adds a custom fact:

https://github.com/camptocamp/puppet-systemd/blob/4d97305fed6d9d6e51ab2468cd4e9f26ecc72237/lib/facter/systemd.rb#L36-L41

In your profile you can use this:
```puppet
if $facts['systemd'] {
  include systemd
}
```

rj667:
good point, thanks!

raphink:
I'm not sure about merging this, as this will be a breaking change that can be avoided as explained by @ekohl in https://github.com/camptocamp/puppet-systemd/pull/98#issuecomment-473332814

Nachtmerrie:
PR https://github.com/camptocamp/puppet-systemd/pull/99 solves my problem, thank you @dhoppe. Closing this now.

PR #99, opened 2019-04-10
Add parameter to enable/disable the management of journald

dhoppe:


dhoppe:
This might solve #98 aswell.

manifests/init.pp
```diff
@@ -95,6 +98,7 @@
   Boolean                                                $manage_accounting,
   Hash[String,String]                                    $accounting,
   Boolean                                                $purge_dropin_dirs,
+  Boolean                                                $manage_journald,
```
bastelfreak: can you please add a unit test that sets this to false?

PR #111, opened 2019-06-07
Allow for lazy/eager systemctl daemon reloading

JohnLyman:
Add param for `systemd::dropin_file` to allow for `lazy` (the default behavior) or `eager` systemctl daemon reloading.

In some cases, the systemctl daemon should be reloaded before the related service is refreshed. For example, when overriding `ExecStart`. With the current behavior, it is impossible to do so if the `systemd::dropin_file` and related service are contained in the same class. There is also a potential for circular dependencies with multiple services that depend on each other and have multiple drop-in files associated with them.

The new lazy option emulates the current behavior, deferring the reload until (most likely) near the end of the Puppet run. It also has the advantage of only reloading the daemon once.

The eager option allows for the daemon to be reloaded immediately after changes to the specified drop-in file, at the cost of causing multiple reloads if there are multiple `systemd::dropin_file` resources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,12 @@ service {'foo':
 }
 ```
 
+Sometimes it's desirable to reload the systemctl daemon before a service is refreshed (for example:
+when overriding `ExecStart` or adding environment variables to the drop-in file).  In that case,
+use `daemon_reload => 'eager'` instead of the default `'lazy'`.  Be aware that the daemon could be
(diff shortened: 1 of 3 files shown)


bastelfreak:
Hi, thanks for the PR!

alexjfisher:
Adding comment to force github into reindexing this PR (so it gets included in Changelogs).  Sorry for the noise! :(

PR #115, opened 2019-07-10
allow Sensitive type for content param

### LINKED ISSUE #114: Allow Sensitive type for systemd::dropin_file::content
Since content this is just passed through to a file resource allow to pass Sensitive values also

TheMeier:
fixes #114

### DIFF
--- a/manifests/dropin_file.pp
+++ b/manifests/dropin_file.pp
@@ -47,18 +47,18 @@
 #   May cause multiple daemon reloads.
 #
 define systemd::dropin_file(
-  Systemd::Unit                     $unit,
-  Systemd::Dropin                   $filename      = $name,
-  Enum['present', 'absent', 'file'] $ensure        = 'present',
-  Stdlib::Absolutepath              $path          = '/etc/systemd/system',
-  Optional[String]                  $content       = undef,
-  Optional[String]                  $source        = undef,
-  Optional[Stdlib::Absolutepath]    $target        = undef,
-  String                            $owner         = 'root',
-  String                            $group         = 'root',
-  String                            $mode          = '0444',
-  Boolean                           $show_diff     = true,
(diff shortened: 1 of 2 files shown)


manifests/dropin_file.pp
```diff
@@ -47,18 +47,18 @@
...
-  String                            $owner         = 'root',
-  String                            $group         = 'root',
-  String                            $mode          = '0444',
-  Boolean                           $show_diff     = true,
-  Enum['lazy', 'eager']             $daemon_reload = 'lazy',
+  Systemd::Unit                               $unit,
+  Systemd::Dropin                             $filename      = $name,
+  Enum['present', 'absent', 'file']           $ensure        = 'present',
+  Stdlib::Absolutepath                        $path          = '/etc/systemd/system',
+  Optional[Variant[String,Sensitive[String]]] $content       = undef,
```
raphink: It might be clearer to make that a `Variant[String,Sensitive[String],Undef]`

TheMeier: @raphink hm IMHO `Optional` is the more idomatic way. Also grepping over a couple of modules i have checked out here from various sources it seems `Optional[Variant]` is what is usually used:

```
$ find . -name "*.pp" |xargs grep -E "Variant.*Undef" | wc -l
40
$ find . -name "*.pp" |xargs grep -E "Optional\[Variant" | wc -l
492
```

bastelfreak: I go with @TheMeier here. The default should be `undef` and that requires the type to be `Optional`

raphink: @bastelfreak 	technically it doesn't require to use `Optional`, as a variant allowing `Undef` works as well.

vchepkov:
I think it should be `Optional[Variant[String,Sensitive]]`
Sensitive is type on it's own

Issue #118, opened 2019-09-20
add timer support


micah:
systemd provides a cron-like system replacement, through the use of unit configuration files that end in ".timer". These unit configurations encode information about timers controlled and supervised by systemd, for timer-based activation.

It would be excellent if this module provided a mechanism for managing these!

bastelfreak:
Hi,
this already works fine:
https://github.com/voxpupuli/puppet-borg/blob/f2c66b51c46d57b25326524415c0e9ced683a469/manifests/service.pp#L11-L26

https://github.com/voxpupuli/puppet-borg/blob/f2c66b51c46d57b25326524415c0e9ced683a469/templates/borg-backup.timer.epp#L1-L13

trevor-vaughan:
@bastelfreak I agree that it works, but it's confusing for end users. It would be nice to have a defined type or something that just wraps it to be more clear.

traylenator:
I was just contemplating writing  a systemd::timer  type , will try something.

traylenator:
The only interface that makes sense would be

```puppet
systemd::timer{'runoften.timer':
    timer_content   => ...
    service_content =>
    enable => true,
    active => true,
}
```

Essentially to specify the content of the service and timer unit.

It does not make sense to specify the individual say `OnCalandar` arguments since this gets tricky because you then have to define the Before, , After, multiple OnCalendar, ...

The alternative is to only support really trivial jobs. Such as 

```puppet
systemd::timer{'myjob':
  exec => 'rm -rf /',
  on_calendar => 'daily',
}
```

along with some other timer options. Anything more complicated would require to resort back to two `systemd::unit_files` as per the above examples.

So is the latter one still worth while.

trevor-vaughan:
@traylenator I think that it is. I also think that a wrapper that creates the rest of the bits for a generally useful `OnCalendar` example would be useful but could be a later addition.

raphink:
Implemented in #138

PR #119, opened 2019-09-26
New systemd::timer define type

traylenator:
The defined type allows easy creation of the timer and service unit at the same time.

Examples of usage

```puppet
systemd::timer{'runoften.timer':
  timer_source   => "puppet:///modules/${module_name}/runoften.timer",
  service_source => "puppet:///modules/${module_name}/runoften.service",
}
```
or a trivial daily run as:

```puppet
systemd::timer{'daily.timer':
  timer_content  => "[Timer]\nOnCalendar=daily\nRandomizedDelaySec=1d\n",
  service_content => "[Service]\nType=oneshot\nExecStart=/usr/bin/touch /tmp/file",
}
```

If neither `service_content` or `service_source` are specified then no service unit will be created.

The service unit name can also be specified.

```puppet
systemd::timer{'daily.timer':
  timer_content  => "[Timer]\nOnCalendar=daily\nRandomizedDelaySec=1d\nUnit=touch-me-today.service",
  service_unit   => 'touch-me-today.service',
  service_content => "[Service]\nType=oneshot\nExecStart=/usr/bin/touch /tmp/file",
}
```

Fixes: #118

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,82 @@ file { '/etc/tmpfiles.d/foo.conf':
 ~> Class['systemd::tmpfiles']
 ```
 
+### timer units
+Create a systemd timer unit and a systemd service unit to execute from
+that timer 
+
+The following will create a timer unit and a service unit file.
+The execution of `systemctl daemon-reload` will occur.
+As active and enable are set true the puppet service `runoften.timer` will be 
+declared and be started and enabled.
+
+```puppet
+systemd::timer{'runoften.timer':
+  timer_source   => "puppet:///modules/${module_name}/runoften.timer",
+  service_source => "puppet:///modules/${module_name}/runoften.service",
+  active         => true,
+  enable         => true,
+}
+```
+
+A trivial daily run.
+In this case enable and active are both unset and so the service `daily.timer` 
+is not declared by the `systemd::timer` type.
+
+```puppet
+$_timer = @(EOT)
+[Timer]
+OnCalendar=daily
+RandomizedDelaySec=1d
+EOT
+
+$_service = @(EOT)
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/touch /tmp/file
+EOT
+
+systemd::timer{'daily.timer':
+  timer_content   => $_timer,
+  service_content => $_service,
+}
+
+service{'daily.timer':
+  ensure    => running,
+  subscribe => Systemd::Timer['daily.timer'],
+}
+
+```
+
+If neither `service_content` or `service_source` are specified then no
+service unit will be created.
+
+The service unit name can also be specified.
+
+```puppet
+$_timer = @(EOT)
+[Timer]
+OnCalendar=daily
+RandomizedDelaySec=1d
+Unit=touch-me-today.service
+EOT
+
+$_service = @(EOT)
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/touch /tmp/file
+EOT
+
+
+systemd::timer{'daily.timer':
+  timer_content   => $_timer,
+  service_unit    => 'touch-me-today.service',
+  service_content => $_service,
+  active          => true,
+  enable          => true,
+} 
+```
+
 ### service limits
 
 Manage soft and hard limits on various resources for executed processes.
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ that timer
 The following will create a timer unit and a service unit file.
 The execution of `systemctl daemon-reload` will occur.
 When `active` and `enable` are set to `true` the puppet service `runoften.timer` will be 
-declared and be started and enabled.
+declared, started and enabled.
 
 ```puppet
 systemd::timer{'runoften.timer':
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ that timer
 
 The following will create a timer unit and a service unit file.
 The execution of `systemctl daemon-reload` will occur.
-As active and enable are set true the puppet service `runoften.timer` will be 
+When `active` and `enable` are set to `true` the puppet service `runoften.timer` will be 
 declared and be started and enabled.
 
 ```puppet
--- a/manifests/timer.pp
+++ b/manifests/timer.pp
@@ -1,0 +1,106 @@
+# @summary Create a timer and optionally a service unit to execute with the timer unit
+#
+# @api public
+#
+# @see https://www.freedesktop.org/software/systemd/man/systemd.timer.html systemd.timer(5)
+#
+# @param name [Pattern['^.+\.timer$]]
+#   The target of the timer unit to create
+#
+# @param path
+#   The main systemd configuration path
+#
+# @param timer_content
+#   The full content of the timer unit file
+#
+#   * Mutually exclusive with ``$timer_source``
+#
+# @param timer_source
+#   The ``File`` resource compatible ``source``
+#
+#   * Mutually exclusive with ``$timer_content``
+#
+# @param service_content
+#   The full content of the service unit file
+#
+#   * Mutually exclusive with ``$timer_source``
+#
+# @param service_source
+#   The ``File`` resource compatible ``source``
+#
+#   * Mutually exclusive with ``$service_content``
+#
+# @param owner
+#   The owner to set on the dropin file
+#
+# @param group
+#   The group to set on the dropin file
+#
+# @param mode
+#   The mode to set on the dropin file
+#
+# @param show_diff
+#   Whether to show the diff when updating dropin file
+#
+# @param service_unit
+#   If set then the service_unit will have this name.
+#   If not set the service unit has the same name
+#   as the timer unit with s/.timer/.service/
+#
+# @param active
+#  If set to true or false the timer service will be maintained.
+#  If true the timer service will be running and enabled, if false it will
+#  explictly stopped and disabled.
+#
+# @param enable
+#   If set, will manage the state of the unit.
+#
+define systemd::timer (
+  Enum['present', 'absent', 'file']        $ensure = 'present',
+  Stdlib::Absolutepath                     $path = '/etc/systemd/system',
+  Optional[String[1]]                      $timer_content = undef,
+  Optional[String[1]]                      $timer_source = undef,
+  Optional[String[1]]                      $service_content = undef,
+  Optional[String[1]]                      $service_source  = undef,
+  String[1]                                $owner = 'root',
+  String[1]                                $group = 'root',
+  Stdlib::Filemode                         $mode = '0444',
+  Optional[Systemd::Unit]                  $service_unit = undef,
+  Boolean                                  $show_diff = true,
+  Optional[Variant[Boolean, Enum['mask']]] $enable = undef,
+  Optional[Boolean]                        $active = undef,
+) {
+  assert_type(Pattern['^.+\.timer$'],$name)
+
+  if $service_unit {
+    $_service_unit = $service_unit
+  } else {
+    $_service_unit = "${basename($name,'.timer')}.service"
+  }
+
+  if $service_content or $service_source {
+    systemd::unit_file{$_service_unit:
+      ensure    => $ensure,
+      content   => $service_content,
+      source    => $service_source,
+      path      => $path,
+      owner     => $owner,
+      group     => $group,
+      mode      => $mode,
+      show_diff => $show_diff,
+    }
+  }
+
+  systemd::unit_file{$name:
+    ensure    => $ensure,
+    content   => $timer_content,
+    source    => $timer_source,
+    path      => $path,
+    owner     => $owner,
+    group     => $group,
+    mode      => $mode,
+    show_diff => $show_diff,
+    enable    => $enable,
+    active    => $active,
+  }
+}
--- a/manifests/timer.pp
+++ b/manifests/timer.pp
@@ -23,7 +23,7 @@
 # @param service_content
 #   The full content of the service unit file
 #
-#   * Mutually exclusive with ``$timer_source``
+#   * Mutually exclusive with ``$service_source``
 #
 # @param service_source
 #   The ``File`` resource compatible ``source``
--- a/spec/defines/timer_spec.rb
+++ b/spec/defines/timer_spec.rb
@@ -1,0 +1,73 @@
+require 'spec_helper'
+
+describe 'systemd::timer' do
+  context 'supported operating systems' do
+    on_supported_os.each do |os, facts|
+      context "on #{os}" do
+        let(:facts) { facts }
+
+        let(:title) { 'foobar.timer' }
+
+        context('with timer_content and service_content') do
+          let(:params) {{
+            :timer_content  => "[Timer]\nOnCalendar=weekly",
+            :service_content => "[Service]\nExecStart=/bin/touch /tmp/foobar"
+          }}
+
+          it { is_expected.to compile.with_all_deps }
+
+          it { is_expected.to contain_systemd__unit_file('foobar.timer').with(
+           :content => "[Timer]\nOnCalendar=weekly",
+          ) }
+
+          it { is_expected.to contain_systemd__unit_file('foobar.service').with(
+           :content => "[Service]\nExecStart=/bin/touch /tmp/foobar",
+          ) }
+        end
+
+        context('with timer_source and service_source') do
+          let(:params) {{
+            :timer_source  => "puppet:///timer",
+            :service_source => "puppet:///source",
+          }}
+
+          it { is_expected.to compile.with_all_deps }
+          it { is_expected.to contain_systemd__unit_file('foobar.timer').with_source('puppet:///timer') }
+          it { is_expected.to contain_systemd__unit_file('foobar.service').with_source('puppet:///source') }
+        end
+
+        context('with timer_source only set') do
+          let(:params) {{
+            :timer_source  => "puppet:///timer",
+          }}
+
+          it { is_expected.to compile.with_all_deps }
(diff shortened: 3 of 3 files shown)


manifests/timer.pp
```diff
@@ -0,0 +1,102 @@
...
+define systemd::timer (
+  Enum['present', 'absent', 'file']        $ensure    = 'present',
+  Stdlib::Absolutepath                     $path      = '/etc/systemd/system',
+  Optional[String]                         $timer_content   = undef,
+  Optional[String]                         $timer_source    = undef,
+  Optional[String]                         $service_content   = undef,
+  Optional[String]                         $service_source    = undef,
+  String                                   $owner     = 'root',
+  String                                   $group     = 'root',
+  String                                   $mode      = '0444',
```
bastelfreak: can you use https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/REFERENCE.md#stdlibfilemode ?

manifests/timer.pp
```diff
@@ -0,0 +1,102 @@
...
+#
+# @param service_unit
+#   If set then the service_unit will have this name.
+#   If not set the service unit has the same name
+#   as the timer unit with s/.timer/.service/
+#
+define systemd::timer (
+  Enum['present', 'absent', 'file']        $ensure    = 'present',
+  Stdlib::Absolutepath                     $path      = '/etc/systemd/system',
+  Optional[String]                         $timer_content   = undef,
```
bastelfreak: for all this strings, an empty string should be prohibited. Can you update them to use `String[1]`?

manifests/timer.pp
```diff
@@ -0,0 +1,102 @@
...
+
+  if $service_unit {
+    $_service_unit = $service_unit
+  } else {
+    $_service_unit = "${basename($name,'.timer')}.service"
+  }
+
+  contain systemd::systemctl::daemon_reload
+
+  if $service_content != undef or $service_source != undef {
```
bastelfreak: if you want, you can shorten this to:
```suggestion
  if $service_content or $service_source {
```

README.md
```diff
@@ -129,6 +129,38 @@ file { '/etc/tmpfiles.d/foo.conf':
 ~> Class['systemd::tmpfiles']
 ```
 
+### timer units
+Create a timer unit and a service unit to execute from
+that timer at the same time.
+
+```puppet
+systemd::timer{'runoften.timer':
```
ekohl: Should the timer be automatically added? It's a bit redundant.

traylenator: @ekohl I'm not sure what you mean here?

ekohl: Should it accept `systemd::timer{'runoften':` and in timer.pp add `.timer` internally?

traylenator: Got you, that `.timer` I went through this myself. I thought it was more consistent with `systemd::unit_file` for instance.

Certainly it would remove all that basename dancing.

traylenator: I think I prefer it with the .timer forced, it makes it consistent with
the `service_unit` parameter as well.

traylenator:
Of course there is another unrelated exercise to write a systemd provider type for the existing cron type.

manifests/timer.pp
```diff
@@ -0,0 +1,102 @@
...
+    systemd::unit_file{$_service_unit:
+      ensure    => $ensure,
+      content   => $service_content,
+      source    => $service_source,
+      path      => $path,
+      owner     => $owner,
+      group     => $group,
+      mode      => $mode,
+      show_diff => $show_diff,
+      before    => Systemd::Unit_file[$name],
```
ekohl: When it's absent, shouldn't this be the other way around?

manifests/timer.pp
```diff
@@ -0,0 +1,102 @@
...
+) {
+  assert_type(Pattern['^.+\.timer$'],$name)
+
+  if $service_unit {
+    $_service_unit = $service_unit
+  } else {
+    $_service_unit = "${basename($name,'.timer')}.service"
+  }
+
+  contain systemd::systemctl::daemon_reload
```
ekohl: Doesn't the unit file handle this already?

manifests/timer.pp
traylenator: Technically yes , when enable it's  important since a timer will fail to start if the service does not exist.

It should be  for enable

* create unit files
* run daemon-reload
* enable/start units

as for disable

* stop/disable units 
* remove files
* daemon-reload

so in fact the relation ship betwee the creation of the unit files is irreverent and this can go completely in fact.

manifests/timer.pp
traylenator: Not quite in my experience the order is not correct and deamon-reload often happens after the service is 
managed, looking a bit harder now. `unit_file` contains:

```puppet
   include systemd    # which contains  daemon_reload'
     file { "${path}/${name}":
     ensure    => $_ensure,
     show_diff => $show_diff,
     notify    => Class['systemd::systemctl::daemon_reload'],
  }
service { $name:
  ensure    => $active,
  subscribe => File["${path}/${name}"],
  require   => Class['systemd::systemctl::daemon_reload'],
}

```
there is nothing in there that forces the exec of daemon-reload to happen before the service is managed.

Probably if unit_file itself contained `systemd::systemctl::daemon_reload`  or prossibly `contain systemd` `it might work.

ekohl: > there is nothing in there that forces the exec of daemon-reload to happen before the service is managed.

The `service` has a `require => Class['systemd::systemctl::daemon_reload']`. Doesn't that take care of it?

traylenator: No it does not. 

The class needs to be contained where the require is done. 
With include class you are saying I require that class be included before.
With contain class you are saying I require that class be included and do everything inside that class.

`require -> Class['abc']`  on its own is almost always not what you want with out a contain (or anchor)

traylenator: Another MR s/include/contain/ problably thing to do. Can do in a week or two ....

ekohl: That's not my understanding of how contain works. AFAIK `contain x` in `class a` it means that if I `require => Class['a']` I'll be sure that `class x` is executed before. However, the service still requires it and within the catalog it should be ordered correctly.

That said, with `include` and just `notify => Class['daemon_reload']` you can't be sure with just `require => Systemd::Unit_file['...']` that it has reloaded.

alexjfisher: I think the `contain` here is likely to cause circular dependencies when you have eg two systemd::timer resources in different profiles and you try to also have an order between those profiles.

mmoll:
what's the status here?

bastelfreak:
@traylenator can you take a look at the above conversations?

manifests/timer.pp
traylenator: contain is gone.  relying on the include inside unit_file.

traylenator:
I believe all comments are addressed.

alexjfisher:
Regarding `systemd::systemctl::daemon_reload` I think (in a separate PR), the docs need tweaking a bit.
in unit_file.pp a [relationship](https://github.com/camptocamp/puppet-systemd/blob/f81339ad10516905e460d2899432ce92d83a35d2/manifests/unit_file.pp#L102) is created, but *only* when [enable or active](https://github.com/camptocamp/puppet-systemd/blob/f81339ad10516905e460d2899432ce92d83a35d2/manifests/unit_file.pp#L89) are used.  If you are managing the service resource yourself, you have to create this relationship yourself.

Also `systemd::systemctl::daemon_reload` is only relevant for puppet 5.  Since puppet 6.1.0, the systemd service provider in core puppet handles the reload itself.
Also see https://github.com/voxpupuli/puppet-prometheus/pull/390/files#diff-0ac866172efd29d0ddc2aa75ae6f182dR188-R192 for an example.

manifests/timer.pp
```diff
@@ -0,0 +1,99 @@
+# Create a timer and optionally a service unit
+# to execute with the timer unit.
+#
+# @api public
+#
+# @see systemd.timer
```
alexjfisher: Was this just a placeholder? Is this supposed to link to an external website?

manifests/timer.pp
```diff
@@ -0,0 +1,99 @@
...
+  Optional[String[1]]                      $timer_content = undef,
+  Optional[String[1]]                      $timer_source = undef,
+  Optional[String[1]]                      $service_content = undef,
+  Optional[String[1]]                      $service_source  = undef,
+  String[1]                                $owner = 'root',
+  String[1]                                $group = 'root',
+  Stdlib::Filemode                         $mode = '0444',
+  Optional[Systemd::Unit]                  $service_unit = undef,
+  Boolean                                  $show_diff = true,
+  Optional[Variant[Boolean, Enum['mask']]] $enable = undef,
```
alexjfisher: `enable` and `active` aren't documented above.  If either are set a service resource will be [created](https://github.com/camptocamp/puppet-systemd/blob/f81339ad10516905e460d2899432ce92d83a35d2/manifests/unit_file.pp#L90).  Is this correct?

manifests/timer.pp
traylenator: I copied it from the other types:

```
tmpfile.pp:# @see systemd-tmpfiles(8)
tmpfiles.pp:# @see systemd-tmpfiles(8)
unit_file.pp:# @see systemd.unit(5)
```

I added the (5) now.

Now if they she be there at all is another matter.

traylenator:
I improved the examples to document both manual and and automatic cases for managing the service.

traylenator:
For the puppet > 6.1.4 case I think this is just a documentation fix which I have done.
I am not including the `systemd::systemctl::daemon_reload` anywhere anymore in this file at least.

manifests/timer.pp
traylenator: @alexjfisher  remove it?

manifests/timer.pp
alexjfisher: Looks like this works.  Slight concern that this is by accident and the `service` resource type wasn't intended to manage timers.
For instance, `puppet resource service` doesn't return any timers.  See the provider's [`instance`](https://github.com/puppetlabs/puppet/blob/4fcf3d953a0467555fc5477a2e28b1d5fd6819ec/lib/puppet/provider/service/systemd.rb#L32) method.

alexjfisher: Less concerned now that I've read [this](https://github.com/puppetlabs/puppet/blob/4fcf3d953a0467555fc5477a2e28b1d5fd6819ec/lib/puppet/provider/service/systemd.rb#L9).

manifests/timer.pp
alexjfisher: Or link to eg. https://www.freedesktop.org/software/systemd/man/systemd.timer.html ??

README.md
```diff
@@ -129,6 +129,80 @@ file { '/etc/tmpfiles.d/foo.conf':
...
+RandomizedDelaySec=1d
+EOT
+
+$_service = @(EOT)
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/touch /tmp/file
+EOT
+
+if versioncmp($facts['puppetversion'],'6.1.0') < 0 {
```
ekohl: When using `enable` and/or `active` I think that it already takes care of this, right? Might be good to explain this to the user.

README.md
```diff
@@ -129,6 +129,80 @@ file { '/etc/tmpfiles.d/foo.conf':
...
+  include systemd::systemctl::daemon_reload
```
alexjfisher: This will always end up included by the main class (which is included by unit_file).
https://github.com/camptocamp/puppet-systemd/blob/f81339ad10516905e460d2899432ce92d83a35d2/manifests/init.pp#L113

README.md
```diff
@@ -129,6 +129,80 @@ file { '/etc/tmpfiles.d/foo.conf':
...
+  Systemd::Timer['daily.timer'] ~> Class[Systemd::Systemctl::Daemon_reload]
```
alexjfisher: This relationship already exists??
Should be
```
Class[Systemd::Systemctl::Daemon_reload] -> Service['daily.timer']
```
???

README.md
alexjfisher: Yes, I think so.  This adds a `service` resource and the required (for puppet 5) relationship between daemon reload and the service.

manifests/timer.pp
```diff
@@ -0,0 +1,107 @@
+# Create a timer and optionally a service unit
```
bastelfreak: this should be `# @summary Create a timer and optionally a service unit to execute with the timer unit`?

README.md
traylenator: Seems bad to notify things inside the unit_file type , assumes it's contents.

alexjfisher: Not sure I follow. https://github.com/camptocamp/puppet-systemd/blob/f81339ad10516905e460d2899432ce92d83a35d2/manifests/unit_file.pp#L86 is the relationship I think you're duplicating.

alexjfisher: Actually, I think it's
```
Systemd::Timer['daily.timer'] ~> Service['daily.timer']
Class[Systemd::Systemctl::Daemon_reload] ~> Service['daily.timer']
```
that would be needed.

Perhaps systemd::timer can always make sure a service resource is declared itself? (which I think would be ok for timer units created with this module)
```puppet
unless $enable or $active {
  # if either $enable or $active are set, unit_file will create a service resource and 
  service { $name:
    provider  => systemd,
    subscribe => [
      Class[Systemd::Systemctl::Daemon_reload],
      Systemd::Unit_file[$name],
    ],
  }
}
```
or
```puppet
service { $name:
  provider  => systemd,
  enable    => $enable,
  active    => $active,
  subscribe => [
    Class[Systemd::Systemctl::Daemon_reload],
    Systemd::Unit_file[$name],
  ],
}
systemd::unit_file{$name:
    # as before but active/enable parameters removed
    ensure    => $ensure,
    content   => $timer_content,
    source    => $timer_source,
    path      => $path,
    owner     => $owner,
    group     => $group,
    mode      => $mode,
    show_diff => $show_diff,
}
```

README.md
traylenator: Link added:

```
@see https://www.freedesktop.org/software/systemd/man/systemd.timer.html systemd.timer(5)
```
seems to be the correct syntax.

README.md
traylenator: > Not sure I follow.
was me confused.

Creating the service inside timer.pp seems a shame to not reuse unit_file code.

README.md
```diff
@@ -157,19 +164,14 @@ Type=oneshot
...
-
 systemd::timer{'daily.timer':
   timer_content   => $_timer,
   service_content => $_service,
 }
 
 service{'daily.timer':
   ensure => running,
-  require => Systemd::Timer['daily.timer'],
+  notify => Systemd::Timer['daily.timer'],
```
alexjfisher: ??? What does this do?

traylenator: I think it must do time travel.

Changed to subscribe - when the unit changes restart the service.

traylenator:
rebased and squashed

manifests/timer.pp
```diff
@@ -0,0 +1,106 @@
...
+#
+# @param timer_source
+#   The ``File`` resource compatible ``source``
+#
+#   * Mutually exclusive with ``$timer_content``
+#
+# @param service_content
+#   The full content of the service unit file
+#
+#   * Mutually exclusive with ``$timer_source``
```
ekohl:
```suggestion
#   * Mutually exclusive with ``$service_source``
```

manifests/timer.pp
```diff
@@ -0,0 +1,106 @@
...
+) {
+  assert_type(Pattern['^.+\.timer$'],$name)
+
+  if $service_unit {
+    $_service_unit = $service_unit
+  } else {
+    $_service_unit = "${basename($name,'.timer')}.service"
+  }
+
+  if $service_content or $service_source {
```
ekohl: Within this block I'd expect something like:
```puppet
if $ensure == 'absent' {
  Systemd::Unit_file[$name] -> Systemd::Unit_file[$_service_unit]
} else {
  Systemd::Unit_file[$_service_unit] -> Systemd::Unit_file[$name]
}
```

That way you know the service is available when enabling the timer. If something goes wrong with defining the service, the timer will not be present. When removing you should go in the other direction.

README.md
```diff
@@ -129,6 +129,82 @@ file { '/etc/tmpfiles.d/foo.conf':
 ~> Class['systemd::tmpfiles']
 ```
 
+### timer units
+Create a systemd timer unit and a systemd service unit to execute from
+that timer 
+
+The following will create a timer unit and a service unit file.
+The execution of `systemctl daemon-reload` will occur.
+As active and enable are set true the puppet service `runoften.timer` will be
```
ekohl:
```suggestion
When `active` and `enable` are set to `true` the puppet service `runoften.timer` will be 
```

README.md
```diff
@@ -129,6 +129,82 @@ file { '/etc/tmpfiles.d/foo.conf':
...
 ```
 
+### timer units
+Create a systemd timer unit and a systemd service unit to execute from
+that timer 
+
+The following will create a timer unit and a service unit file.
+The execution of `systemctl daemon-reload` will occur.
+As active and enable are set true the puppet service `runoften.timer` will be 
+declared and be started and enabled.
```
ekohl:
```suggestion
declared, started and enabled.
```

mmoll:
@traylenator please rebase :)

mmoll:
ping?

raphink:
@traylenator could you rebase your PR please?

mmoll:
GH-138

bastelfreak:
merged in https://github.com/camptocamp/puppet-systemd/pull/138

traylenator:
Many thanks @mmoll

PR #120, opened 2019-09-29
Manage logind service and configuration

fraenki:
Make it possible to manage logind settings. Example configuration:

```
systemd::manage_logind: true

systemd::logind_settings:
  HandleSuspendKey: 'ignore'
  KillUserProcesses: 'no'
  RemoveIPC: 'no'
  UserTasksMax: 50000
```

This will allow me (and others) to migrate from https://github.com/NTTCom-MS/eyp-systemd to this module :)

This is (to some degree) a continuation of the discussion in https://github.com/voxpupuli/puppet-confluence/pull/173#issuecomment-468970053. Ping @bastelfreak.

### DIFF
--- a/manifests/logind.pp
+++ b/manifests/logind.pp
@@ -1,0 +1,31 @@
+# **NOTE: THIS IS A [PRIVATE](https://github.com/puppetlabs/puppetlabs-stdlib#assert_private) CLASS**
+#
+# This class manages systemd's login manager configuration.
+#
+# https://www.freedesktop.org/software/systemd/man/logind.conf.html
+class systemd::logind {
+
+  assert_private()
+
+  service{'systemd-logind':
+    ensure => running,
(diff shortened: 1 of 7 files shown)


manifests/logind.pp
```diff
@@ -0,0 +1,31 @@
+# **NOTE: THIS IS A [PRIVATE](https://github.com/puppetlabs/puppetlabs-stdlib#assert_private) CLASS**
```
ekohl: Perhaps this should use `@api private` so puppet-strings can understand it?

fraenki: Sure thing. I just copied from existing classes in this module... fixed in https://github.com/camptocamp/puppet-systemd/pull/120/commits/ee36d4eb19de6376509e18836c478a41b0d69164.

PR #122, opened 2019-10-23
Correct order when ensuring unit files are absent

ekohl:
This ensures the service is stopped before the unit file is removed. Also does some other cleanups on the unit file.


bastelfreak:
Thanks @ekohl !

PR #125, opened 2019-11-01
Add SLES support

msurato:
Adding SLES support. I do not have a copy of OpenSuse or I would add that as well.


msurato:
I don't exactly know what happened on the squash rebase, but the only file that should be changed is the metadata.json.

PR #127, opened 2019-11-12
add option for persistent logging

djvl:


manifests/journald.pp
```diff
@@ -7,6 +7,19 @@
 
   assert_private()
 
+  if $::systemd::journald_persist_log {
```
bastelfreak: Please don't add `::` after `$`. That's legacy Puppet 3 syntax and not required anymore:
```suggestion
  if $systemd::journald_persist_log {
```

manifests/init.pp
```diff
@@ -106,6 +109,7 @@
   Boolean                                                $purge_dropin_dirs,
   Boolean                                                $manage_journald,
   Systemd::JournaldSettings                              $journald_settings,
+  Boolean                                                $journald_persist_log,
```
bastelfreak: you need to add a default value in `data/common.yaml`.

bastelfreak:
thanks for the PR!

bastelfreak:
I noticed that his is a backwards-incompatible change because it purges /var/log/journal. We should change the default from `absent` to `directory`

PR #128, opened 2019-12-04
Add OOMScoreAdjust to Systemd::ServiceLimits type

jlutran:
As per [systemd.exec](https://www.freedesktop.org/software/systemd/man/systemd.exec.html), ```OOMScoreAdjust``` takes an integer between -1000 and 1000.


bastelfreak:
Thanks for the PR @jlutran !

PR #129, opened 2019-12-05
Force tmpfiles.d drop file to end in .conf

traylenator:
From tmpfiles.d(5)

```
SYNOPSIS
  /etc/tmpfiles.d/*.conf
  /run/tmpfiles.d/*.conf
  /usr/lib/tmpfiles.d/*.conf
```

Just like unit drop in files drop files for the tmpfiles.d must
end in .conf to be noticed.

### DIFF
--- a/manifests/tmpfile.pp
+++ b/manifests/tmpfile.pp
@@ -4,10 +4,10 @@
 #
 # @see systemd-tmpfiles(8)
 #
-# @attr name [String]
+# @attr name [Pattern['^.+\.conf$']] (filename)
 #   The name of the tmpfile to create
 #
-#   * May not contain ``/``
+#   * May not contain ``/`` and must end in .conf
 #
 # @param $ensure
 #   Whether to drop a file or remove it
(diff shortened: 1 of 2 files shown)


manifests/tmpfile.pp
```diff
@@ -26,23 +26,24 @@
...
+  Systemd::Dropin                   $filename = $name,
+  Stdlib::Absolutepath              $path     = '/etc/tmpfiles.d',
+  Optional[String]                  $content  = undef,
+  Optional[String]                  $source   = undef,
 ) {
   include systemd::tmpfiles
 
-  if $name =~ Pattern['/'] {
-    fail('$name may not contain a forward slash "(/)"')
+  if $filename =~ Pattern['/'] {
```
bastelfreak: This should work as well:
```puppet
if $name =~ Systemd::Dropin` {}
```
You don't need to introduce an additional parameter.

traylenator: `Systemd::Dropin` does not include the not no / in it.

As for additional parameter I was just copying the style of `systemd::dropin_file`

Advantage of  making it a parameter is the error is a lot more obvious I would say when you hit it.

The not containing a `/` should be added to the `Systemd::Dropin` type of course.

bastelfreak: ah thanks for pointing that out!

PR #131, opened 2020-01-06
Add loginctl_user type/provider

raphink:
Example usage:

```puppet
loginctl_user { 'foo':
  linger => enabled,
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,3 +242,13 @@ systemd::logind_settings:
     ensure: absent
   UserTasksMax: 10000
 ```
+
+### User linger
+
+A `loginctl_user` resource is available to manage user linger enablement:
+
+```puppet
+loginctl_user { 'foo':
+  linger => enabled,
+}
+```
--- a/lib/puppet/provider/loginctl_user/ruby.rb
+++ b/lib/puppet/provider/loginctl_user/ruby.rb
@@ -1,0 +1,29 @@
+Puppet::Type.type(:loginctl_user).provide(:ruby) do
+  commands :loginctl => 'loginctl'
+
+  def self.instances
+    users = loginctl('list-users', '--no-legend').split("\n").map { |l| l.split[0] }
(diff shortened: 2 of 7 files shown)


bastelfreak:
@raphink should this be documented in the README.md?

raphink:
Actually, before we release 2.8.0, I'm wondering if this is the right interface for the type. Maybe this would be better:

```puppet
user_linger { 'foo':
  ensure => enabled,
```

It makes the type ensurable and removes the implementation (`loginctl`) from the type.

What do you think @bastelfreak, @mcanevet?

bastelfreak:
From a user point of view I think both ways are. The implementation for your `user_linger` example might be easier. Another idea I had is that another parameter to the user core type would be nice. but getting stuff into puppet core takes a long time...

trevor-vaughan:
Given how many users/resources there may be on a given system, you may want to have a Hash interface that only creates one resource to prevent death by 10k resources. Probably future work, but something to keep in mind before this completely goes down the `file_line`/augeas rabbit hole.

raphink:
Yes, the core type was an option, too. That would require adding a new feature also. It is probably possible to provide this as a plugin with some monkey patching... but supporting that on the long run might not be fun... and does it make sense to implement linger for all OSes?

trevor-vaughan:
Looking at the options a bit more, I would prefer the current implementation since it reduces system resources. Technically, you could implement both and make them conflict with each other but that may be too much work.

raphink:
@trevor-vaughan why does it reduce system resources?

trevor-vaughan:
@raphink Because `loginctl_user` is one resource with many future options whereas `user_linger` is one resource per option. In the OO model, the user is the object and `linger` is the attribute. If you start modeling based on attributes, it explodes quickly.

I still think it's a horrible idea to support different daemons with different versions in the same module, but I lost that fight a while ago.

raphink:
Good point... so let's keep it the way it is then. I'll cut the release tomorrow I think.

PR #136, opened 2020-01-08
add default params

fe80:
This is a suggestion.

Actually, when we add the class without defined params the class fail. This is a feature for doing nothing per default.

This is very interesting when you have lot a server to manage.


raphink:
If you have a lot of servers to manage, you could also put neutral defaults in your `common.yaml` hiera file.

fe80:
My architecture is too complicated. We have 5k nodes with different context, and a existing systemd class to remove.

The hiera datas are cross environment,  so if I want change a default params it's apply directly in production.

And I can see it's necessarily to set  multiple params not used. Example journald_settings, but if manage_journald is false, this params is useless and consomme memory on puppetserver for nothing.

In fact this is not obligatorily to use hiera. I think this is important when you add a puppet class it's work in default comportment.

You can close this PR if you judge useless, this is a suggestion :)

reenberg:
We are in the process of considering moving to capmtocamp/systemd, and i largely agree with @fe80.  The defaults seems quite reasonable to me, and you can't beat a module that just works out of the box with an simple include.

manifests/journald.pp
```diff
@@ -10,21 +10,23 @@
...
-      $option:
-        path    => '/etc/systemd/journald.conf',
-        section => 'Journal',
-        setting => $option,
-        notify  => Service['systemd-journald'],
-    }
-    if $value =~ Hash {
-      Ini_setting[$option]{
-        * => $value,
+  if $systemd::journald_settings {
```
raphink: How about using an empty hash as a default instead, so you can keep the loop but it just does nothing?

manifests/init.pp
```diff
@@ -82,37 +82,39 @@
...
+  Enum['stopped','running']                              $timesyncd_ensure    = 'running',
+  Optional[Variant[Array,String]]                        $ntp_server          = undef,
+  Optional[Variant[Array,String]]                        $fallback_ntp_server = undef,
+  Boolean                                                $manage_accounting   = false,
+  Optional[Hash[String,String]]                          $accounting          = undef,
+  Boolean                                                $purge_dropin_dirs   = false,
+  Boolean                                                $manage_journald     = false,
+  Optional[Systemd::JournaldSettings]                    $journald_settings   = undef,
+  Boolean                                                $manage_logind       = false,
+  Optional[Systemd::LogindSettings]                      $logind_settings     = undef,
```
raphink: Or maybe put the defaults in the module's hiera data.

fe80: https://github.com/camptocamp/puppet-systemd/pull/136#issuecomment-572047779

manifests/journald.pp
fe80: That can be empty, this is check with `Systemd::JournaldSettings` format ?

manifests/init.pp
raphink: Yes, that's why I suggest to put it in the module's hiera data, so you don't have to put it in your site's hiera data.

manifests/journald.pp
raphink: All the keys in that type are optional, so yes an empty hash looks legit.

fe80: Sorry, I just not add a empty hash per default for puppet server memory, this is useless for me but we can change to an empty hash if you want

raphink: Well it would avoid patching the loops, so I think it's much simpler.

fe80: I'm not totally agree with that but it's done :)

manifests/logind.pp
```diff
@@ -10,21 +10,23 @@
...
-      $option:
-        path    => '/etc/systemd/logind.conf',
-        section => 'Login',
-        setting => $option,
-        notify  => Service['systemd-logind'],
-    }
-    if $value =~ Hash {
-      Ini_setting[$option]{
-        * => $value,
+  if $systemd::logind_settings {
```
raphink: The new default for `logind_settings` makes this moot

raphink:
@bastelfreak what do you think of these defaults?

manifests/init.pp
```diff
@@ -82,37 +82,39 @@
...
-  Optional[Variant[Array,String]]                        $ntp_server,
-  Optional[Variant[Array,String]]                        $fallback_ntp_server,
-  Boolean                                                $manage_accounting,
-  Hash[String,String]                                    $accounting,
-  Boolean                                                $purge_dropin_dirs,
-  Boolean                                                $manage_journald,
-  Systemd::JournaldSettings                              $journald_settings,
-  Boolean                                                $manage_logind,
-  Systemd::LogindSettings                                $logind_settings,
+  Optional[Hash[String,Hash[String, Any]]]               $service_limits      = undef,
```
bastelfreak: The default for hashes and arrays should always be an emtpy hash / array. That's the case in the master branch. Why whould you change that?

manifests/logind.pp
bastelfreak: from my point of view those if statements aren't a good approach. They just add a useless layer of indentation. They aren't required if the default value is an empty hash/array instead of undef.

manifests/init.pp
fe80: Hello,

I doesn't like empty value, It more memory and CPU usage, but I can change if you want.

raphink: Generally speaking, `undef` should only be the default for strings, as other types have more sensible defaults. Using `undef` as default for enumerable types results in unnecessarily complex code such as used below (`if` instead of a simple loop on an empty set).

raphink: Btw, this also means enumerable types should not be marked as `Optional`.

raphink:
@fe80 could you fix the last comments please?

fe80:
done for the rebase

bastelfreak:
I still don't understand why you did this change. Also the default values are defined in https://github.com/camptocamp/puppet-systemd/blob/master/data/common.yaml and need to be deleted their if their are defined here. The only advance that we get from this change is that puppet-strings now knows the default value. puppet-strings cannot read data-in-modules at the moment.

fe80:
https://github.com/camptocamp/puppet-systemd/pull/136#issuecomment-572047779
https://github.com/camptocamp/puppet-systemd/pull/136#issuecomment-580005307

I don't know what else to say. Hiera is not mandatory.

Close this MR if you want.

bastelfreak:
I'm not sure if you understand this setup with data-in-modules.You don't need to use hiera to use this module. This module ships its own hiera folder that's only used for default data. That's why a simple include works (as in: the catalog compiles properly). please ping raphael or me on slack at https://slack.puppet.com/ if you've any questions on how to use this module. The changes in this PR should not be a requirement for the module to be working in your environment.

fe80:
I know I can use without hiera, but for me a module can work juste with a simple include or require.

And with lots of module (we use something like 150modules), we don't have the time to set all useless parameters when we upgrade the module's release.

I can't use this code on my architecture and my leagacy code without this basic patch. Too bad for me for I need to use internal fork of this code :/

bastelfreak:
sorry but I don't see the issue. this module work by doing `include systemd`. You don't need to set any parameters. do you get any error messages if you do a simple include? Which puppetserver version are you on?

manifests/init.pp
```diff
@@ -85,33 +85,33 @@
...
+  Optional[Variant[Boolean,Enum['udp','tcp']]]           $dns_stub_listener    = undef,
+  Boolean                                                $use_stub_resolver    = false,
+  Boolean                                                $manage_networkd      = false,
+  Enum['stopped','running']                              $networkd_ensure      = 'running',
+  Boolean                                                $manage_timesyncd     = false,
+  Enum['stopped','running']                              $timesyncd_ensure     = 'running',
+  Optional[Variant[Array,String]]                        $ntp_server           = undef,
+  Optional[Variant[Array,String]]                        $fallback_ntp_server  = undef,
+  Boolean                                                $manage_accounting    = false,
+  Optional[Hash[String,String]]                          $accounting           = undef,
```
ekohl: Why not 
```suggestion
  Hash[String,String]                          $accounting           = {},
```

ekohl:
IMHO getting defaults in puppet-strings is actually a good thing and I dislike the `data/common.yaml` pattern for all parameters. The only things in `data/common.yaml` should be the ones that are overridden in other layers so you know what's hiera-defined vs really shared defaults.

trevor-vaughan:
I completely agree with @ekohl. `puppet strings` is pretty much useless if I have to go dig around the module for sensible default settings.

bastelfreak:
I agree with you two, but still those changes aren't required to make the module work. At vox pupuli we also moved back from from a common.yaml because that make puppet-strings more useful. Still I think at @fe80 has another issue. maybe a legacy puppetserver that fails to work with hiera 5.

manifests/system.pp
```diff
@@ -2,11 +2,13 @@
 #
 # This class provides a solution to enable accounting
 #
-class systemd::system {
+class systemd::system (
+  Hash[String,String] $accounting = $systemd::accounting
```
trevor-vaughan: If `systemd::accounting` is `Optional`, this MUST be `Optional`

PR #137, opened 2020-01-17
Add SLES 12/15 support

msurato:
I used OpenSUSE docker images in the acceptance test areas as that is what is available on the Docker Hub. They should be a good approximation, however.


bastelfreak:
thanks for the PR!

PR #142, opened 2020-02-24
fix Issue 113, allow setting to "no"

schlitzered:
allow setting "Cache = no"


raphink:
The specs need to be adapted to pass

raphink:
This PR will result in a change of default behavior, setting the cache to `no` by default.

@bastelfreak @ekohl what do you think?

PR #143, opened 2020-03-07
Use Storage parameter of $systemd::journald_settings to handle storage

brunoleon:
Following up on  #127 

It does not look needed to add a new parameter for journald persistend, when this is actually handled by $systemd::journald_settings in the key 'Storage'.

Also, creation of the folder should be done only if Storage=auto , otherwise this is handled by journald itself (as per man page)

### DIFF
--- a/manifests/journald.pp
+++ b/manifests/journald.pp
@@ -6,19 +6,15 @@
 class systemd::journald {
 
   assert_private()
-
-  if $systemd::journald_persist_log {
-    $journald_dir = 'directory'
-  } else {
-    $journald_dir = 'absent'
-  }
-
-  file { '/var/log/journal':
-    ensure => $journald_dir,
-    force  => true,
-    owner  => 'root',
-    group  => 'systemd-journal',
-    mode   => '2755',
+  if 'Storage' in $systemd::journald_settings {
+    if $systemd::journald_settings['Storage'] == 'persistent' {
+      file { '/var/log/journal':
+        ensure => 'directory',
+        owner  => 'root',
+        group  => 'systemd-journal',
+        mode   => '2755',
+      }
+    }
   }
 
   service{'systemd-journald':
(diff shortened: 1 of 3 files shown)


bastelfreak:
@brunoleon can you take a look at the failing spec tests?

manifests/journald.pp
```diff
@@ -6,19 +6,15 @@
...
-    $journald_dir = 'absent'
-  }
-
-  file { '/var/log/journal':
-    ensure => $journald_dir,
-    force  => true,
-    owner  => 'root',
-    group  => 'systemd-journal',
-    mode   => '2755',
+  if 'Storage' in $systemd::journald_settings {
```
bastelfreak: I don't think we need this condition? 
/me goes testing

manifests/journald.pp
```diff
@@ -7,18 +7,13 @@
...
-    $journald_dir = 'absent'
-  }
-
-  file { '/var/log/journal':
-    ensure => $journald_dir,
-    force  => true,
-    owner  => 'root',
-    group  => 'systemd-journal',
-    mode   => '2755',
+  if $systemd::journald_settings['Storage'] == 'auto' {
```
trevor-vaughan: Looking at the `journald` and `systemd` documentation, I would recommend not managing this space at all.

Systemd should "do the right thing" and this may interfere with that.

If there is a bug that needs to be worked around [like this one](https://github.com/systemd/systemd/issues/5264) then users should do that explicitly instead of having puppet fight with systemd.

bastelfreak:
Hi @brunoleon, we're going to revert https://github.com/camptocamp/puppet-systemd/pull/127 and I think this PR can then be closed. Please reopen / create a new one if you think we should merge it.

brunoleon:
I agree with your choices @bastelfreak 
Just tried to made behavior more systemd compatible.
Thanks for you module

PR #146, opened 2020-03-09
Revert "add option for persistent logging (#127)"

bastelfreak:
This reverts commit b2ddb4c4fe87ec78c4ff8a44beb16a5ee052a0a0.


bastelfreak:
As discussed with @trevor-vaughan on slack:
* This code would purge your /var/log/journald by default, which is really bad
* Systemd should create the directory, with correct ACLs, by default
* Since we don't understand the initial use case, we are going to revert this PR instead of fixing it

PR #147, opened 2020-03-31
Allow CPUQuota greater than 100%

Hexta:
Allow CPUQuota greater than 100% to allow to use more than one CPU core.

https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html
"Use values > 100% for allotting CPU time on more than one CPU. "


Hexta:
Hello.
Could anybody please take a look at the PR?

PR #148, opened 2020-04-28
Only do daemon reloads on Puppet 5

ekohl:
Puppet 6.1 contains [PUP-3483](https://tickets.puppetlabs.com/browse/PUP-3483) which means daemon reloads are not needed.

This is useful because the systemd daemon reloads can easily trigger dependency cycles. For example, managing service x and y where y depends on x.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ Create a systemd timer unit and a systemd service unit to execute from
 that timer
 
 The following will create a timer unit and a service unit file.
-The execution of `systemctl daemon-reload` will occur.
+The execution of `systemctl daemon-reload` will occur in Puppet < 6.1. Puppet >= 6.1 handles this itself.
 When `active` and `enable` are set to `true` the puppet service `runoften.timer` will be
 declared, started and enabled.
 
--- a/manifests/dropin_file.pp
+++ b/manifests/dropin_file.pp
@@ -46,6 +46,7 @@
 #   Set to `eager` to immediately reload after the dropin file is updated.
 #   Useful if the daemon needs to be reloaded before a service is refreshed.
 #   May cause multiple daemon reloads.
+#   Only relevant on Puppet < 6.1.
 #
 define systemd::dropin_file(
   Systemd::Unit                               $unit,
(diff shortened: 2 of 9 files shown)


ekohl:
I know this PR isn't a very high quality now. I think right now I'm looking for some feedback on whether this is desired at all.

raphink:
The new feature is cool, but I find that the current logic with Puppet >= 6.1 makes this code pretty complex… Could we maybe add a generic `$manage_daemon_reload` parameter and set it to default to false on Puppet >= 6.1 instead?

ekohl:
I was considering the same thing, but thought about a fact. That was IMHO ugly. A variable on the main class is IMHO a good way to reduce the duplication.

ekohl:
Updated.

bastelfreak:
I'm wondering if we get any benefit from exposing the internal variable as parameter? What about people that want to use the puppet-builtin but also use 3rd party modules that depend on the exec resource from this module?

trevor-vaughan:
@bastelfreak People really should not be triggering off of non-exposed API endpoints. If that functionality is needed, it should be wrapped in its own class as a specific trigger endpoint.

bastelfreak:
I agree. But that approach was once documented in the README.md: https://github.com/camptocamp/puppet-systemd/tree/0.4.0#unit-files

So I'm pretty sure many modules out there still rely on it. question is if we want to be nice to them allow them to workaround this or force them to update code.

ekohl:
@bastelfreak the daemon_reload class is still there and functions as it did before. It's just that right now the module will not trigger it itself anymore unless it's needed. Users are still free to include it and chain it themselves.

I also considered keeping the class and making it noop in 6.1+ but that still keeps the dependency cycles.

I think now this is a breaking change since users can no longer rely on the class being present though. They have to include it themselves.

bastelfreak:
ah, just wanted to approve it and add a label, but I don't have merge permissions anymore :(

raphink:
Oh?

raphink:
@ekohl could you rebase this please?

ekohl:
Rebased and applied a fix that I think should also work on Puppet 5. I had some troubles getting that env to even run on my dev setup so I'm waiting on Travis to confirm it really works.

ekohl:
I haven't had time to look into this. Now that Puppet 5 is EOL, any thoughts about dropping Puppet 5 support?

PR #149, opened 2020-05-29
add factory for dropin files

TheMeier:
Add a factory for dropin files so they can be simply defined in hiera

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -81,6 +81,8 @@
 # @param logind_settings
 #   Config Hash that is used to configure settings in logind.conf
 #
+# @param drpoin_files
+#   Configure dropin files via hiera with factory pattern
 class systemd (
   Hash[String,Hash[String, Any]]                         $service_limits,
   Boolean                                                $manage_resolved,
@@ -108,6 +110,7 @@ class systemd (
   Systemd::JournaldSettings                              $journald_settings,
   Boolean                                                $manage_logind,
   Systemd::LogindSettings                                $logind_settings,
+  Hash                                                   $dropin_files = {},
 ){
 
   contain systemd::systemctl::daemon_reload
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -108,6 +110,7 @@
   Systemd::JournaldSettings                              $journald_settings,
   Boolean                                                $manage_logind,
   Systemd::LogindSettings                                $logind_settings,
+  Optional[Hash]                                         $dropin_files = undef,
```
raphink, line 113: It'd be more simple to have:

```suggestion
  Hash                                                   $dropin_files = {},
```

manifests/init.pp
```diff
@@ -137,4 +140,12 @@
...
     contain systemd::logind
   }
+
+  if $dropin_files {
+    $dropin_files.each |$name, $resource| {
+      systemd::dropin_file { $name:
+        * => $resource
+      }
+    }
+  }
```
raphink, line 150: If the Hash is not optional but defaults to `{}`, then you don't need the condition anymore.
```suggestion

  $dropin_files.each |$name, $resource| {
    systemd::dropin_file { $name:
      * => $resource,
    }
  }
```

Ideally even, it'd be good to have a data type for that, like it was done for `$journald_settings` and `$logind_settings` above.

TheMeier:
I have made the changes with the empty hash default. About dedicated type, isn't that a repitition as the structure of the variables is already set in the definition of the defined resource.

manifests/init.pp
```diff
@@ -137,4 +140,10 @@
   if $manage_logind {
     contain systemd::logind
   }
+
+  $dropin_files.each |$name, $resource| {
+    systemd::dropin_file { $name:
+      * => $resource
```
raphink, line 146:
```suggestion
      * => $resource,
```

PR #151, opened 2020-05-30
Add selinux_ignore_defaults support to dropin_file and service_limits

tobias-urdin:
This commits add support for passing the service_ignore_defaults parameter to the file resources being handled in the dropin_file definition. It then adds the parameter to service_limits which use dropin_file so that consumers of this resource can pass it.

This is needed in order to support RHEL based operating systems again in puppet-rabbitmq module where the Puppet run is idempotent after moving to the systemd module [1].

[1] https://github.com/voxpupuli/puppet-rabbitmq/issues/836

### DIFF
--- a/manifests/dropin_file.pp
+++ b/manifests/dropin_file.pp
@@ -10,6 +10,9 @@
 # @attr path
 #   The main systemd configuration path
 #
+# @attr selinux_ignore_defaults
+#   If Puppet should ignore the default SELinux labels.
+#
 # @attr content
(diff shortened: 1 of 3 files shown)


tobias-urdin:
Is it possible to cut a new release? Seems like there is atleast two issues that is not released since March.

tobias-urdin:
Forget to mention @raphink @bastelfreak

tobias-urdin:
ping, any possibility to get a new release?

PR #156, opened 2020-07-22
Fix typo in parameter name in class documentation

ekohl:
90cc85cbccc00cc51b775639eb9c20293446e63e introduced this but with a typo in the
name so puppet-strings couldn't find it.


raphink:
Good catch!

PR #163, opened 2020-10-20
Add parameter for ENCs to make loginctl_users easily

jcpunk:
This should make it much easier to build `loginctl_users` for folks running with an ENC.

### DIFF
--- a/manifests/logind.pp
+++ b/manifests/logind.pp
@@ -27,4 +27,10 @@ class systemd::logind {
       }
     }
   }
+
+  $systemd::loginctl_users.each |$loginctl_name, $loginctl_settings| {
+    loginctl_user { $loginctl_name:
+      * => $loginctl_settings,
+    }
+  }
 }
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -296,7 +296,7 @@ describe 'systemd' do
               value: '10000',
             )
           }
-          it { is_expected.to have__loginctl_user('foo').with(linger: 'enabled') }
+          it { is_expected.to contain_loginctl_user('foo').with(linger: 'enabled') }
         end
         context 'when passing dropin_files' do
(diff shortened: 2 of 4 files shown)


manifests/logind.pp
```diff
@@ -27,4 +27,8 @@
       }
     }
   }
+
+  if !(empty($systemd::loginctl_users) {
+    create_resources(loginctl_user, $systemd::loginctl_users)
```
ekohl, line 32: It's a bit redundant to check for an empty hash. These days it's better to avoid `create_resources`. There may be typos in the following code, but roughly like this:
```suggestion
  $systemd::loginctl_users.each |$loginctl_name, $loginctl_user| {
    loginctl_user { $loginctl_name:
      * => $loginctl_users,
```

jcpunk: I'll merge this in with some minor tweaks.

jcpunk:
I have no idea what syntax the spec tests want...

spec/classes/init_spec.rb
```diff
@@ -293,6 +296,7 @@
               value: '10000',
             )
           }
+          it { is_expected.to have__loginctl_user('foo').with(linger: 'enabled') }
```
ekohl, line 299: I think this fixes it:
```suggestion
          it { is_expected.to contain__loginctl_user('foo').with(linger: 'enabled') }
```

spec/classes/init_spec.rb
```diff
@@ -293,6 +296,7 @@
               value: '10000',
             )
           }
+          it { is_expected.to contain__loginctl_user('foo').with(linger: 'enabled') }
```
ekohl, line 299: Looks like I missed another thing in my suggestion:
```suggestion
          it { is_expected.to contain_loginctl_user('foo').with(linger: 'enabled') }
```

PR #164, opened 2020-10-20
Permit using arrays to make extending lists easier

jcpunk:
For the loginctl user parameter lists, it is easier to set them as an array (that way I can append or de-dupe easily).

Psudo hiera:

```yaml
systemd::logind_settings:
  KillExcludeUsers: 
  - root
```

```yaml
systemd::logind_settings:
  KillExcludeUsers: 
  - other_user
```

### DIFF
--- a/manifests/logind.pp
+++ b/manifests/logind.pp
@@ -21,6 +21,10 @@ class systemd::logind {
       Ini_setting[$option] {
         * => $value,
       }
+    } elsif $value =~ Array {
+      Ini_setting[$option] {
+        value   => join($value, ' '),
+      }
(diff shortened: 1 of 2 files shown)


ekohl (approved):
Technically only `KillOnlyUsers` and `KillExcludeUsers` can be lists, but for those this is a good improvement. I didn't find other lists that could have a different separator when reading `man 5 logind.conf`.

raphink:
That should be fine, considering [the type](https://github.com/camptocamp/puppet-systemd/blob/master/types/logindsettings.pp) validates these parameters properly.

PR #165, opened 2020-10-23
Add management of udev objects

jcpunk:
With `udev` part of systemd these days, it is handy to have a place to set custom udev rules (for folks that need that like me).

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,7 @@ class{'systemd':
   manage_networkd  => true,
   manage_timesyncd => true,
   manage_journald  => true,
+  manage_udevd     => true,
   manage_logind    => true,
 }
 ```
@@ -318,6 +319,41 @@ systemd::journald_settings:
   MaxRetentionSec: 5day
   MaxLevelStore:
     ensure: absent
+```
+
+### udevd configuration
+
+It allows you to manage the udevd configuration.  You can set the udev.conf values via the `udev_log`, `udev_children_max`, `udev_exec_delay`, `udev_event_timeout`, `udev_resolve_names`, and `udev_timeout_signal` parameters.
+
+Additionally you can set custom udev rules with the `udev_rules` parameter.
+
+```puppet
+class { 'systemd':
+  manage_udevd => true,
+  udev_rules   => { 
+      'example_raw.rules' => {
+      'rules'             => [
+        'ACTION=="add", KERNEL=="sda", RUN+="/bin/raw /dev/raw/raw1 %N"',
+        'ACTION=="add", KERNEL=="sdb", RUN+="/bin/raw /dev/raw/raw2 %N"',
+      ],
+    },
+  },
+}
+```
+
+### udev::rules configuration
+
+Custom udev rules can be defined for specific events.
+
+```yaml
+systemd::udev::rule:
+  ensure: present
+  path: /etc/udev/rules.d
+  selinux_ignore_defaults: false
+  notify: "Service[systemd-udevd']"
+  rules:
+    - 'ACTION=="add", KERNEL=="sda", RUN+="/bin/raw /dev/raw/raw1 %N"'
+    - 'ACTION=="add", KERNEL=="sdb", RUN+="/bin/raw /dev/raw/raw2 %N"',
 ```
 
 ### logind configuration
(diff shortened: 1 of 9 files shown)


templates/udev_rule.erb
```diff
@@ -0,0 +1,4 @@
+# This file managed by Puppet - DO NOT EDIT
```
raphink, line 1: Why not use EPP?

jcpunk: Old habbits die hard.... I'll switch over to EPP and have a new version of this up soon.

manifests/udev/rule.pp
```diff
@@ -0,0 +1,53 @@
...
+  Enum['present', 'absent', 'file'] $ensure                  = 'present',
+  Stdlib::Absolutepath              $path                    = '/etc/udev/rules.d',
+  Boolean                           $restart_service         = true,
+  Optional[Boolean]                 $selinux_ignore_defaults = false,
+) {
+  include systemd
+
+  if $name !~ Pattern['^.+\.rules$'] {
+    fail('$name must match Pattern["^.+\.rules)$"]')
+  }
```
raphink, line 36: I wonder if `assert_type` would be more elegant.

templates/udev_rule.epp
```diff
@@ -0,0 +1,4 @@
+# This file managed by Puppet - DO NOT EDIT
+<% @rules.each do | rule | -%>
```
raphink, line 2: This is not EPP syntax 😁

templates/udev_conf.epp
```diff
@@ -0,0 +1,23 @@
+# This file managed by Puppet - DO NOT EDIT
+#
+# The initial syslog(3) priority: "err", "info", "debug" or its
+# numerical equivalent. For runtime debugging, the daemons internal
+# state can be changed with: "udevadm control --log-priority=<value>".
+<% if scope['systemd::udev_log'] -%>
```
raphink, line 6: I would rather pass the parameters explicitly.

manifests/udev/rule.pp
jcpunk: Does this look better?

templates/udev_conf.epp
jcpunk: something like this work?

README.md
```diff
@@ -320,6 +321,40 @@ systemd::journald_settings:
...
+
+```puppet
+class{'systemd':
+  manage_udevd => true,
+  udev_rules => { 'example_raw.rules' => {
+                    'rules' => [
+                      'ACTION=="add", KERNEL=="sda", RUN+="/bin/raw /dev/raw/raw1 %N"',
+                      'ACTION=="add", KERNEL=="sdb", RUN+="/bin/raw /dev/raw/raw2 %N"',
+                    ] } }
+}
```
raphink, line 338:
```suggestion
class { 'systemd':
  manage_udevd => true,
  udev_rules   => { 
      'example_raw.rules' => {
      'rules'             => [
        'ACTION=="add", KERNEL=="sda", RUN+="/bin/raw /dev/raw/raw1 %N"',
        'ACTION=="add", KERNEL=="sdb", RUN+="/bin/raw /dev/raw/raw2 %N"',
      ],
    },
  },
}
```

manifests/udev/rule.pp
```diff
@@ -0,0 +1,53 @@
...
+    group                   => 'root',
+    mode                    => '0444',
+    path                    => join([$path, $name], '/'),
+    selinux_ignore_defaults => $selinux_ignore_defaults,
+    content                 => epp("${module_name}/udev_rule.epp", {'rules' => $rules}),
+  }
+
+  if $restart_service {
+    File[$name] {
+      notify => Service['systemd-udevd'],
```
raphink, line 50: I think I prefer a `$notify` parameter set to undef by default and passed directly to the resource.

raphink:
Almost there 😁

manifests/udev/rule.pp
jcpunk: I've taken a stab at it.

jcpunk:
My local PDK seems to be b0rked, so I'll have to rely on travis to validate my tests for now.

raphink:
Now we wait…

I'll tackle the migration to GH Actions soon(-ish).

jcpunk:
I think travis is happy again.

Issue #166, opened 2020-12-16
definition of $cache in init.pp  doesn't include 'no-negative'


usagijer:
in manifests/resolved.pp
the $cache param is defined as:
  Optional[Variant[Boolean,Enum['no-negative']]] $cache              = $systemd::cache,

in init.pp , the no-negative option is not included.

PR #169, opened 2021-01-11
Allow additional option for $cache parameter

bryangwilliam:
Update the cache parameter on init class to match the allowed values for
resolved.

This addresses the issue brought up here: https://github.com/camptocamp/puppet-systemd/issues/166 

https://github.com/camptocamp/puppet-systemd/pull/140 initally added the ability to use the no-negative option for resolved, however it did not include the corresponding adjustment to allow controlling the value from the init class.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -40,7 +40,7 @@
 #   Takes a boolean argument or "opportunistic"
 #
 # @param cache
-#   Takes a boolean argument.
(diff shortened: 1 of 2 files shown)


raphink:
LGTM. Could you add a unit test for this please?

bryangwilliam:
Added testing around the cache variants.

PR #171, opened 2021-01-13
Drop Puppet 4 and 5 support + daemon-reload code

ekohl:
Since Puppet 6.1.0 it's no longer needed to run daemon-reload manually when restarting a service. That means it's possible to drop this code. Since Puppet 4 & 5 are now EOL, this should be ok.

Additionally it contains some documentation fixes (https://github.com/camptocamp/puppet-systemd/pull/172).

Alternative to https://github.com/camptocamp/puppet-systemd/pull/148.


manifests/service_limits.pp
```diff
@@ -75,7 +75,6 @@
       path        => $::path,
       refreshonly => true,
       subscribe   => File["${path}/${name}.d/90-limits.conf"],
-      require     => Class['systemd::systemctl::daemon_reload'],
```
ekohl, line 78: I'm a bit unsure about this (which is also why this is a draft). This isn't a service type so it doesn't perform the reload if needed. Not sure how to deal with this.

ekohl:
Rebased since the included PR was merged so the diff is smaller. No actual change.

manifests/service_limits.pp
ekohl: @raphink any thoughts on this?

raphink: Looks likes it should stay then...

ekohl: But if the class to require is gone, should there be a local daemon reload?

ekohl (commented):
I have since found out that there are edge cases where you do need daemon-reload. Not sure if we want account for that. I can see about writing up something in the README.

tuxmea:
Alternative to #179

ekohl:
Just a trivial rebase now to see if the tests are green, no actual code change.

ekohl:
I added a section to the README but then I realized I hadn't addressed the service limits case. Can anyone take a look if this is a good direction or would you prefer to solve it with code?

trevor-vaughan:
@ekohl Since it only occurs during a removal, I'd much rather have it handled in code if possible. Without that, we'll all end up with a bunch of confused end users at some point.

When the upstream code is corrected, a new version can be released that removes the workaround.

ekohl:
It happens in an odd edge case. The following code doesn't suffer from it:
```puppet
systemd::unit_file { 'myservice.service':
  ensure => absent,
  active => true,
}
```

That's because the module considers that an invalid state.

This does:
```puppet
systemd::unit_file { 'myservice.service':
  ensure => absent,
}

service { 'myservice':
  ensure  => running,
  require => Systemd::Unit_file['myservice.service'],
}
```

What happened in puppetlabs-postgresql was exactly that. It switched from a module provided unit file to a package provided unit file.

I'm pushing the update as a separate commit for easier review. Let me know what you think.

trevor-vaughan (commented):
:eye: :+1:

raphink (approved):
LGTM

bastelfreak (approved):

Issue #159, opened 2020-09-08
Allow service reloading


k0ka:
There is no easy way to use 
`systemctl reload <service>`
for a specified service. Puppet will always restart it. 

I suggest adding some parameter like `$use_reload` to `systemd::unit_file` that will set the `restart` argument of `service`.

raphink:
That seems like a reasonable idea. Could you provide a PR for this please?

k0ka:
Hi,

here it is https://github.com/camptocamp/puppet-systemd/pull/175

PR #175, opened 2021-01-25
Allow service reloading #159

k0ka:


manifests/unit_file.pp
```diff
@@ -46,18 +46,22 @@
...
+  Optional[String]                         $content    = undef,
+  Optional[String]                         $source     = undef,
+  Optional[Stdlib::Absolutepath]           $target     = undef,
+  String                                   $owner      = 'root',
+  String                                   $group      = 'root',
+  String                                   $mode       = '0444',
+  Boolean                                  $show_diff  = true,
+  Optional[Variant[Boolean, Enum['mask']]] $enable     = undef,
+  Optional[Boolean]                        $active     = undef,
+  Optional[Boolean]                        $use_reload = undef,
```
raphink, line 64: Or maybe pass a restart command? This would be more flexible.

k0ka: The real command is written into the unit file. This variable switches between default `run ExecStop,ExecStart commands` and `run ExecReload commands`.

Here is an example unit file: 
```
[Service]
ExecStart=docker run my-container
ExecStop=docker stop my-container
ExecReload=docker exec my-container kill -HUP 1
```

So in case of `$use_reload = false` it will stop/run the container. And in the other case if will send SIGHUP. 

I think reload commands should be written inside of unit file, so they work without puppet. Here is the list of  systemctl commands, that could be used to restart the service:

1. restart
2. reload
3. try-restart
4. reload-or-restart
5. try-reload-or-restart

We may accept one of these 5 commands, so the user won't make a typo in their names.

Btw, I suggest setting the default value to `reload-or-restart`. So if the user add `ExecReload` into the unit file, it will be automatically used.

raphink: Right, so there's more than one value acceptable, which is why I think a Boolean is too narrow, and will commit us to a wrong level of abstraction.

An `Enum` as you're suggesting is a bit more open (and allows for a more generic pattern eventually), though it looks to me like this should just be an open field, to allow users to specify the command to be execute.

k0ka: Do you mean just adding the `$restart` argument and passing it to the `service`?

In my opinion, it breaks the encapsulation as the user has to pass `systemctl reload <service>`, and if this command has to be like `systemctl-new reload <service>` in some environment (probably in the future), they will have to handle it by themselves. And I think this is where the package should help them. 

Anyways, if that's what you think right - I've updated this PR.

raphink: @bastelfreak what's your opinion?

manifests/unit_file.pp
```diff
@@ -46,6 +46,9 @@
 # @param active
 #   If set, will manage the state of the unit.
 #
+# @param restart
+#   Specify a restart command manually. If left unspecified, the service will be stopped and then started.
```
ekohl, line 50: I read https://puppet.com/docs/puppet/7.3/type.html#service-attribute-restart and I think it's outdated. if `hasrestart` is true, the restart command is used. Note that since Puppet 2.7 `hasrestart` defaults to true.

Perhaps you can avoid it this way:

```suggestion
#   Specify a restart command manually. If left unspecified, a standard Puppet service restart happens.
```

For reference:

manifests/unit_file.pp
ekohl: I can see an argument for `Variant[Enum['restart', 'reload', 'try-restart', 'reload-or-restart', 'try-reload-or-restart'], String[1]]`. However, I'm pretty sure `restart` is the default so that option doesn't really make sense.

https://github.com/puppetlabs/puppet/blob/902024b83762aabfa99f10a6018f0d26d92272cf/lib/puppet/provider/service/systemd.rb#L170-L172

manifests/unit_file.pp
k0ka: the standard Puppet service restart uses command `systemctl restart <service>` that stops and than starts the service.

ekohl: That is the implementation, but at the module level we can't say much more than "we'll do what Puppet does".

manifests/unit_file.pp
raphink: So are you fine with the current patch @ekohl ?

ekohl (approved):
:+1: looks good to me.

PR #176, opened 2021-01-25
Use EPP rather than ERB for template files

jcpunk:
Switch the existing template to use EPP rather than ERB


raphink:
Good idea! It just needs `limits.erb` to be ported to the EPP syntax, and the `$limits` array to be passed as parameter.

jcpunk:
... Hmmm my attempts to port over line 9 failed.... I think I'll have to close this out and let someone with a better grip on EPP tackle this one.

raphink:
Ah yes, that's a fun line indeed 😁

PR #181, opened 2021-03-02
Fix clash with modules that manage the unit folder too

hdeheer:
Other puppet modules in a puppet setup might manage the `{unit}.d` folder as well. For example puppetlabs/docker does (See https://github.com/puppetlabs/puppetlabs-docker/blob/main/manifests/service.pp).
If one still wants to add limits overrides via this systemd module, puppet will complain about duplicate resource declaration.

### DIFF
--- a/manifests/dropin_file.pp
+++ b/manifests/dropin_file.pp
@@ -65,7 +65,7 @@ define systemd::dropin_file (
     }
   }
 
-  if $ensure != 'absent' {
+  if $ensure != 'absent' and !defined(File["${path}/${unit}.d"]) {
     ensure_resource('file', "${path}/${unit}.d", {
         ensure                  => directory,
         owner                   => 'root',


ekohl (commented):
Isn't that exactly what `ensure_resource` is supposed to do?

hdeheer:
@ekohl You would think so but unfortunately in our case it doesn't.
The detail is in the explanation of the function `ensure_resource`.

> If the resource already exists but does not match the specified parameters,
> this function will attempt to recreate the resource leading to a duplicate
> resource definition error.

You can test yourself:
**ensure.pp**
```
file { '/tmp/foo.d':
  ensure => directory
}

ensure_resource('file', '/tmp/foo.d', {
  ensure                  => directory,
  owner                   => 'root',
  group                   => 'root',
  recurse                 => false,
  purge                   => false,
  selinux_ignore_defaults => false,
})
```
Results in:
```
$ puppet apply ensure.pp
Error: Evaluation Error: Error while evaluating a Function Call, Duplicate declaration: File[/tmp/foo.d]
is already declared at (file: /tmp/ensure.pp, line: 1); cannot redeclare (file: /tmp/ensure.pp, line: 5)
(file: /tmp/ensure.pp, line: 5, column: 1) on node
```

raphink:
Iirc, `ensure_resource` only works if the parameters are exactly the same in both declarations.

hdeheer:
That's why we had to add the `if !defined` check. For standalone usage inside the systemd module, `ensure_resource` works like it should. But you will run into puppet failures when the `{unit}.d` folder is being managed by another puppet module.
@raphink how can I get this change merged?

ekohl:
@hdeheer do you know which parameters don't match?

hdeheer:
@ekohl Yes.
For example https://github.com/puppetlabs/puppetlabs-docker/blob/main/manifests/service.pp only uses: 
- `ensure`

```
file { '/etc/systemd/system/docker.service.d':
  ensure => 'directory',
}
```
But the systemd module `manifests/dropin_file.pp`  uses:
- `ensure`
- `owner`
- `group`
- `recurse`
- `purge`
- `selinux_ignore_defaults`

Easy fix in our case is to use `if !defined`.

ekohl:
I still think it's a bad idea. This module depends on certain behavior, such as purge. If that's not specified, you get unpredicatable behavior. Namely, depending on the parse order it could purge unmanaged files or not. That can be a nasty surprise.

raphink:
I'm also not a fan of this at all. I'd rather contribute to the Docker module to depend on the systemd module.

jhoblitt (changes requested):
I think this is a bad idea. This module should have the expectation that it has ownership of the `.d` directory.

traylenator:
This should be closed.

Will close after the solstice.

TheMeier:
Winter solstice 2023 @traylenator? :D

PR #191, opened 2021-04-22
Add notify_service support to dropin_file

ekohl:
It is very common that a drop in file should notify a service that's managed elsewhere. This adds a parameter to do so which then uses a collector.

Review notes:
* Should the parent directory also notify the service? My main reasoning be that if it purges unknown files, it should also notify the service. I don't even know if that works
* I made it default to false to be compatible. This makes it a bit hard to use though.
* Should this be exposed on any defined type that uses dropin_file?

### DIFF
--- a/manifests/dropin_file.pp
+++ b/manifests/dropin_file.pp
@@ -40,6 +40,9 @@
 # @param show_diff
 #   Whether to show the diff when updating dropin file
 #
+# @param notify_service
+#   Notify a service for the unit, if it exists
+#
 define systemd::dropin_file (
   Systemd::Unit                               $unit,
   Systemd::Dropin                             $filename                = $name,
@@ -53,6 +56,7 @@ define systemd::dropin_file (
   String                                      $group                   = 'root',
   String                                      $mode                    = '0444',
   Boolean                                     $show_diff               = true,
+  Boolean                                     $notify_service          = false,
 ) {
   include systemd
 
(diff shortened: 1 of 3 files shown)


manifests/dropin_file.pp
```diff
@@ -40,6 +40,9 @@
 # @attr show_diff
 #   Whether to show the diff when updating dropin file
 #
+# @attr notify_service
```
ekohl, line 43: I noticed this was wrong and opened https://github.com/camptocamp/puppet-systemd/pull/192

bastelfreak (approved):

kenyon (approved):

manifests/dropin_file.pp
raphink: #192 is merged. Do you want to rebase?

manifests/dropin_file.pp
```diff
@@ -65,8 +69,10 @@
     }
   }
 
+  $full_filename = "${path}/${unit}.d/${filename}"
+
   if $ensure != 'absent' {
-    ensure_resource('file', "${path}/${unit}.d", {
+    ensure_resource('file', dirname($full_filename), {
```
raphink, line 75: Is that because `$filename` might contain a slash?

ekohl: It was mostly so that I could really use the variable `$full_filename` everywhere.

raphink: Ack.

manifests/dropin_file.pp
ekohl: Rebased & fixed. Also regenerated REFERENCE.md

raphink (approved):

PR #193, opened 2021-04-27
Remove restart_service on service_limits define

### LINKED ISSUE #190: daemon reload problem with 3.0.0
After updating to 3.0.0 limits configuration seems to be broken.
In the manifest I have:
```
systemd::service_limits { 'php7.4-fpm.service':
  limits => {
    'LimitNOFILE' => 1048576,
  }
}
```

Puppet applied this code and tried to perform service restart:
`Exec[restart php7.4-fpm.service because limits]`

But in fact, new limits were not applied, because of lack of daemon reload. By code I see there is call of:
`systemctl restart php7.4-fpm.service`

By running it by hand after puppet run, I got:
```
~# systemctl restart php7.4-fpm.service
Warning: The unit file, source configuration file or drop-ins of php7.4-fpm.service changed on disk. Run 'systemctl daemon-reload' to reload units.
```
So, I need to reload daemon manually, and then restart the service.

OS Debian 10.9
Puppet server 7.1.0
Puppet agent 7.5.0

ekohl:
Since 97dd16fa32886b5b0f77a6f38a4953d4c1511c0e this parameter is a bad idea. It doesn't do a daemon reload and it may restart at the wrong time. In fact, I'd argue it's always been a bad idea.

The recommended alternative to this is to manage the service explicitly and let Puppet handle it. There is an automatic notify that takes care of it.

Fixes #190

### DIFF
--- a/manifests/service_limits.pp
+++ b/manifests/service_limits.pp
@@ -26,16 +26,12 @@
 #
 #   * Mutually exclusive with ``$limits``
 #
-# @param restart_service
-#   Restart the managed service after setting the limits
-#
 define systemd::service_limits (
   Enum['present', 'absent', 'file'] $ensure                  = 'present',
   Stdlib::Absolutepath              $path                    = '/etc/systemd/system',
   Boolean                           $selinux_ignore_defaults = false,
   Optional[Systemd::ServiceLimits]  $limits                  = undef,
   Optional[String]                  $source                  = undef,
-  Boolean                           $restart_service         = true
 ) {
   include systemd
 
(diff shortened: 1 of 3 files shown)


ekohl:
This is a backwards incompatible change, but I can't apply the label.

bastelfreak (approved):

kenyon (approved):

ghoneycutt (approved):

ghoneycutt:
@ekohl looks like this just needs a rebase

ekohl:
Rebased

ekohl:
No urgency. I'm fine with keeping this till the next major version.

manifests/service_limits.pp
```diff
@@ -67,15 +63,6 @@
...
-
-  if $restart_service {
-    exec { "restart ${name} because limits":
-      command     => "systemctl restart ${name}",
-      path        => $facts['path'],
-      refreshonly => true,
-    }
-
-    Systemd::Dropin_file["${name}-90-limits.conf"] ~> Exec["restart ${name} because limits"]
+    notify_service          => true,
```
ekohl, line 66: Should this be configurable?

jhoblitt (approved):
LGTM. We should have done this long ago...

PR #197, opened 2021-05-26
Add a socket_service definition

ekohl:
For systemd socket activated services it's needed to have both the .socket and .service unit defintions in place when starting the service.  Prior to 97dd16fa32886b5b0f77a6f38a4953d4c1511c0e the systemctl daemon-reload took care of it, but now this must be done explicitly.

This new defined type makes it easy and reduces the burden on modules that define the unit/service pair.

I haven't tested this yet, but submitting it early so I can run CI on it and link it elsewhere.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -35,6 +35,7 @@
 * [`systemd::modules_load`](#systemd--modules_load): Creates a modules-load.d drop file
 * [`systemd::network`](#systemd--network): Creates network config for systemd-networkd
 * [`systemd::service_limits`](#systemd--service_limits): Adds a set of custom limits to the service
+* [`systemd::socket_service`](#systemd--socket_service): Create a systemd socket activated service
 * [`systemd::timer`](#systemd--timer): Create a timer and optionally a service unit to execute with the timer unit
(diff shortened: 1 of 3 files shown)


bastelfreak (approved):

ekohl:
In https://github.com/theforeman/puppet-pulpcore/pull/199 I'm testing if the approach works. A draft until then.

bastelfreak (approved):

bastelfreak (approved):

ekohl:
This now fails on https://tickets.puppetlabs.com/browse/MODULES-11018. I'll add a workaround.

ekohl:
I've rebased this, but too tired to fix the dependency cycles now.

jcpunk:
Since `manage_unit` and `manage_dropin` both support `socket` elements, is something like this still needed since you could - in theory - make the defines depend on each other when declared?

ekohl:
I completely forgot about this PR and I don't have a direct need for it anymore. I'd be happy for someone to take it over.

jcpunk:
I'll close it out for now. If someone else discovers the need for it, then we could probably drive it forward faster at that time.

PR #199, opened 2021-06-16
Do a daemon reload for static units

simondeziel:
Services that are deployed by systemd::timer need to trigger a `systemctl daemon-reload` when the .service file is changed.

```puppet
systemd::timer { 'foo.timer':
    timer_content   => $foo_timer,
    service_content => $foo_service,
    active          => true,
    enable          => true,
}
```
results in a `foo.service` that is neither active nor enabled. It is a "static" unit that will be triggered by the `foo.timer` alone. This fixes voxpupuli#190 and more specifically: https://github.com/voxpupuli/puppet-systemd/issues/190#issuecomment-851701455

### DIFF
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -104,8 +104,9 @@ define systemd::unit_file (
     } else {
       File["${path}/${name}"] ~> Service[$name]
     }
-  } elsif $ensure == 'absent' {
+  } else {
     # Work around https://tickets.puppetlabs.com/browse/PUP-9473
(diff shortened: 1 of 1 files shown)


kenyon:
Can you also update the tests in https://github.com/voxpupuli/puppet-systemd/blob/master/spec/defines/unit_file_spec.rb to cover this? That way we don't have a regression.

ekohl (commented):
I think this change is good, but as @kenyon said: tests would be nice.

PR #200, opened 2021-06-25
Add support for strict mode for DNS over TLS

ghost:
Updated parameter in systemd::resolve::dnsovertls

Support for a strict mode was added in systemd release 243.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,7 +269,7 @@ $manage_networkd is required if you want to reload it for new
 
 When configuring `systemd::resolved` you could set `dns_stub_resolver` to false (default) to use a *standard* `/etc/resolved.conf`, or you could set it to `true` to use the local resolver provided by `systemd-resolved`.
 
-Systemd has introduced `DNS Over TLS` in the release 239. Currently two states are supported `no` and `opportunistic`. When enabled with `opportunistic` `systemd-resolved` will start a TCP-session to a DNS server with `DNS Over TLS` support. Note that there will be no host checking for `DNS Over TLS` due to missing implementation in `systemd-resolved`.
+Systemd has introduced `DNS Over TLS` in the release 239. Currently three states are supported `yes` (since systemd 243), `opportunistic` (true) and `no` (false, default). When enabled with `yes` or `opportunistic` `systemd-resolved` will start a TCP-session to a DNS server with `DNS Over TLS` support. When enabled with `yes` (strict mode), queries will fail if the configured DNS servers do not support `DNS Over TLS`. Note that there will be no host checking for `DNS Over TLS` due to missing implementation in `systemd-resolved`.
 
 It is possible to configure the default ntp servers in `/etc/systemd/timesyncd.conf`:
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -37,7 +37,8 @@
 #   Takes a boolean argument or "allow-downgrade".
 #
 # @param dnsovertls
-#   Takes a boolean argument or "opportunistic"
+#   Takes a boolean argument or one of "yes", "opportunistic" or "no". "true" corresponds to
+#   "opportunistic" and "false" (default) to "no".
 #
 # @param cache
 #   Takes a boolean argument or "no-negative".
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -265,7 +265,7 @@ $manage_networkd is required if you want to reload it for new
 
 When configuring `systemd::resolved` you could set `dns_stub_resolver` to false (default) to use a *standard* `/etc/resolved.conf`, or you could set it to `true` to use the local resolver provided by `systemd-resolved`.
 
-Systemd has introduced `DNS Over TLS` in the release 239. Currently two states are supported `no` and `opportunistic`. When enabled with `opportunistic` `systemd-resolved` will start a TCP-session to a DNS server with `DNS Over TLS` support. Note that there will be no host checking for `DNS Over TLS` due to missing implementation in `systemd-resolved`.
+Systemd has introduced `DNS Over TLS` in the release 239. Currently three states are supported `yes`, `no`, and `opportunistic`. When enabled with `yes` or `opportunistic` `systemd-resolved` will start a TCP-session to a DNS server with `DNS Over TLS` support. When running with the `yes` setting (strict mode), queries will fail if the configured DNS servers do not support `DNS Over TLS`. Note that there will be no host checking for `DNS Over TLS` due to missing implementation in `systemd-resolved`.
```
ekohl, line 268: Should this also make a note that `yes` is only available in 243 or newer?

wobblesprout: Good point. I've since updated it.

kenyon (approved):

manifests/init.pp
```diff
@@ -113,42 +113,42 @@
...
+  Hash[String,Hash[String, Any]]                               $service_limits,
+  Boolean                                                      $manage_resolved,
+  Enum['stopped','running']                                    $resolved_ensure,
+  Optional[Variant[Array[String],String]]                      $dns,
+  Optional[Variant[Array[String],String]]                      $fallback_dns,
+  Optional[Variant[Array[String],String]]                      $domains,
+  Optional[Variant[Boolean,Enum['resolve']]]                   $llmnr,
+  Optional[Variant[Boolean,Enum['resolve']]]                   $multicast_dns,
+  Optional[Variant[Boolean,Enum['allow-downgrade']]]           $dnssec,
+  Optional[Variant[Boolean,Enum['yes','opportunistic', 'no']]] $dnsovertls,
```
bastelfreak, line 125: hey, could you please put `Variant[Boolean,Enum['yes','opportunistic', 'no']]` into it's own custom type? I think that would make the whole thing more readable and we don't need to add so much whitespace everywhere

wobblesprout: Sounds like a good idea. Just the one custom type for resolved::dnsovertls or should i create more while I'm at it?

ghoneycutt: The custom types are not worth it unless it can be used multiple times. If it is only used once or twice, adding custom types just adds complexity to the module making it harder to understand and maintain.

bastelfreak (approved):

manifests/init.pp
wobblesprout: I've reverted all the whitespace changes. The alignment does look a bit weird since they aren't aligned to anything. Saying that the alignment for init.pp and resolved.pp is inconsistent:

init.pp:
` Hash                                                   $dropin_files = {},
  Hash                                                   $udev_rules = {},`

resolved.pp
` Enum['stopped','running'] $ensure                                  = $systemd::resolved_ensure,
  Optional[Variant[Array[String],String]] $dns                       = $systemd::dns,`

The first the variables are aligned, the second the equals are.

ekohl (commented):
My only concern now is that `true` maps to `opportunistic` but if you just read the code `yes` may feel more intuitive. We should at least explain this.
https://github.com/voxpupuli/puppet-systemd/blob/f184b624f3bbe8544df9105ea058b6fa8c13e079/manifests/resolved.pp#L182-L186

wobblesprout:
@ekohl - changed to:

```
  $_dnsovertls = $dnsovertls ? {
    'yes'   => true, 
    true    => 'opportunistic',
    false   => false,
    default => $dnsovertls,
  }
```
Ideally, true would be mapped to true and false to false.

bastelfreak:
@wobblesprout we fixed the CI pipeline. Can you please rebase against our latest master branch?

kenyon:
Rebased on latest master.

bastelfreak (approved):

alexjfisher:
Adding comment to force github into reindexing this PR (so it gets included in Changelogs). Sorry for the noise! :(

PR #204, opened 2021-07-27
Dont add puppetlabs_spec_helper to Gemfile

bastelfreak:
voxpupuli-test already pulls it in. We can check if we really need to
add it explicitly to the Gemfile.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,6 @@ group :release do
   gem 'puppet-strings', '>= 2.2',                 :require => false
 end
 
-gem 'puppetlabs_spec_helper', '~> 2.0', :require => false
 gem 'rake', :require => false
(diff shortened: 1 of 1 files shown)


ekohl (changes requested):
Are you planning to submit this to our modulesync config? In https://github.com/voxpupuli/modulesync_config/commit/9d32848f855c56807f83d2653622c77e630b529c I added this, since voxpupuli-test is not present in acceptance. However, that may no longer be needed since  https://github.com/voxpupuli/voxpupuli-acceptance/commit/1d6d055b9d3cd627f007ab79f9e3477d61c779a7. We do not yet load this in our Rakefile.

Also beware that the `rake beaker` task is slightly different than what puppetlabs_spec_helper did.

Please test this on a module with acceptance tests.

bastelfreak:
we also tested it in https://github.com/voxpupuli/puppet-redis/pull/410#pullrequestreview-715799707

PR #205, opened 2021-07-27
systemd::network: Validate if content/source are set for file resource

bastelfreak:

### DIFF
--- a/manifests/network.pp
+++ b/manifests/network.pp
@@ -21,10 +21,10 @@ define systemd::network (
 
   if $ensure == 'file' {
     if $content =~ Undef and $source =~ Undef {
-      fail('Either content or source must be set but not both')
+      fail('Either content or source must be set')
     }
     if $content =~ NotUndef and $source =~ NotUndef {
-      fail('Either content or source must be set')
+      fail('Either content or source must be set but not both')
     }
   }
   file { "${path}/${name}":
--- a/manifests/network.pp
+++ b/manifests/network.pp
@@ -19,6 +19,14 @@ define systemd::network (
     $notify = undef
   }
 
+  if $ensure == 'file' {
+    if $content =~ Undef and $source =~ Undef {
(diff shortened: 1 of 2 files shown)


bastelfreak:
copied it from https://github.com/voxpupuli/puppet-bird/blob/master/manifests/snippet.pp#L25-L30

manifests/network.pp
```diff
@@ -19,6 +19,14 @@
     $notify = undef
   }
 
+  if $ensure == 'file' {
+    if $content =~ Undef and $source =~ Undef {
+      fail('you need to set content or source')
```
ekohl, line 24:
```suggestion
      fail('Either content or source must be set but not both')
```

manifests/network.pp
```diff
@@ -19,6 +19,14 @@
...
+    }
+    if $content =~ NotUndef and $source =~ NotUndef {
+      fail('you can only set $content or $source')
```
ekohl, line 27:
```suggestion
      fail('Either content or source must be set')
```

spec/defines/network_spec.rb
```diff
@@ -61,6 +61,25 @@
 
           it { is_expected.to create_file("/etc/systemd/network/#{title}").that_notifies('Service[systemd-networkd]') }
         end
+        context 'without content and without source' do
+          let(:title) { 'wg0.netdev' }
```
ekohl, line 65: There is already an inherited title, right?
```suggestion
```

spec/defines/network_spec.rb
```diff
@@ -61,6 +61,25 @@
...
+          let :params do
+            {}
+          end
+
+          it { is_expected.to compile.and_raise_error(%r{you need to set}) }
+        end
+        context 'with content and source' do
+          let(:title) { 'wg0.netdev' }
```
ekohl, line 73: Same here
```suggestion
```

PR #212, opened 2021-08-12
Add additional hash parameters for the remaining defined types

silug:
This adds parameters to manage the following resources via Hiera:
* `systemd::network`
* `systemd::timer`
* `systemd::tmpfile`
* `systemd::unit_file`

It also enables deep merging by default for all of the resource hash parameters.

N/A

### DIFF
--- a/data/common.yaml
+++ b/data/common.yaml
@@ -1,5 +1,9 @@
 ---
 systemd::service_limits: {}
+systemd::networks: {}
+systemd::timers: {}
+systemd::tmpfiles: {}
+systemd::unit_files: {}
 systemd::manage_resolved: false
 systemd::resolved_ensure: 'running'
 systemd::dns: ~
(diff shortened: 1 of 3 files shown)


data/common.yaml
```diff
@@ -34,3 +38,32 @@ systemd::manage_logind: false
 systemd::logind_settings: {}
 systemd::manage_all_network_files: false
 systemd::network_path: '/etc/systemd/network'
+systemd::loginctl_users: {}
+systemd::dropin_files: {}
+systemd::udev_rules: {}
+
+lookup_options:
```
ghoneycutt, line 45: This is not a standard pattern and would caution against specifying the lookup strategy. Usually this is not needed and if it is, that is managed outside a module.

manifests/init.pp
```diff
@@ -154,11 +170,15 @@
   Systemd::LogindSettings                                $logind_settings,
   Boolean                                                $manage_all_network_files,
   Stdlib::Absolutepath                                   $network_path,
-  Hash                                                   $loginctl_users = {},
-  Hash                                                   $dropin_files = {},
-  Hash                                                   $udev_rules = {},
+  Hash[String,Hash[String, Any]]                         $loginctl_users,
```
ghoneycutt, line 173: Could you please add spec tests for these? Here's an example at https://github.com/ghoneycutt/puppet-module-bind/blob/master/spec/classes/init_spec.rb#L511-L534 

You specify params with a valid hash and then check to see that the defined types are in the catalog.

ghoneycutt (changes requested):

ghoneycutt:
Thanks for this contribution!

bastelfreak:
Hi,
thanks for the PR. I'm going to close this since we merged the changes in https://github.com/voxpupuli/puppet-systemd/pull/223.

PR #220, opened 2021-09-06
Add systemd::escape function

traylenator:

Add a systemd escape function.

```puppet
systemd::escape('/foo/bar')
```
returns `-foo-bar`

while the path escape can be executed as

```puppet
systemd::escape('/foo/bar',true)
```

to return 'foo-bar'

### DIFF
--- a/functions/escape.pp
+++ b/functions/escape.pp
@@ -1,0 +1,43 @@
+# @summary Escape strings as systemd-escape does.
+# @param input Input string
+# @param path Use path (-p) ornon-path  style escaping.
+# @return String
+# @example Escaping a string
+#   $result = systemd::escape('foo::bar')
+# @example Escaping a path
+#   $result = systemd::escape('/mnt/foobar',true)
+function systemd::escape(String[1] $input, Boolean $path = false) >> String {
+  # Escape method is defined
+  # https://www.freedesktop.org/software/systemd/man/systemd.unit.html
+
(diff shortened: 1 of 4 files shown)


functions/escape.pp
```diff
@@ -0,0 +1,43 @@
+# @summary Escape strings as systemd-escape does.
+# @param input Input string
+# @param path Use path (-p) ornon-path  style escaping.
+# @return String
+# @example Escaping a string
+#   $result = systemd::escape('foo::bar')
+# @example Escaping a path
+#   $result = systemd::escape('/mnt/foobar',true)
+#
+function systemd::escape(String $input, Boolean $path = false) >> String {
```
bastelfreak, line 10: I think we can prohibit empty strings?
```suggestion
function systemd::escape(String[1] $input, Boolean $path = false) >> String {
```

traylenator: Did first but then 
```
systemd-escape ''
```
is not an error... 

Should test it though.

bastelfreak: sure, but that doesn't make any sense?

traylenator: Though the fact that

```
 systemd-escape -p  ''
-
```
makes we want to avoid the question.

bastelfreak (approved):

bastelfreak (approved):

ghoneycutt (approved):

PR #221, opened 2021-09-07
Migrate static data from hiera to init.pp

bastelfreak:
This allows us to get rid of one file and represents the data within the
REFERENCE.md


manifests/init.pp
```diff
@@ -123,50 +123,50 @@
...
+  Hash[String,Hash[String, Any]]                      $service_limits = {},
+  Boolean                                             $manage_resolved = false,
+  Enum['stopped','running']                           $resolved_ensure = 'running',
+  Optional[Variant[Array[String],String]]             $dns = undef,
+  Optional[Variant[Array[String],String]]             $fallback_dns = undef,
+  Optional[Variant[Array[String],String]]             $domains = undef,
+  Optional[Variant[Boolean,Enum['resolve']]]          $llmnr = undef,
+  Optional[Variant[Boolean,Enum['resolve']]]          $multicast_dns = undef,
+  Optional[Variant[Boolean,Enum['allow-downgrade']]]  $dnssec = undef,
+  Variant[Boolean,Enum['yes', 'opportunistic', 'no']] $dnsovertls = false,
```
bastelfreak, line 141: I changed the datatype of $dnsovertls and $cache. It was Optional but the common.yaml set both to false.

REFERENCE.md
```diff
@@ -322,30 +382,40 @@ Data type: `Boolean`
...
+Default value: ``false``
+
 ##### <a name="accounting"></a>`accounting`
 
 Data type: `Hash[String,String]`
 
-Hash of the different accounting options. This highly depends on the used systemd version. The module provides sane defaults.
+Hash of the different accounting options. This highly depends on the used systemd version. The module provides sane defaults per operating system.
+
+Default value: `{}`
```
ekohl, line 409: This may be a bit misleading now but I'm not sure how to do it better.

manifests/init.pp
```diff
@@ -123,50 +123,50 @@
 #   when enabled, the different accounting options (network traffic, IO, CPU util...) are enabled for units
 #
 # @param accounting
-#   Hash of the different accounting options. This highly depends on the used systemd version. The module provides sane defaults.
+#   Hash of the different accounting options. This highly depends on the used systemd version. The module provides sane defaults per operating system.
```
ekohl, line 126: Perhaps add something like
```suggestion
#   Hash of the different accounting options. This highly depends on the used systemd version. The module provides sane defaults per operating system using Hiera.
```

ekohl (approved):
Overall :+1:

REFERENCE.md
bastelfreak: I think we should drop the default empty hash from the init.pp

bastelfreak: updated it

PR #223, opened 2021-09-07
Add additional hash parameters for defined types

bastelfreak:
This is a rebase of https://github.com/voxpupuli/puppet-systemd/pull/212

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -6,6 +6,18 @@
 #   May be passed a resource hash suitable for passing directly into the
 #   ``create_resources()`` function as called on ``systemd::service_limits``
 #
+# @param networks
+#   Hash of `systemd::network` resources
+#
+# @param timers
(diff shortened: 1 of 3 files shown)


TuningYourCode (approved):

manifests/init.pp
```diff
@@ -168,7 +184,31 @@
   Hash                                                $dropin_files = {},
   Hash                                                $udev_rules = {},
 ) {
-  create_resources('systemd::service_limits', $service_limits)
+  $service_limits.each |String[1] $service_limit, Hash[String[1], Any] $service_limit_data| {
```
ekohl, line 187: This feels overly strict. Is it really needed to set data types in these lines? IMHO you don't get any additional type safety while it is longer to read and also means you need to update it in 2 places.

bastelfreak: cleaned it up

ekohl (approved):

PR #224, opened 2021-09-07
Release 3.5.0

bastelfreak:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
+## [v3.5.0](https://github.com/voxpupuli/puppet-systemd/tree/v3.5.0) (2021-09-13)
+
+[Full Changelog](https://github.com/voxpupuli/puppet-systemd/compare/v3.4.0...v3.5.0)
+
+**Implemented enhancements:**
+
+- Add Gentoo support [\#227](https://github.com/voxpupuli/puppet-systemd/pull/227) ([bastelfreak](https://github.com/bastelfreak))
+- Add CentOS/RHEL 9 support [\#226](https://github.com/voxpupuli/puppet-systemd/pull/226) ([mbaldessari](https://github.com/mbaldessari))
+- Use os.family for RedHat based Hiera data [\#225](https://github.com/voxpupuli/puppet-systemd/pull/225) ([treydock](https://github.com/treydock))
+- Add additional hash parameters for defined types [\#223](https://github.com/voxpupuli/puppet-systemd/pull/223) ([bastelfreak](https://github.com/bastelfreak))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
...
+
+**Implemented enhancements:**
+
+- Add additional hash parameters for defined types [\#223](https://github.com/voxpupuli/puppet-systemd/pull/223) ([bastelfreak](https://github.com/bastelfreak))
+- Add Debian 11 support [\#222](https://github.com/voxpupuli/puppet-systemd/pull/222) ([bastelfreak](https://github.com/bastelfreak))
+- Add systemd::escape function [\#220](https://github.com/voxpupuli/puppet-systemd/pull/220) ([traylenator](https://github.com/traylenator))
+
+**Merged pull requests:**
+
+- Migrate static data from hiera to init.pp [\#221](https://github.com/voxpupuli/puppet-systemd/pull/221) ([bastelfreak](https://github.com/bastelfreak))
```
ekohl, line 19: I'm starting to wake up, but we should consider how we classify this. I think technically this can be breaking for some users. To verify this: previously you could do `lookup('systemd::$VAR')` even if the module wasn't included from other modules. Now you can't. This is probably not used a lot, but it can break some use cases. That's why I think https://github.com/puppetlabs/puppet-strings/pull/273 would be a better long term solution, but I don't have time to finish it.

bastelfreak: yes, the puppet-strings PR would be the better approach, I also have that on my todo list since years but no time/not enough knowledge. Do you think we should here do a major release instead?

ekohl: I'd like to avoid that, but we should keep it in mind.

CHANGELOG.md
```diff
@@ -4,6 +4,23 @@ All notable changes to this project will be documented in this file.
...
+
+- Add Gentoo support [\#227](https://github.com/voxpupuli/puppet-systemd/pull/227) ([bastelfreak](https://github.com/bastelfreak))
+- Add CentOS/RHEL 9 support [\#226](https://github.com/voxpupuli/puppet-systemd/pull/226) ([mbaldessari](https://github.com/mbaldessari))
+- Add additional hash parameters for defined types [\#223](https://github.com/voxpupuli/puppet-systemd/pull/223) ([bastelfreak](https://github.com/bastelfreak))
+- Add Debian 11 support [\#222](https://github.com/voxpupuli/puppet-systemd/pull/222) ([bastelfreak](https://github.com/bastelfreak))
+- Add systemd::escape function [\#220](https://github.com/voxpupuli/puppet-systemd/pull/220) ([traylenator](https://github.com/traylenator))
+
+**Merged pull requests:**
+
+- Use os.family for RedHat based Hiera data [\#225](https://github.com/voxpupuli/puppet-systemd/pull/225) ([treydock](https://github.com/treydock))
```
ekohl, line 21: Enhancement? This makes it work on all RHEL derivatives instead of just RHEL/CentOS.

bastelfreak: sounds good, I added the label

ekohl (approved):

PR #226, opened 2021-09-09
Add CentOS/RHEL 9 support

mbaldessari:
Add CentOS/RHEL 9 support. Without this change puppet-systemd will
likely fail with the following error:

    Evaluation Error: Error while evaluating a Function Call, Class[Systemd]:
    expects a value for parameter 'accounting'
    (file: /etc/puppet/modules/systemd/manifests/service_limits.pp, line: 40, column: 3)

Tested on Centos 9 Stream.

Signed-off-by: Michele Baldessari <michele@acksyn.org>

### DIFF
--- a/data/RedHat-9.yaml
+++ b/data/RedHat-9.yaml
@@ -1,0 +1,8 @@
+---
+systemd::accounting:
+  DefaultCPUAccounting: 'yes'
+  DefaultBlockIOAccounting: 'yes'
+  DefaultMemoryAccounting: 'yes'
+  DefaultTasksAccounting: 'yes'
+  DefaultIOAccounting: 'yes'
+  DefaultIPAccounting: 'yes'
--- a/metadata.json
+++ b/metadata.json
@@ -46,7 +46,8 @@
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
         "7",
-        "8"
+        "8",
+        "9"
       ]
     },
     {


bogdando (approved):
well done

data/CentOS-9.yaml
```diff
@@ -0,0 +1,8 @@
+---
```
ekohl, line 1: Since 2ef9d94c2e0ec388a5c1013d2c0148719a6edcca I think this file is redundant.

ekohl:
How do you verify this? Do you install the EL8 puppet-agent.rpm on EL9?

data/CentOS-9.yaml
mbaldessari: A thanks, indeed. I'll leave only one.

mbaldessari:
> How do you verify this? Do you install the EL8 puppet-agent.rpm on EL9?

I installed puppet 7.8 el9 onto a centos stream9 box. I used https://trunk.rdoproject.org/centos9-master-preview/current/delorean.repo + https://trunk.rdoproject.org/centos9-master-preview/delorean-deps.repo. The puppet rpm itself I was using is https://trunk.rdoproject.org/centos9-master-preview/deps/latest/noarch/puppet-7.8.0-1.el9.noarch.rpm

mbaldessari:
I removed CentOS-9 and retested the whole thing on my env.

mbaldessari:
I forgot to mention that when using puppet-7.8 you need to manually patch in https://github.com/puppetlabs/puppet/pull/8694/commits/b8711b541c04fde2326a6471da965fe1feafe9f0. 7.10 already includes that.

ekohl (approved):
Thanks!

PR #236, opened 2021-11-05
Link the unit file to /dev/null when "enable => mask"

### LINKED ISSUE #188: Parameter value 'mask' for 'enable' does not work
It would be nice to be able to mask unit files using `enable => mask` but although this value is an [accepted parameter](https://github.com/camptocamp/puppet-systemd/blob/97dd16fa32886b5b0f77a6f38a4953d4c1511c0e/manifests/unit_file.pp#L62), it doesn't set `target => '/dev/null'` as expected.

simondeziel:
Fixes #188

### DIFF
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -67,7 +67,13 @@ define systemd::unit_file (
 
   assert_type(Systemd::Unit, $name)
 
-  if $target {
+  if $enable == 'mask' {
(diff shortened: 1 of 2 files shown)


kenyon (approved):

root-expert (approved):

ekohl (approved):
Thanks!

PR #237, opened 2021-11-07
Dropin daemon reload

### LINKED ISSUE #234: systemd::dropin_file doesn't cause a systemd daemon-reload

- Puppet: 6.23.0
- Ruby: 2.5.9
- Distribution: Ubuntu 18.04
- Module version: 3.5.1

```puppet
systemd::dropin_file { 'runtimedirectory.conf':
  unit    => 'nginx.service',
  content => @(EOT),
    # Managed by Puppet
    [Service]
    RuntimeDirectory=nginx
    | EOT
}
```
## What are you seeing
systemd not reloaded.
`systemctl daemon-reload` should be done.

simondeziel:
Fixes #234

### DIFF
--- a/manifests/dropin_file.pp
+++ b/manifests/dropin_file.pp
@@ -69,11 +69,20 @@ define systemd::dropin_file (
     show_diff               => $show_diff,
   }
 
+  exec { "${name}-dropin-systemctl-daemon-reload":
+    command     => 'systemctl daemon-reload',
+    refreshonly => true,
+    path        => $facts['path'],
+    subscribe   => File[$full_filename],
+  }
+
   if $notify_service {
     File[$full_filename] ~> Service <| title == $unit or name == $unit |>
+    Exec["${name}-dropin-systemctl-daemon-reload"] -> Service <| title == $unit or name == $unit |>
     if $unit =~ /\.service$/ {
       $short_service_name = regsubst($unit, /\.service$/, '')
       File[$full_filename] ~> Service <| title == $short_service_name or name == $short_service_name |>
+      Exec["${name}-dropin-systemctl-daemon-reload"] -> Service <| title == $short_service_name or name == $short_service_name |>
     }
   }
 }
--- a/spec/defines/dropin_file_spec.rb
+++ b/spec/defines/dropin_file_spec.rb
@@ -127,6 +127,15 @@ describe 'systemd::dropin_file' do
           }
         end
 
+        context 'when using default values' do
+          it {
+            expect(subject).to create_exec("#{title}-dropin-systemctl-daemon-reload").with(
+              command: 'systemctl daemon-reload',
+              refreshonly: true
+            )
+          }
+        end
+
         context 'with sensitve content' do
           let(:title) { 'sensitive.conf' }
           let(:params) do


kenyon (approved):

ekohl:
I'm not sure this is correct. Puppet should already reload services that are managed. In https://github.com/voxpupuli/puppet-systemd/issues/234#issuecomment-962998689 I asked for some more info.

simondeziel:
> I'm not sure this is correct. Puppet should already reload services that are managed.

Which it does but the bug is for services that only dropin files are managed. I use those all the time to add a delta on top of what the distro/packager provided for example.

manifests/dropin_file.pp
```diff
@@ -69,11 +69,20 @@
     show_diff               => $show_diff,
   }
 
+  exec { "${name}-dropin-systemctl-daemon-reload":
+    command     => 'systemctl daemon-reload',
+    refreshonly => true,
```
ekohl, line 74: Rather than `refreshonly` I wonder if `onlyif => 'systemctl show -p NeedDaemonReload | grep yes'` would work.

traylenator: Seems like a good idea :thinking:

ekohl:
In https://github.com/voxpupuli/puppet-systemd/issues/234#issuecomment-964077912 I suggested to use a define to avoid multiple reloads for the same service, in case multiple drop ins are used.

manifests/dropin_file.pp
traylenator: Needs the unit name in there of course.
```
onlyif => 'systemctl show -p NeedDaemonReload $unit | grep yes'
```
The downside is this being run potentially many times per unit per puppet run always but the simplicity wins for me. 

`systemd::unit_file` would need the same code for an unmanned .service and .timer so would be nice to really only have one exec per unit. As mentioned else where - defined type and `ensure_resource`

simondeziel: The `onlyif` incurs overhead for every run. Probably a personal preference, but `refreshonly` coupled with `subscribe` feels simple and self descriptive to me.

simondeziel: The `onlyif => "systemctl show -p NeedDaemonReload $unit | grep yes"` way makes sense if for some reason a user has both a `systemd::unit_file` and a `systemd::dropin_file` for the same service.

I consider this to be an edge case because if you deploy the unit, why bother with a dropin? However, if you do need this unusual setup, it will work but `systemctl daemon-reload` will be called twice. The overhead of that second extraneous call that happens only on changes is probably not too far off from the `NeedDaemonReload` test it's avoiding for every single runs.

simondeziel:
>     * GDM with `hidepid=2` requires an updated to `systemd-logind` via a dropin file.

Sorry if that's not related to the core issue but I've been wanting to go back to `hidepid=2` for sometimes but didn't know how to make it work. What's the `systemd-logind` tweak that's needed, please? :)

trevor-vaughan:
@simondeziel See https://github.com/simp/pupmod-simp-gdm.

Also, are you up for getting this changed now or should I refactor into another PR? At a minimum, we need to get the `exec` into its own defined type.

trevor-vaughan:
@simondeziel Can you try this? https://github.com/voxpupuli/puppet-systemd/tree/daemon-reload

It works for me, but I'm working on a section to fix all of the systems that got broken by the current 3.X branch (`daemon-reload` was never called)

simondeziel:
@trevor-vaughan the `daemon-reload` branch works for me as well, thanks!

trevor-vaughan:
Closing in favor of #277 per the conversation above

PR #243, opened 2022-01-23
Add function systemd::systemd_escape

### LINKED ISSUE #242: systemd::escape function is does not escape a lot of other characters
The function [`systemd::escape`](https://github.com/voxpupuli/puppet-systemd/blob/master/functions/escape.pp) introduced in https://github.com/voxpupuli/puppet-systemd/pull/220 supports only a minimalistic subset of escape sequences.

For mount units, the `-` needs to be escaped and replaced by `\x2d` (hex representation).

In general, numbers, letters and some special chars are [allowed](https://github.com/systemd/systemd/blob/16d41892c36bc31921c112ee6529f2d15a8fd449/src/basic/unit-name.c#L19-L32) and others needs to be [replaced by a hex presentation](https://github.com/systemd/systemd/blob/16d41892c36bc31921c112ee6529f2d15a8fd449/src/basic/unit-name.c#L293-L302).

Before someone ask 'Can you create a PR?'. 

First I want to know if it worth to reimplement the systemd-escape logic into the puppet DSL or should the puppet function replaced by a function it's simply calling `systemd-escape`? In combination with [`Defferred functions`](https://puppet.com/docs/puppet/6/deferring_functions.html), it should NOT add the dependency having systemd installed on the puppet master.

In case the current function should extend, I need to know how I can convert strings/chars to octal or hex decimal and do comparisons like `if "x".ord > 127`?

jkroepke:
This PR is a proposal for #242 and its intended to replace the current function `systemd::escape`.

This PR add a function `systemd::systemd_escape` based on the system call systemd-escape.

See #220 and #95

`rubocop:disable Style/OptionalBooleanParameter` is required here, otherwise puppet [wont detect](https://github.com/voxpupuli/puppet-systemd/runs/4913907935?check_suite_focus=true) the optional parameter.

Fixes #242

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,7 +425,7 @@ or as a hash via the `systemd::loginctl_users` parameter.
 ### Systemd Escape Function
 Partially escape strings as `systemd-escape` command does.
 
-This functions only escapes a subset of chars. Non ASCII character will not escape.
+This functions only escapes a subset of chars. Non-ASCII character will not escape.
 
 ```puppet
 $result = systemd::escape('foo::bar/')
--- a/README.md
+++ b/README.md
@@ -423,7 +423,9 @@ loginctl_user { 'foo':
 or as a hash via the `systemd::loginctl_users` parameter.
 
 ### Systemd Escape Function
-Escapes strings as `systemd-escape` command does.
+Partially escape strings as `systemd-escape` command does.
+
+This functions only escapes a subset of chars. Non ASCII character will not escape.
 
 ```puppet
 $result = systemd::escape('foo::bar/')
@@ -436,6 +438,23 @@ or path escape as if with `-p` option.
 $result = systemd::escape('/mnt/foobar/', true)
 ```
 `$result` would be `mnt-foobar`.
+
+### Systemd Escape Function (uses systemd-escape)
+Escape strings by call the `systemd-escape` command in the background.
+
+It's highly recommend running the function as [deferred function](https://puppet.com/docs/puppet/6/deferring_functions.html) since it executes the command on the agent.
+
+```puppet
+$result = Deferred('systemd::systemd_escape', ["foo::bar"])
+```
+`$result` would be `foo::bar-`
+
+or path escape as if with `-p` option.
+
+```puppet
+$result = Deferred('systemd::systemd_escape', ["/mnt/foo-bar/", true])
+```
+`$result` would be `mnt-foo\x2dbar`.
 
 ## Transfer Notice
 
--- a/lib/puppet/functions/systemd/systemd_escape.rb
+++ b/lib/puppet/functions/systemd/systemd_escape.rb
@@ -10,7 +10,8 @@ Puppet::Functions.create_function(:'systemd::systemd_escape') do
     return_type 'String'
   end
 
-  def escape(input, path: false)
+  # rubocop:disable Style/OptionalBooleanParameter
+  def escape(input, path = false)
     args = []
 
     args.push('--path') if path
@@ -18,6 +19,7 @@ Puppet::Functions.create_function(:'systemd::systemd_escape') do
     args.push(input)
     exec_systemd(args)
   end
+  # rubocop:enable Style/OptionalBooleanParameter
 
   def exec_systemd(*args)
     exec_args = { failonfail: true, combine: true }
--- a/lib/puppet/functions/systemd/systemd_escape.rb
+++ b/lib/puppet/functions/systemd/systemd_escape.rb
@@ -22,7 +22,7 @@ Puppet::Functions.create_function(:'systemd::systemd_escape') do
   # rubocop:enable Style/OptionalBooleanParameter
 
   def exec_systemd(*args)
-    exec_args = { failonfail: true, combine: true }
-    Puppet::Util::Execution.execute(['systemd-escape', args], **exec_args)
+    exec_args = { failonfail: true, combine: false }
+    Puppet::Util::Execution.execute(['systemd-escape', args], **exec_args).to_s.trim
   end
 end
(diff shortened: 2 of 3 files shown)


lib/puppet/functions/systemd/systemd_escape.rb
```diff
@@ -0,0 +1,29 @@
...
+    args.push('--path') if path
+
+    args.push(input)
+    exec_systemd(args)
+  end
+  # rubocop:enable Style/OptionalBooleanParameter
+
+  def exec_systemd(*args)
+    exec_args = { failonfail: true, combine: false }
+    escaped = Puppet::Util::Execution.execute(['systemd-escape', args], **exec_args)
```
smortex, line 26: Relying on an external command should be avoided.  Can't we just improve the current `systemd::escape` function if it's current behavior is wrong?

spec/functions/systemd_escape_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+describe 'systemd::systemd_escape' do
```
smortex, line 4: These tests unit will not work on systems where systemd is not installed (e.g. FreeBSD). While end-users running such os will not use this modules, maintainers at voxpupuli will want to run the test suite locally to ensure they do not break the module with refactoring :wink: 

A more portable way to test this is to mock the behavior of the execution:

```
    allow(Puppet::Util::Execution).to receive(:execute).with('systemd-escape', 'foo').and_return("foo\n")
```

then chek it happened

```
    expect(Puppet::Util::Execution).to have_received(:execute).with('systemd-escape', 'foo')
```

smortex (commented):
Can't we just improve the current `systemd::escape` function if it's current behavior is wrong?

lib/puppet/functions/systemd/systemd_escape.rb
jkroepke: The current function is a different implementation. From my point of view, this is a breaking change, but I‘m fine with it.

jkroepke:
> Can't we just improve the current `systemd::escape` function if it's current behavior is wrong?

#242 is here to discuss this. This PR is a proposal for one solution.

lib/puppet/functions/systemd/systemd_escape.rb
bastelfreak: My personal opinion: Calling out to external tools should always be the last option. But Puppet already does that to systemd. Now it would be nice if there would be a systemd API binding for Ruby, but that doesn't really exist and that would require us to install a gem, which also sucks. We can keep the old function as well, so people can choose between using `systemd::systemd_escape` vs `systemd::escape`. I think that's fine.

jkroepke: > so people can choose between

This was my first intension, too. And thats why I decide to use a different name. In combination with Defferred functions, systemd-escape is executed on the agent, not on the master.

I understand, calling an external tool is slow, but the puppet lange does not provide functions to translate characters to hex which is required to improve the existing function.

bastelfreak: The hex translation could be handled by another function/it could be rewritten as a Ruby function, but IMO that only makes it more complicated and there is no guarantee that our reimplementation works like the actual systemd-escape. That's why I think having both implementations is fine.

lib/puppet/functions/systemd/systemd_escape.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# @summary Escape strings by call the `systemd-escape` command in the background.
+Puppet::Functions.create_function(:'systemd::systemd_escape') do
+  # @param input Input string
+  # @param path Use path (-p) ornon-path  style escaping.
+  dispatch :escape do
+    param 'String', :input
+    optional_param 'Optional[Boolean]', :path
```
ekohl, line 9: I think optional booleans are odd. Can't we simply have a boolean with a default false? Or put in a different way: what's the difference here between `nil` / `undef` and `false`?

README.md
```diff
@@ -374,7 +374,9 @@ loginctl_user { 'foo':
 or as a hash via the `systemd::loginctl_users` parameter.
 
 ### Systemd Escape Function
-Escapes strings as `systemd-escape` command does.
+Partially escape strings as `systemd-escape` command does.
+
+This functions only escapes a subset of chars. Non ASCII character will not escape.
```
ekohl, line 379:
```suggestion
This functions only escapes a subset of chars. Non-ASCII character will not escape.
```

lib/puppet/functions/systemd/systemd_escape.rb
jkroepke: I copy this from here https://github.com/puppetlabs/puppetlabs-stdlib/blob/276e7b7cd97580c9c08b264b3727e5902df1f487/lib/puppet/functions/to_json_pretty.rb#L48 where `Optional[Boolean]` is defined, but the default value is `false` (like line 14 here).

jkroepke:
Whats next here? Planned to merge the PR? Otherwise, I will close the stale PR.

bastelfreak (approved):

smortex (commented):
I feel like we would have the same coverage with less tests: since we mock the execution we just want to test if the parameters passed to `#execute` are the expected ones.

jkroepke:
Rebased. Test reduced. removed robocup comments.

bastelfreak (approved):

jkroepke:
@smortex Are you fine with the latest changes?

smortex (approved):

PR #244, opened 2022-01-25
Manage entries in modules-load.d directory

traylenator:

A new type `systemd::modules_load` can add entries
to the directory `/etc/modules-load.d`.

```puppet
systemd::modules_load{'ipmi.conf':
  content => "ipmi\n",
}
```
Upon modification `systemctl  start systemd-modules-load.service` will
be called to activate the new file.

### DIFF
--- a/manifests/modules_load.pp
+++ b/manifests/modules_load.pp
@@ -1,0 +1,59 @@
+# Creates a modules-load.d drop file
+#
+# @api public
+#
+# @see modules-load.d(5)
+#
+# @param filename
+#   The name of the modules-load.d file to create
+#
+# @param ensure
+#   Whether to drop a file or remove it
+#
+# @param path
+#   The path to the main systemd modules-load.d directory
+#
+# @param content
+#   The literal content to write to the file
+#
+#   * Mutually exclusive with ``$source``
+#
+# @param source
+#   A ``File`` resource compatible ``source``
+#
+#   * Mutually exclusive with ``$content``
+#
+# @example load a module
+#   systemd::modules_load{'impi.conf':
+#      content => "ipmi\n",
+#   }
+#
+# @example override /lib/modules-load.d/myservice.conf in /etc/modules-load.d/myservice.conf
+#   systemd::modules_load{'myservice.conf':
+#      content => "# Cancel system version of the file\n",
+#   }
+#
+define systemd::modules_load (
+  Enum['present', 'absent', 'file'] $ensure   = 'file',
+  Systemd::Dropin                   $filename = $name,
+  Stdlib::Absolutepath              $path     = '/etc/modules-load.d',
+  Optional[String[1]]               $content  = undef,
+  Optional[String[1]]               $source   = undef,
+) {
+  include systemd::modules_loads
+
+  $_tmp_file_ensure = $ensure ? {
+    'present' => 'file',
+    default   => $ensure,
+  }
+
+  file { "${path}/${filename}":
+    ensure  => $_tmp_file_ensure,
+    content => $content,
+    source  => $source,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0444',
+    notify  => Class['systemd::modules_loads'],
+  }
+}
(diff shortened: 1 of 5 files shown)


manifests/modules_loads.pp
```diff
@@ -0,0 +1,13 @@
+# Activate the  modules contained in modules-loads.d
```
ekohl, line 1:
```suggestion
# Activate the modules contained in modules-loads.d
```

spec/defines/modules_load_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        it { is_expected.to contain_class('systemd::modules_loads') }
+        it { is_expected.to contain_exec('systemd-modules-load').with_command('systemctl start systemd-modules-load.service') }
+
+        context 'with a bad modules-load name' do
+          let(:title) { 'test.badtype' }
+
+          it {
+            expect do
+              expect(subject).to compile.with_all_deps
+            end.to raise_error(%r{expects a match for Systemd::Dropin})
```
ekohl, line 37: Isn't this the right way to test for it?
```suggestion
            is_expected.to compile.and_raise_error(%r{expects a match for Systemd::Dropin})
```

manifests/modules_load.pp
```diff
@@ -0,0 +1,59 @@
...
+#   systemd::modules_load{'myservice.conf':
+#      content => "# Cancel system version of the file\n",
+#   }
+#
+define systemd::modules_load (
+  Enum['present', 'absent', 'file'] $ensure   = 'file',
+  Systemd::Dropin                   $filename = $name,
+  Stdlib::Absolutepath              $path     = '/etc/modules-load.d',
+  Optional[String]                  $content  = undef,
+  Optional[String]                  $source   = undef,
```
bastelfreak, line 41:
```suggestion
  Optional[String[1]]               $source   = undef,
```
do we need to allow empty strings?

manifests/modules_load.pp
```diff
@@ -0,0 +1,59 @@
...
+# @example override a file in /lib/modules-load.d
+#   systemd::modules_load{'myservice.conf':
+#      content => "# Cancel system version of the file\n",
+#   }
+#
+define systemd::modules_load (
+  Enum['present', 'absent', 'file'] $ensure   = 'file',
+  Systemd::Dropin                   $filename = $name,
+  Stdlib::Absolutepath              $path     = '/etc/modules-load.d',
+  Optional[String]                  $content  = undef,
```
bastelfreak, line 40:
```suggestion
  Optional[String[1]]               $content  = undef,
```

manifests/modules_loads.pp
bastelfreak: should that be:
```suggestion
# @summary Activate the modules contained in modules-loads.d
```

?

spec/defines/modules_load_spec.rb
traylenator: That's better - I basically copied the tmpfile  type and modified.

manifests/modules_load.pp
traylenator: empty strings are useful since that is I actual use case where I want to clobber a broken package config - https://bugzilla.redhat.com/show_bug.cgi?id=2044789

Having said that forcing a comment into these files is a good idea..

spec/defines/modules_load_spec.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'systemd::modules_load' do
+  context 'supported operating systems' do
+    on_supported_os.each do |os, facts|
+      context "on #{os}" do
+        let(:facts) { facts }
```
bastelfreak, line 9:
```suggestion
    on_supported_os.each do |os, os_facts|
      context "on #{os}" do
        let(:facts) { os_facts }
```

otherwise people might get confused by facts vs facts

spec/defines/modules_load_spec.rb
ekohl: It's a bad pattern I've seen in various places. It probably gets copied and copied.

bastelfreak (approved):

PR #245, opened 2022-02-08
systemd::escape: Also escape - (dash)

weaselp:
We also need to escape dashes in unit names, for instance for .swap services on device nodes with dashes in them.

systemd-escape escapes dashes:
```
# systemd-escape -p /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_drive-ua-swap
dev-disk-by\x2did-scsi\x2d0QEMU_QEMU_HARDDISK_drive\x2dua\x2dswap
```
but this function does not (yet).

In the long term, we may want to more closely match the behaviour of
systemd's unit-name.c, but this is all I need for now.

Fix systemd::escape a bit

### DIFF
--- a/functions/escape.pp
+++ b/functions/escape.pp
@@ -34,6 +34,7 @@ function systemd::escape(String[1] $input, Boolean $path = false) >> String {
       }
       '/': { $_escaped = '-' }
       ',': { $_escaped = '\x2c' }
+      '-': { $_escaped = '\x2d' }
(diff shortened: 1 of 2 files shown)


smortex:
Thanks!  Can you please update the test suite in `spec/functions/escape_spec.rb` to make sure we do not break this it in the future?

smortex (approved):
LGTM!

PR #246, opened 2022-02-11
Install systemd-resolved on CentOS 8 and 9

traylenator:

On CentOS 8 and 9 systemd-resolved must be installed to use
systemd-resolved.

Install the package if `manage_resolved` is `true` which is not the default.

Since the `systemd_internal_services` fact will only be populated
once the package is installed a second puppet run will be required
to actually configure systemd-resolved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,6 +292,9 @@ class{'systemd':
 }
 ```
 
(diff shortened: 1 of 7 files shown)


kenyon (approved):

olifre:
I cross-checked with a CentOS 8 Stream system, and the same seems to apply there (no separate `systemd-resolved` package, but contained in `systemd`).

PR #250, opened 2022-02-16
document systemd::unit_file example with puppet-strings

bastelfreak:

### DIFF
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -58,6 +58,13 @@
 # @param selinux_ignore_defaults
 #   maps to the same param on the file resource for the unit. false in the module because it's false in the file resource type
 #
+# @example manage unit file + service
+#   systemd::unit_file { 'foo.service':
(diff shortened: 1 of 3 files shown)


manifests/unit_file.pp
```diff
@@ -58,6 +58,13 @@
 # @param selinux_ignore_defaults
 #   maps to the same param on the file resource for the unit. false in the module because it's false in the file resource type
 #
+# @example manage unit file + service
+#   systemd::unit_file { 'foo.service':
+#     source => "puppet:///modules/${module_name}/foo.service",
```
ekohl, line 63: Personally I'd always use `content => file("${module_name}/foo.service")` but I suppose both work.

ekohl (approved):

kenyon (approved):

manifests/unit_file.pp
bastelfreak: ah yeah, I usually do that as well. fixed it.

PR #251, opened 2022-02-16
Manage systemd-coredump config and setup

traylenator:

Manage the systemd-coredump configuration.

* `systemd::manage_coredump` to enable everything
* `systemd::coredump_settings` will manage `/etc/systemd/coredump.conf`
* `systemd::backtrace` if true will add `--backtrace` to the `systemd-coredump` call.

So typical use would be

```puppet
class{'systemd':
  manage_coredump   => true,
  coredump_settings => {
    'Storage' => 'external',
  },
  coredump_backtrace => true,
}
```

to enable storing cores on disk and writing backtraces to
the journal.

### DIFF
--- a/manifests/coredump.pp
+++ b/manifests/coredump.pp
@@ -1,0 +1,22 @@
+# @api private
+# @summary This class manages the systemd-coredump configuration.
+# @see https://www.freedesktop.org/software/systemd/man/systemd-coredump.html
+class systemd::coredump {
+  assert_private()
+
+  $systemd::coredump_settings.each |$option, $value| {
+    ini_setting {
+      "coredump_${option}":
+        path    => '/etc/systemd/coredump.conf',
+        section => 'Coredump',
+        setting => $option,
+        value   => $value,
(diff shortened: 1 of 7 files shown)


manifests/coredump.pp
```diff
@@ -0,0 +1,34 @@
...
+  file { '/etc/sysctl.d/kernel-core_pattern.conf':
+    ensure  => bool2str($systemd::coredump_backtrace, 'file', 'absent'),
+    owner   => root,
+    group   => root,
+    mode    => '0644',
+    content => "# Puppet managed\nkernel.core_pattern=|/usr/lib/systemd/systemd-coredump --backtrace ${_sysctl_args}\n",
+    notify  => Exec['reload_core_pattern'],
+  }
+
+  exec { 'reload_core_pattern':
```
traylenator, line 30: Should probably write a `systemd::sysctl` type first and use that.

bastelfreak (approved):

PR #252, opened 2022-02-17
Don't manage /etc/resolv.conf if systemd-resolved is stopped

### LINKED ISSUE #203: systemd-resolved cannot be fully disabled because /etc/resolv.conf is managed

- Puppet: all
- Ruby: all
- Distribution: Ubuntu 20.04 (client)
- Module version: 3.0.0, 3.1.0

```puppet
class demo {
  class { 'systemd':
    manage_resolved => true,
    resolved_ensure => 'stopped',
  }

  file { '/etc/resolv.conf':
    ensure => 'file',
  }
}
```
## What are you seeing
Catalog compilation fails with with a duplicate resource declaration error.

`File[/etc/resolv.conf]` should not be managed when `systemd-resolved` is managed to the `stopped` state.

## Output log

> Duplicate declaration: File[/etc/resolv.conf] is already declared at (.../modules/systemd/manifests/resolved.pp, line: 77); cannot redeclare (.../modules/demo/manifests/init.pp, line: 7) (file: .../modules/demo/spec/fixtures/modules/demo/manifests/init.pp, line: 7, column: 3) on node ...

Just stopping systemd-resolved on a system that has had it configured completely breaks name resolution.  Something else needs to manage /etc/resolv.conf afterward, or at least to update it once, but the module interferes with that.

For context, on domain-joined Ubuntu 20.04, systemd-resolved needs to be disabled in order for sssd to authenticate users against Active Directory.

Here is the RSpec test used to demonstrate the issue:
```ruby
# frozen_string_literal: true

require 'spec_helper'

describe 'demo' do
  on_supported_os.each do |os, os_facts|
    context "on #{os}" do
      let(:facts) { os_facts.merge(systemd_internal_services: { 'systemd-resolved.service' => 'enabled' }) }

      it { is_expected.to compile }
    end
  end
end
```

traylenator:

If `systemd-resolved` is stopped but managed via

```
class{'systemd:
  managed_resolved => true,
  resolved_ensure  => 'stopped',
}
```
Then the files in `/var/run/systemd/resolv` will either
be stale(just stopped) or missing(after reboot). In both cases it
makes no sense to manage `/etc/resolv.conf`.

More over for the case where `systemd-resolved` has just
been stopped reinstate `/etc/resolv.conf` as `/run/systemd/resolve/resolv.conf`.

Fixes #203

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,18 @@ When configuring `systemd::resolved` you could set `use_stub_resolver` to false 
 
 Systemd has introduced `DNS Over TLS` in the release 239. Currently three states are supported `yes` (since systemd 243), `opportunistic` (true) and `no` (false, default). When enabled with `yes` or `opportunistic` `systemd-resolved` will start a TCP-session to a DNS server with `DNS Over TLS` support. When enabled with `yes` (strict mode), queries will fail if the configured DNS servers do not support `DNS Over TLS`. Note that there will be no host checking for `DNS Over TLS` due to missing implementation in `systemd-resolved`.
 
+Stopping `systemd-resolved` once running can be problematic and care should be taken.
+
+```puppet
+class{'systemd':
+  manage_resolved => true,
+  resolved_ensure => false,
+}
+```
+
+will stop the service and should also copy `/run/systemd/resolve/resolv.conf` to `/etc/resolve.conf`.
+* Writing your own file to `/etc/resolv.conf` is also possible.
+
 It is possible to configure the default ntp servers in `/etc/systemd/timesyncd.conf`:
 
 ```puppet
@@ -298,6 +310,8 @@ when `manage_systemd` is true any required sub package, e.g. `systemd-resolved` 
 systemd-resolved will only occur on second puppet run after that installation.
 
 This requires [puppetlabs-inifile](https://forge.puppet.com/puppetlabs/inifile), which is only a soft dependency in this module (you need to explicitly install it). Both parameters accept a string or an array.
+
+
 
 ### Resource Accounting
 
--- a/manifests/resolved.pp
+++ b/manifests/resolved.pp
@@ -72,14 +72,25 @@ class systemd::resolved (
   }
 
   if $systemd::manage_resolv_conf {
-    $_resolv_conf_target = $use_stub_resolver ? {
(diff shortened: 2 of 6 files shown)


manifests/resolved.pp
```diff
@@ -71,14 +71,25 @@
...
+      target  => $_resolv_conf_target,
+      require => Service['systemd-resolved'],
+    }
+  } else {
+    # If systemd not running make a last ditch attempt to restore
+    # /etc/resolv.conf to something that might actually work on
+    # reboot.
+    exec { 'restore_resolv.conf_if_possible':
+      command => 'cp -f /run/systemd/resolve/resolv.conf /etc/resolv.conf',
+      onlyif  => 'test $(readlink /etc/resolv.conf) = "/run/systemd/resolve/resolv.conf"  || test $(readlink /etc/resolv.conf) = "/run/systemd/resolve/stub-resolv.conf"',
```
ekohl, line 90: I think this is actually rather elegant in that it at least gives a decent chance of working. Implementation wise I do think you should first test it's really a symlink. `readlink` on a regular file is empty so it becomes `test  = "/run/systemd/resolve/resolv.conf"`. Or quote the `$()`.

traylenator: Sure.
Have yet to actually test in real life and indeed above is very broken, it cp's to the synlink destination :-)

README.md
```diff
@@ -284,6 +284,17 @@ When configuring `systemd::resolved` you could set `use_stub_resolver` to false
 
 Systemd has introduced `DNS Over TLS` in the release 239. Currently three states are supported `yes` (since systemd 243), `opportunistic` (true) and `no` (false, default). When enabled with `yes` or `opportunistic` `systemd-resolved` will start a TCP-session to a DNS server with `DNS Over TLS` support. When enabled with `yes` (strict mode), queries will fail if the configured DNS servers do not support `DNS Over TLS`. Note that there will be no host checking for `DNS Over TLS` due to missing implementation in `systemd-resolved`.
 
+Stopping `systemd-resolved` once running can be problematic and care should be taken.
+
+```puppet
+class{'systemd':
+  manage_resolved => true,
+  resolved_ensure => false,
```
bastelfreak, line 292: there is a `}` missing

bastelfreak (approved):

PR #253, opened 2022-02-19
timer: timer unit must depend on service unit.

olifre:
Without this, issues like:
```
cvmfs-fsck.timer: Refusing to start, unit cvmfs-fsck.service to trigger not loaded.
```
may be observed.

### DIFF
--- a/spec/defines/timer_spec.rb
+++ b/spec/defines/timer_spec.rb
@@ -29,7 +29,7 @@ describe 'systemd::timer' do
           it {
             expect(subject).to contain_systemd__unit_file('foobar.service').with(
               content: "[Service]\nExecStart=/bin/touch /tmp/foobar"
-            )
+            ).that_comes_before('Systemd::Unit_file[foobar.timer]')
           }
         end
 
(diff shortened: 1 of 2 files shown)


kenyon:
Should we update https://github.com/voxpupuli/puppet-systemd/blob/master/spec/defines/timer_spec.rb with this too?

olifre:
That seems like a good idea, I'll add a `that_comes_before` there ;-).

olifre:
@kenyon I'm not sure why the `spec` tests fail — can you give me a pointer?

bastelfreak:
mhm, odd that Inuwed this pattern a lot in the past and never run into the error

spec/defines/timer_spec.rb
```diff
@@ -29,7 +29,7 @@
           it {
             expect(subject).to contain_systemd__unit_file('foobar.service').with(
               content: "[Service]\nExecStart=/bin/touch /tmp/foobar"
-            )
+            ).that_comes_before('Systemd::Unit_File[\'foobar.timer\']')
```
ekohl, line 32: No need to quote this. I if you apply this everywhere the tests will pass
```suggestion
            ).that_comes_before('Systemd::Unit_File[foobar.timer]')
```

olifre: Many thanks for pointing this out! That was unexpected, I thought the syntax would be similar to that used in Puppet directly. 
Fixed and force-pushed :-).

ekohl (approved):
This is actually correct in that systemd requires it. It's probably that the definition order generally speaking is "good enough". This is explicit and thus better. :+1: if the tests are green.

spec/defines/timer_spec.rb
olifre: It seems tests still did not like this, I changed:
`Systemd::Unit_File` => `Systemd::Unit_file` now, assuming capitalization needs to match explicitly. Let's see if that makes things more green.

olifre: That did the trick :smile:.

kenyon (approved):

PR #255, opened 2022-02-21
Addition of Trivial Acceptance Tests

traylenator:
Addition of trivial acceptance tests

### DIFF
--- a/spec/acceptance/resolved_spec.rb
+++ b/spec/acceptance/resolved_spec.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'spec_helper_acceptance'
+
+describe 'systemd' do
+  context 'configure systemd resolved' do
+    it 'works idempotently with no errors' do
+      pp = <<-PUPPET
+      class{'systemd':
+        manage_resolved    => true,
+        manage_resolv_conf => #{default[:hypervisor] != 'docker'},
+      }
+      PUPPET
+      apply_manifest(pp, catch_failures: true)
+      # RedHat 9 and newer installs package first run before fact  $facts['internal_services'] is set
+      apply_manifest(pp, catch_failures: true) if Gem::Version.new(fact('os.release.major')) >= Gem::Version.new('9') && (fact('os.family') == 'RedHat')
+      apply_manifest(pp, catch_changes: true)
+    end
+
+    # RedHat 7 does not have systemd-resolved available at all.
+    describe service('systemd-resolved'), unless: (fact('os.release.major') == '7' and fact('os.family') == 'RedHat') do
+      it { is_expected.to be_running }
+      it { is_expected.to be_enabled }
+    end
+  end
+end
--- a/spec/acceptance/systemd_spec.rb
+++ b/spec/acceptance/systemd_spec.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+describe 'systemd' do
+  it_behaves_like 'an idempotent resource' do
+    let(:manifest) do
+      <<-PUPPET
+      include systemd
+      PUPPET
+    end
+  end
+end
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,9 +57,33 @@ jobs:
       - name: Run tests
         run: bundle exec rake parallel_spec
 
+  acceptance:
+    needs: setup_matrix
+    runs-on: ubuntu-latest
+    env:
+      BUNDLE_WITHOUT: development:test:release
+    strategy:
+      fail-fast: false
+      matrix:
+        include: ${{fromJson(needs.setup_matrix.outputs.github_action_test_matrix)}}
+    name: ${{ matrix.puppet.name }} - ${{ matrix.setfile.name }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.0'
+          bundler-cache: true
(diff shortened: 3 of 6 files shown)


bastelfreak:
We need a new modulesync to update https://github.com/voxpupuli/puppet-systemd/blob/master/.github/workflows/ci.yml. when the tests are present and modulesync is run, the CI file is updated to run the acceptance tests.

bastelfreak (approved):

spec/acceptance/systemd_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+require 'spec_helper_acceptance'
+
+describe 'systemd' do
+  it_behaves_like 'an idempotent resource' do
+    let(:manifest) do
+      <<-PUPPET
+      include systemd
+      PUPPET
+    end
```
ekohl: Only partly related discussion and you probably know this, but the short version of this is:
```suggestion
    let(:manifest) { 'include systemd' }
```

I wonder if we should encourage one style over the other or just see what individuals prefer.

ekohl (commented):
I like this. Acceptance tests are so useful.

traylenator:
This is what are feared, you very quickly start running to weird things about these hosts:

```
Error: /Stage[main]/Systemd::Resolved/File[/etc/resolv.conf]/ensure: change from 'file' to 'link' failed: Device or resource busy @ apply2files - /etc/resolv.conf
```

spec/acceptance/systemd_spec.rb
traylenator: There HERE document has the advantage that you can correct puppet style quoting once the line becomes more interesting.

ekohl: I agree and I think for consistency that's why I also commonly use heredoc. It makes it easy to add multiple lines without reformatting the whole block.

spec/acceptance/resolved_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    it 'works idempotently with no errors' do
+      pp = <<-PUPPET
+      class{'systemd':
+        manage_resolved    => true,
+        manage_resolv_conf => false,
+      }
+      PUPPET
+      apply_manifest(pp, catch_failures: true)
+      # RedHat 9 requires an extra run
+      apply_manifest(pp, catch_failures: true)
```
bastelfreak, line 16: can you wrap that in a condition so it only runs twice on RHEL9? Also why are two runs required? that shouldn't be the case at all?

traylenator:
That;s probably enough in here for now:

* I'll pull out the `manage_resolv_conf` to it's own MR.
* Disable the checks on CentOS as there is indeed no resolved to be had.

spec/acceptance/resolved_spec.rb
ekohl: Perhaps rephrase it to "extra run to populate $facts['internal_services']"?

traylenator:
#256 first for the extra parameter.

spec/acceptance/resolved_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+        manage_resolv_conf => #{default[:hypervisor] != 'docker'},
+      }
+      PUPPET
+      apply_manifest(pp, catch_failures: true)
+      # RedHat 9 and newer installs package first run before fact  $facts['internal_services'] is set
+      apply_manifest(pp, catch_failures: true) if !%w[7 8].include?(fact('os.release.major')) && (fact('os.family') == 'RedHat')
+      apply_manifest(pp, catch_changes: true)
+    end
+
+    # RedHat 7 does not have systemd-resolved available at all.
```
ekohl, line 20: On CentOS 7 there is a `systemd-resolved` package with a `/usr/lib/systemd/system/systemd-resolved.service`. In practice I noticed it was harder to use due to a lack of `resolvectl` so I ended up not using it, but it's not entirely correct. I'd be surprised if RHEL 7 and CentOS 7 differed in this.

traylenator: Indeed that package  is available CentOS 7 also, just not in 8. :-(

spec/acceptance/resolved_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    it 'works idempotently with no errors' do
+      pp = <<-PUPPET
+      class{'systemd':
+        manage_resolved    => true,
+        manage_resolv_conf => #{default[:hypervisor] != 'docker'},
+      }
+      PUPPET
+      apply_manifest(pp, catch_failures: true)
+      # RedHat 9 and newer installs package first run before fact  $facts['internal_services'] is set
+      apply_manifest(pp, catch_failures: true) if Gem::Version.new(fact('os.release.major')) >= Gem::Version.new('9') && (fact('os.family') == 'RedHat')
```
ekohl, line 16: This would be more explicit. Not insisting on it, but it could help the reader?
```suggestion
      separate_resolved_package = Gem::Version.new(fact('os.release.major')) >= Gem::Version.new('9') && (fact('os.family') == 'RedHat')
      apply_manifest(pp, catch_failures: true) if separate_resolved_package
```

spec/acceptance/resolved_spec.rb
ekohl: Does that mean we should include the package name in the Hiera data?

spec/acceptance/resolved_spec.rb
traylenator: With the comment I think I prefer, seems more obvious what to delete when C8 is dead.

PR #256, opened 2022-02-21
New parmater manage_resolv_conf for /etc/resolv.conf

traylenator:
If the parameter is true (default) then `/etc/resolv.conf` will
be managed when `manage_resolved` is true.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -81,6 +81,7 @@ The following parameters are available in the `systemd` class:
 * [`dnsovertls`](#dnsovertls)
 * [`cache`](#cache)
 * [`dns_stub_listener`](#dns_stub_listener)
+* [`manage_resolv_conf`](#manage_resolv_conf)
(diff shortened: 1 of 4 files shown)


ekohl (approved):
I wonder if this could also be a partial solution to https://github.com/voxpupuli/puppet-systemd/issues/203.

traylenator:
> I wonder if this could also be a partial solution to #203.

It's an alternate solution to document with this parameter false in addition 

But 

```puppet
class{'systemd:
    manage_resolved => true,
    resolved_ensure => false,
 ```
 would still result in a dead box which is what I was really trying to avoid really with #203.
 
 ( #203 will need rebasing after this I'm sure )

ekohl:
Time to do that rebase then I guess ;) And it's indeed what I was saying with "partial". A user could also have a profile module that then manages `resolv.conf` which would be good enough for them.

PR #259, opened 2022-03-01
dropin_file: Implement service_parameters hash

bastelfreak:
This allows people to pass all options the service resource type accepts
to `systemd::dropin_file`. This is required to enable people to make a
migration from legacy code with systemd reload + file/service resource
to this module.

### DIFF
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -58,6 +58,9 @@
 # @param selinux_ignore_defaults
 #   maps to the same param on the file resource for the unit. false in the module because it's false in the file resource type
 #
+# @param service_parameters
+#   hash that will be passed with the splat operator to the service resource
+#
 # @example manage unit file + service
 #   systemd::unit_file { 'foo.service':
 #     content => file("${module_name}/foo.service"),
@@ -81,6 +84,7 @@ define systemd::unit_file (
   Optional[Boolean]                        $hasrestart = undef,
   Boolean                                  $hasstatus = true,
   Boolean                                  $selinux_ignore_defaults = false,
+  Hash[String[1], Any]                     $service_parameters = {},
 ) {
   include systemd
 
@@ -121,6 +125,7 @@ define systemd::unit_file (
       provider   => 'systemd',
       hasrestart => $hasrestart,
       hasstatus  => $hasstatus,
+      *          => $service_parameters,
     }
 
     if $ensure == 'absent' {
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1224,6 +1224,7 @@ The following parameters are available in the `systemd::unit_file` defined type:
 * [`hasrestart`](#hasrestart)
 * [`hasstatus`](#hasstatus)
 * [`selinux_ignore_defaults`](#selinux_ignore_defaults)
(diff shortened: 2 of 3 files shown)


manifests/unit_file.pp
```diff
@@ -121,6 +125,7 @@
       provider   => 'systemd',
       hasrestart => $hasrestart,
       hasstatus  => $hasstatus,
```
ekohl, line 127: I do think we should deprecate these parameters in favor of `$service_parameters`.

manifests/unit_file.pp
```diff
@@ -81,6 +84,7 @@
   Optional[Boolean]                        $hasrestart = undef,
   Boolean                                  $hasstatus = true,
   Boolean                                  $selinux_ignore_defaults = false,
+  Hash                                     $service_parameters = {},
```
ekohl, line 87: Didn't check the whitespace, but how about this?
```suggestion
  Hash[String[1], Any]                                     $service_parameters = {},
```

ekohl (approved):

manifests/unit_file.pp
bastelfreak: I considered it, but that would be a breaking change. I propose we merge this PR and I create a new one that's a breaking change and we merge that in the future?

manifests/unit_file.pp
bastelfreak: ah yes, implemented it.

manifests/unit_file.pp
bastelfreak: we cold also already implement something like `warn('hastatus is deprecated')`?

ekohl: I said deprecate, not remove :)

So what you can do is to add a note in the `@param` tag. Then in the code:
```puppet
$deprecated_service_params = {
  'hasrestart' => $hasrestart,
  'hasstatus'  => $hasstatus,
}

service { '...':
  * => $deprecated_service_params + $service_params,
}
```

bastelfreak: I implemented this pattern already. For a clearer changelog I would like to add the deprecation in another PR.

manifests/unit_file.pp
```diff
@@ -114,13 +118,16 @@
...
-      enable     => $enable,
-      restart    => $restart,
-      provider   => 'systemd',
-      hasrestart => $hasrestart,
-      hasstatus  => $hasstatus,
+      ensure   => $active,
+      enable   => $enable,
+      restart  => $restart,
+      provider => 'systemd',
+      *        => $service_parameters + $deprecated_service_params,
```
ekohl, line 130: What takes precedence here? The deprecated ones or the new ones?

ekohl (changes requested):
Could you add notes to `@param hasrestart` and `@param hasstatus` that they're deprecated and that they should be specified in `$service_parameters`?

manifests/unit_file.pp
bastelfreak: I reverted this for now. Can we implement the deprecation in another PR? That would also make a cleaner changelog.

kenyon (approved):

PR #260, opened 2022-03-01
systemd::udev::rule: param rules now defaults to `[]` / fix broken tests

bastelfreak:

### DIFF
--- a/manifests/udev/rule.pp
+++ b/manifests/udev/rule.pp
@@ -23,7 +23,7 @@
 #   The literal udev rules you want to deploy
 #
 define systemd::udev::rule (
-  Array                             $rules,
+  Array                             $rules                   = [],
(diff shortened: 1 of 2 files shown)


manifests/udev/rule.pp
```diff
@@ -23,8 +23,8 @@
 #   The literal udev rules you want to deploy
 #
 define systemd::udev::rule (
-  Array                             $rules,
-  Enum['present', 'absent', 'file'] $ensure                  = 'present',
+  Array                             $rules                   = [],
```
bastelfreak, line 26: I changed this so it always has a value. if you use the defined resource to remove a udev rule you now only have to set `$ensure` to absent. Without this change you also would need to set `$rules`

ekohl (approved):

PR #261, opened 2022-03-02
unit_file: deprecate hasrestart/hasstatus params

bastelfreak:
We now have a `$service_parameters` hash that's passed to the service
resource. It should be used to set hasrestart/hasstatus.

### DIFF
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -50,10 +50,10 @@
 #   Specify a restart command manually. If left unspecified, a standard Puppet service restart happens.
 #
 # @param hasrestart
-#   maps to the same param on the service resource. Optional in the module because it's optional in the service resource type
+#   maps to the same param on the service resource. Optional in the module because it's optional in the service resource type. This param is deprecated. Set it via $service_parameters.
 #
 # @param hasstatus
-#   maps to the same param on the service resource. true in the module because it's true in the service resource type
+#   maps to the same param on the service resource. true in the module because it's true in the service resource type. This param is deprecated. Set it via $service_parameters.
 #
 # @param selinux_ignore_defaults
 #   maps to the same param on the file resource for the unit. false in the module because it's false in the file resource type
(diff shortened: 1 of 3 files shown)


manifests/unit_file.pp
```diff
@@ -90,6 +90,14 @@
 
   assert_type(Systemd::Unit, $name)
 
+  if $hasrestart {
```
ekohl, line 93: Since it's a boolean, shouldn't this be check if it's not undef:
```suggestion
  if $hasrestart =~ NotUndef {
```

manifests/unit_file.pp
```diff
@@ -90,6 +90,14 @@
...
+    warning("systemd::unit_file - ${title}: hasrestart is deprecated and will be removed in Version 4 of the module")
+  }
+
+  if $hasstatus == false {
```
ekohl, line 97: Should we actually make it `Optional[Boolean]`? The default of `hasstatus` in `service` is `true` anyway and it allows you to find out if it was passed in explicitly, which is a problem.

manifests/unit_file.pp
```diff
@@ -90,6 +90,14 @@
 
   assert_type(Systemd::Unit, $name)
 
+  if $hasrestart {
+    warning("systemd::unit_file - ${title}: hasrestart is deprecated and will be removed in Version 4 of the module")
```
ekohl, line 94: Does it make sense to use `deprecation()` from stdlib for this?

manifests/unit_file.pp
bastelfreak: ah yes

manifests/unit_file.pp
bastelfreak: I wasn't aware of that. I will take a look at what it does and use it

manifests/unit_file.pp
bastelfreak: good idea, implemented it.

manifests/unit_file.pp
bastelfreak: I implemented `deprecation()`. Depending on the puppetserver configuration nothing will be logged:
https://github.com/puppetlabs/puppetlabs-stdlib/blob/ea43f2ecd217ff85fe6797b177c7df7718266a94/lib/puppet/functions/deprecation.rb#L29-L38

should we keep the `warning()` because of that, or remove it?

ekohl: I'd say remove it. @alexjfisher usually has some good ideas about this though.

bastelfreak: removed it for now..

ekohl (approved):

PR #270, opened 2022-04-18
Add parameter to manage default target

jcpunk:
I was a bit surprised to discover that the default target for systemd was not already something this class managed.

Merging this should probably result in a meaningful version bump since folks who want a different target will need to set one explicitly.  The default from systemd upstream is multi-user.target thus that is what I selected.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -2,6 +2,9 @@
 #
 # @api public
 #
+# @param default_target
+#   The default systemd boot target, unmanaged if set to undef.
+#
 # @param service_limits
 #   May be passed a resource hash suitable for passing directly into the
 #   ``create_resources()`` function as called on ``systemd::service_limits``
@@ -160,6 +163,7 @@
 #   Add --backtrace to systemd-coredump call systemd-coredump@.service unit
 #
 class systemd (
+  Optional[Pattern['^.+\.target$']]                   $default_target = undef,
   Hash[String,String]                                 $accounting = {},
   Hash[String[1],Hash[String[1], Any]]                $service_limits = {},
   Hash[String[1],Hash[String[1], Any]]                $networks = {},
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -210,6 +214,16 @@
 ) {
   contain systemd::install
 
+  service { $default_target:
+    enable =>  true
```
kenyon, line 218:
```suggestion
    enable => true,
```

kenyon (approved):

manifests/init.pp
```diff
@@ -160,6 +163,7 @@
 #   Add --backtrace to systemd-coredump call systemd-coredump@.service unit
 #
 class systemd (
+  Pattern['^.+\.target$']                             $default_target = 'multi-user.target',
```
smortex, line 166: Per se this is a breaking change (people with a non-default config will lose it). The title of the PR makes me think the default should be `undef` and only manage the default target if a value is specifically set.

I think we should adjust the PR title or the code so that they match :wink:

manifests/init.pp
```diff
@@ -210,6 +214,17 @@
 ) {
   contain systemd::install
 
+  service { $default_target:
+    enable => true,
+    ensure => 'running',
```
smortex, line 219: Fix CI by setting `ensure` first.

manifests/init.pp
```diff
@@ -210,6 +214,17 @@
...
+  }
+
+  exec { "systemctl set-default ${default_target}":
+    command => "systemctl set-default ${default_target}",
+    unless  => "test \"$(systemctl get-default)\" = \"${default_target}\"",
```
smortex, line 224: `default_target` should probably be shell escaped :trollface:

```puppet
class { 'systemd':
  default_target => "$(cp /etc/passwd /var/www/html)multi-user.target"
}
```

manifests/init.pp
smortex: Sure, but then we need a `Boolean $manage_default_target = true` to allow users who do *not* want to manage this to do as they which by setting it to false explicitely.

I trend to prefer a single parameter that accept undef and is not a breaking change :shrug:

manifests/init.pp
jcpunk: good call, this look better?

jcpunk:
I've reworked it as a non-breaking change using `undef` as the default-target.  This look better?

manifests/init.pp
```diff
@@ -2,6 +2,10 @@
 #
 # @api public
 #
+# @param default_target
+#   The default systemd boot target, unmanaged if set to undef.
+#   Default: undef
```
kenyon, line 7: No need to write the default here, Puppet Strings puts that in REFERENCE.md automatically.

root-expert (approved):
LGTM

jcpunk:
I'd love to see this merged in...

PR #272, opened 2022-04-19
Add machine-info information management

jcpunk:
It can be handy to set elements of `/etc/machine-info` from some facts.  For example. if you're using Foreman's location features, being able to populate the location field from that is very nice.


manifests/machine_info.pp
```diff
@@ -0,0 +1,14 @@
+# @api private
+# @summary This class manages systemd's machine-info file (hostnamectl)
+# @see https://www.freedesktop.org/software/systemd/man/machine-info.html
+class systemd::machine_info {
+  assert_private()
+
+  $systemd::machine_info_settings.each |$option, $value| {
+    file_line { "machine-info ${option}":
```
bastelfreak, line 8: It can get quite tricky to prevent duplicate lines with file_line. I suggest using https://forge.puppet.com/modules/herculesteam/augeasproviders_shellvar

jcpunk: I was concerned about adding a new dependency to this module, but if you're not worried I'll get that in place.

manifests/init.pp
```diff
@@ -257,6 +261,10 @@
     contain systemd::system
   }
 
+  if $machine_info_settings {
```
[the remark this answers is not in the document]
jcpunk: fixed

manifests/machine_info.pp
jcpunk: I've reworked with shellvars and pushed up.  If that looks semi-viable, I can drop the `file_line` implementation.

root-expert (commented):
Do we consider the new dependency as hard?
Because you won't normally need it if you don't manage machine-info

jcpunk:
I don't know how to specify weak deps in `metadata.json`...

smortex:
> I don't know how to specify weak deps in `metadata.json`...

You just don't list them in `metadata.json`… :shrug:

jcpunk:
In theory, this is probably ready...

bastelfreak (approved):

root-expert (approved):

PR #277, opened 2022-06-10
Fix systemctl daemon-reload after file additions

### LINKED ISSUE #234: systemd::dropin_file doesn't cause a systemd daemon-reload

trevor-vaughan:
* Add a `systemd::daemon_reload` defined type
* Ensure that any file additions to the /etc/systemd/system space are
  followed by a call to `systemd::daemon_reload`
* Allow users to disable the calls to `systemd::daemon_reload`
* Allow users to globally disable the `systemctl daemon-reload` exec
  using a resource collector if necessary
* Hook the daemon reload between the file creation and the service as is
  usually necessary, where possible
* Add a 'best effort' optional exec as `systemd::lazy_daemon_reload` to
  try and clean up systems that were modified betweedn 2.9.0 and this
  release

Fixes #234
Fixes #199

### DIFF
--- a/manifests/daemon_reload.pp
+++ b/manifests/daemon_reload.pp
@@ -11,7 +11,7 @@
 #   * Added in case users want to disable the reload globally using a resource collector
 #
 define systemd::daemon_reload (
-  Boolean $enable = true
+  Boolean $enable = true,
 ) {
   if $enable {
     exec { "${module_name}-${name}-systemctl-daemon-reload":
--- a/manifests/daemon_reload.pp
+++ b/manifests/daemon_reload.pp
@@ -10,29 +10,10 @@
 #
 #   * Added in case users want to disable the reload globally using a resource collector
 #
-# @param lazy_reload
-#   Enable a global lazy reload.
-#
-#   * This is meant for cleaning up systems that may have gotten out of sync so no particular
-#     care is taken in trying to actually "fix" things, that would require a custom type that
-#     manipulates the running catalog tree.
-#
 define systemd::daemon_reload (
-  Boolean $enable      = true,
-  Boolean $lazy_reload = false,
+  Boolean $enable = true
 ) {
   if $enable {
-    if $lazy_reload {
-      exec { "${module_name}-${name}-global-systemctl-daemon-check":
-        command => 'systemctl daemon-reload',
-        onlyif  => 'systemctl show "*" --property=NeedDaemonReload | grep -qxFm1 "NeedDaemonReload=yes"',
-        path    => $facts['path'],
-      }
-
-      # Give services a fighting chance of coming up properly
-      Exec["${module_name}-${name}-global-systemctl-daemon-check"] -> Service <||>
-    }
-
     exec { "${module_name}-${name}-systemctl-daemon-reload":
       command     => 'systemctl daemon-reload',
       refreshonly => true,
--- a/manifests/daemon_reload.pp
+++ b/manifests/daemon_reload.pp
@@ -25,7 +25,7 @@ define systemd::daemon_reload (
     if $lazy_reload {
       exec { "${module_name}-${name}-global-systemctl-daemon-check":
         command => 'systemctl daemon-reload',
-        onlyif  => 'systemctl show "*" --property=NeedDaemonReload | grep -q "=yes"',
+        onlyif  => 'systemctl show "*" --property=NeedDaemonReload | grep -qxFm1 "NeedDaemonReload=yes"',
         path    => $facts['path'],
       }
 
--- a/manifests/daemon_reload.pp
+++ b/manifests/daemon_reload.pp
@@ -1,0 +1,42 @@
+# @summary Run systemctl daemon-reload
+#
+# @api public
+#
+# @param name
+#   A globally unique name for the resource
+#
+# @param enable
+#   Enable the reload exec
+#
+#   * Added in case users want to disable the reload globally using a resource collector
+#
+# @param lazy_reload
+#   Enable a global lazy reload.
+#
+#   * This is meant for cleaning up systems that may have gotten out of sync so no particular
+#     care is taken in trying to actually "fix" things, that would require a custom type that
+#     manipulates the running catalog tree.
+#
+define systemd::daemon_reload (
+  Boolean $enable      = true,
+  Boolean $lazy_reload = false,
+) {
+  if $enable {
+    if $lazy_reload {
+      exec { "${module_name}-${name}-global-systemctl-daemon-check":
+        command => 'systemctl daemon-reload',
+        onlyif  => 'systemctl show "*" --property=NeedDaemonReload | grep -q "=yes"',
+        path    => $facts['path'],
+      }
+
+      # Give services a fighting chance of coming up properly
+      Exec["${module_name}-${name}-global-systemctl-daemon-check"] -> Service <||>
+    }
+
+    exec { "${module_name}-${name}-systemctl-daemon-reload":
+      command     => 'systemctl daemon-reload',
+      refreshonly => true,
+      path        => $facts['path'],
+    }
+  }
+}
--- a/manifests/dropin_file.pp
+++ b/manifests/dropin_file.pp
@@ -72,16 +72,16 @@ define systemd::dropin_file (
   }
 
   if $daemon_reload {
-    ensure_resource('systemd::daemon_reload', $name)
+    ensure_resource('systemd::daemon_reload', $unit)
 
-    File[$full_filename] ~> Systemd::Daemon_reload[$name]
+    File[$full_filename] ~> Systemd::Daemon_reload[$unit]
   }
 
   if $notify_service {
     File[$full_filename] ~> Service <| title == $unit or name == $unit |>
 
     if $daemon_reload {
-      Systemd::Daemon_reload[$name] ~> Service <| title == $unit or name == $unit |>
+      Systemd::Daemon_reload[$unit] ~> Service <| title == $unit or name == $unit |>
     }
 
     if $unit =~ /\.service$/ {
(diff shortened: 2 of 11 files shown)


trevor-vaughan:
@op-ct For reference. Fixes the issues with simp/pupmod-simp-gdm/pull/77.

manifests/daemon_reload.pp
```diff
@@ -0,0 +1,42 @@
...
+#
+define systemd::daemon_reload (
+  Boolean $enable      = true,
+  Boolean $lazy_reload = false,
+) {
+  if $enable {
+    if $lazy_reload {
+      exec { "${module_name}-${name}-global-systemctl-daemon-check":
+        command => 'systemctl daemon-reload',
+        onlyif  => 'systemctl show "*" --property=NeedDaemonReload | grep -q "=yes"',
```
simondeziel, line 28: May I suggests using `systemctl show "*" --property=NeedDaemonReload | grep -qxFm1 "NeedDaemonReload=yes"`

trevor-vaughan: @simondeziel Since it's the only field being returned we don't need an exact match. If there's a `=yes` anywhere in there, it's go time.

trevor-vaughan: 🤔 Definitely could stop at the first match though

trevor-vaughan: Yeah, the efficiency makes sense. Will update.

simondeziel: That is definitely nitpick territory but just found out about the `--value` option: `systemctl show "*" --value --property=NeedDaemonReload | grep -qxFm1 "yes"`

trevor-vaughan: @simondeziel Let me see if it works on EL7

trevor-vaughan: `systemd` has added so many options over time I get nervous about any but the basics across multiple OS versions

simondeziel: TIL, thanks for testing!

manifests/dropin_file.pp
```diff
@@ -69,11 +71,26 @@
     show_diff               => $show_diff,
   }
 
+  if $daemon_reload {
+    ensure_resource('systemd::daemon_reload', $name)
```
ekohl, line 75: I think this should be one per unit, in case there are multiple dropin_files per unit. Possibly also normalized to a short name (like on line 88)
```suggestion
    ensure_resource('systemd::daemon_reload', $unit)
```

trevor-vaughan: That works. Fewer resources in the catalog that way.

manifests/daemon_reload.pp
```diff
@@ -0,0 +1,42 @@
...
+#   * This is meant for cleaning up systems that may have gotten out of sync so no particular
+#     care is taken in trying to actually "fix" things, that would require a custom type that
+#     manipulates the running catalog tree.
+#
+define systemd::daemon_reload (
+  Boolean $enable      = true,
+  Boolean $lazy_reload = false,
+) {
+  if $enable {
+    if $lazy_reload {
```
ekohl, line 25: This looks odd in a define. Why would you need multiple instances of this? It feels to me this belongs in a class instead.

alexjfisher: I'm guessing because you might have to use it more than once for 2 separate services where one of those services also depends on the other??

trevor-vaughan: @ekohl You usually don't but, if for some reason, you want to hit *all* declarations of the defined type without having to include an extra class in the defined type (and potentially risk dependency loops), this gives you a way to do it.

Given that the only reason that is there is to try and fix the mess from 2.9.0 to now, it seemed reasonable.

trevor-vaughan: Honestly, it would probably work better with a dose of `simplib::dlookup()` but I don't know that the world is ready for that level of fun yet.

trevor-vaughan:
@ekohl and/or @alexjfisher Are you all OK with this?

manifests/daemon_reload.pp
ekohl: Still, there is no determination where it shows up. It happens before **all** services, but you have no idea where it shows up. Worst, it also creates a second exec that ends up in the catalog but nothing that refreshes it. It still feels like something I can't wrap my head around. If the `refreshonly` exec is in the `else` part of `if $lazy_reload` it would make more sense to me.

trevor-vaughan: @ekohl Did this make sense or do I need to remove the parameter for a sign-off?

manifests/init.pp
```diff
@@ -316,4 +320,8 @@
       * => $resource,
     }
   }
+
+  if $lazy_daemon_reload {
+    systemd::daemon_reload { 'global-lazy': lazy_reload => $lazy_daemon_reload }
```
ekohl, line 325: Perhaps this is easier to follow for the reader, though technically the same.
```suggestion
    systemd::daemon_reload { 'global-lazy':
      lazy_reload => true,
     }
```

manifests/daemon_reload.pp
trevor-vaughan: @ekohl Well, it will show up directly before all of the services :-D.

Honestly, I think I'll just remove it since it's causing issues with getting the fix through.

trevor-vaughan:
@ekohl I removed the lazy reload and am creating a ticket for follow-on work to fix it properly.

trevor-vaughan:
Proper management of NFS also needs this fix.

manifests/daemon_reload.pp
```diff
@@ -0,0 +1,23 @@
...
+# @param name
+#   A globally unique name for the resource
+#
+# @param enable
+#   Enable the reload exec
+#
+#   * Added in case users want to disable the reload globally using a resource collector
+#
+define systemd::daemon_reload (
+  Boolean $enable = true
```
ekohl, line 14: Small nit: I'd have thought our linting would ensure a trailing comma
```suggestion
  Boolean $enable = true,
```

manifests/timer.pp
```diff
@@ -72,6 +75,7 @@
   Boolean                                  $show_diff = true,
   Optional[Variant[Boolean, Enum['mask']]] $enable = undef,
   Optional[Boolean]                        $active = undef,
+  Boolean                                  $daemon_reload = true,
```
ekohl, line 78: If `$active == true`, the daemon-reload is redundant, right? Wouldn't Puppet itself handle this since it inspects whether a reload is needed before (re)starting a service? And `$active == true` implies the service is managed.

spec/classes/init_spec.rb
```diff
@@ -21,6 +21,7 @@
         it { is_expected.not_to contain_class('systemd::coredump') }
         it { is_expected.not_to contain_class('systemd::oomd') }
         it { is_expected.not_to contain_exec('systemctl set-default multi-user.target') }
+        it { is_expected.not_to contain_systemd__daemon_reload('global-lazy') }
```
ekohl, line 24: This looks like a remainder of the old code.
```suggestion
```

spec/defines/daemon_reload.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        it { is_expected.to compile.with_all_deps }
+
+        context 'with defaults' do
+          it do
+            expect(subject).to contain_exec("systemd-#{title}-systemctl-daemon-reload").
+              with_command('systemctl daemon-reload').
+              with_refreshonly(true)
+
+            expect(subject).not_to contain_exec("systemd-#{title}-global-systemctl-daemon-check")
```
ekohl, line 20: Remainder of the old code?
```suggestion
```

ekohl (commented):
Some small nits. We do have a few more redundant daemon-reload calls now which does worry me a bit. Puppet 6.1.0 and higher should automatically call systemctl daemon-reload if needed on a managed service. Do you have drop in files for unmanaged services that you run into this or is the deamon-reload code in Puppet broken? (I did run into https://github.com/theforeman/puppet-puppet/issues/832 but that was only on Ubuntu 20.04).

I think I can live with some redundant daemon-reloads but there are a few small code suggestions inline that I'd like to see applied.

manifests/timer.pp
trevor-vaughan: The daemon-reload generally needs to be sandwiched between a file drop and a subsequent service.

There's no good way to ensure that we only have to call it once without a very high risk of dependency loops.

trevor-vaughan: For `active` vs `daemon_reload`:

* We need to call `daemon_reload` after the file no matter what
* If the service is managed, we need to call it before the service
* If, for whatever reason, you *don't* want to call it (I don't know why, but I was trying to preserve the current state if users wanted it) then we need to do neither
That's what's currently implemented in `unit_file.pp`

trevor-vaughan: @ekohl Puppet might be doing fine but items like GDM and NFS need refreshes at very specific times which may be prior to the service loading.

trevor-vaughan: At any rate, I couldn't get it to work consistently without this chain in place. Might be able to debug later but this fixes several items for now.

trevor-vaughan: Convo bounced through IRC and into https://github.com/voxpupuli/puppet-systemd/issues/284

ekohl (approved):
As discussed partially on IRC (captured in #284) this looks good to me. I'd label this an enhancement since IMHO it deserves a minor version bump rather than just a patch bump.

ekohl:
Oh, and can we squash this? Can be done on merge.

@alexjfisher any last thoughts?

PR #280, opened 2022-06-17
systemd::timer: move variable definition close to where it is used

simondeziel:


manifests/timer.pp
```diff
@@ -75,13 +75,12 @@
...
-    $_service_unit = $service_unit
-  } else {
-    $_service_unit = "${basename($name,'.timer')}.service"
-  }
-
   if $service_content or $service_source {
+    $_service_unit = $service_unit ? {
+      undef   => "${basename($name,'.timer')}.service",
+      default => $service_unit,
+    }
```
ekohl, line 82: Shouldn't this be using `pick()`?
```suggestion
    $_service_unit = pick($service_unit, "${basename($name,'.timer')}.service")
```

ekohl (approved):
Potential further improvement inline.

PR #289, opened 2022-06-29
Correct docs for name var of systemd::dropin_file

traylenator:
The filename and unit parameter were documented incorrectly
as to which was the name variable.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -637,13 +637,13 @@ The following parameters are available in the `systemd::dropin_file` defined typ
 
 Data type: `Systemd::Unit`
 
-the The target unit file to create, the value will be set to the `filename` parameter as well
+The target unit file to create
 
 ##### <a name="filename"></a>`filename`
 
 Data type: `Systemd::Dropin`
 
-The target unit file to create
+The filename of the drop in. The full path is determined using the path, unit and this filename.
 
 Default value: `$name`
 
--- a/manifests/dropin_file.pp
+++ b/manifests/dropin_file.pp
@@ -4,8 +4,8 @@
 #
 # @see systemd.unit(5)
 #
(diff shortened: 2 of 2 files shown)


manifests/dropin_file.pp
```diff
@@ -4,8 +4,8 @@
 #
 # @see systemd.unit(5)
 #
-# @param unit the The target unit file to create, the value will be set to the `filename` parameter as well
-# @param filename The target unit file to create
+# @param unit the The target unit file to create
```
ekohl, line 7: Looks like there was also another issue:
```suggestion
# @param unit The target unit file to create
```

REFERENCE.md
```diff
@@ -637,13 +637,13 @@ The following parameters are available in the `systemd::dropin_file` defined typ
...
+the The target unit file to create
 
 ##### <a name="filename"></a>`filename`
 
 Data type: `Systemd::Dropin`
 
-The target unit file to create
+The target unit file to create, the value will default to the name of the type
 
 Default value: `$name`
```
ekohl, line 648: I wonder if this isn't redundant. It already generates the default value so in my modules I tend to avoid describing the default value unless it really is special. Perhaps it would be more useful to describe that it's really the filename and that the full path is a combination of `$path`, $unit` and `$filename`.

manifests/dropin_file.pp
```diff
@@ -4,8 +4,8 @@
...
+# @param filename The target unit file to create, the value will default to the name of the type
```
ekohl, line 8: Following through about my previous comment, perhaps it should be something like:
```suggestion
# @param filename The filename of the drop in. The full path is determined using the path, unit and this filename.
```

ekohl (commented):
Very valid point to improve the docs. They were really incorrect.

smortex (approved):

PR #290, opened 2022-07-01
Make Systemd::Unit type stricter

traylenator:

Previously lots of unit names like

```
this is a service with spaces in.service
```
were permitted for instance.

From systemd.unit

> Valid unit names consist of a "name prefix" and a dot and a suffix specifying the unit type. The "unit prefix" must consist of one or more valid characters (ASCII letters, digits, ":", "-", "_", ".", and "\"). The total length of the unit name including the suffix must not exceed 256 characters. The type suffix must be one of ".service", ".socket", ".device", ".mount", ".automount", ".swap", ".target", ".path", ".timer", ".slice", or ".scope".

in addition we allow `@` to cover the case of template or template instance.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -58,7 +58,7 @@
 * [`Systemd::MachineInfoSettings`](#systemdmachineinfosettings): Matches Systemd machine-info (hostnamectl) file Struct
 * [`Systemd::OomdSettings`](#systemdoomdsettings): Configurations for oomd.conf
 * [`Systemd::ServiceLimits`](#systemdservicelimits): Matches Systemd Service Limit Struct
-* [`Systemd::Unit`](#systemdunit): custom datatype that validates different filenames for systemd units
+* [`Systemd::Unit`](#systemdunit): custom datatype that validates different filenames for systemd units and unit templates
 
 ## Classes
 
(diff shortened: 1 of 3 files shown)


types/unit.pp
```diff
@@ -1,2 +1,3 @@
-# @summary custom datatype that validates different filenames for systemd units
-type Systemd::Unit = Pattern['^[^/]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope)$']
+# @summary custom datatype that validates different filenames for systemd units and unit templates
+# @see https://www.freedesktop.org/software/systemd/man/systemd.unit.html
+type Systemd::Unit = Pattern[/^[[a-z][A-Z][0-9]:\-_.\\@]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope)$/]
```
ekohl, line 3: This also works
```suggestion
type Systemd::Unit = Pattern[/^[a-zA-Z0-9:\-_.\\@]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope)$/]
```

ekohl (commented):
Small suggestion to simplify, otherwise :+1:

REFERENCE.md
```diff
@@ -1803,11 +1803,14 @@ Struct[{
...
+custom datatype that validates different filenames for systemd units and unit templates
+
+* **See also**
+  * https://www.freedesktop.org/software/systemd/man/systemd.unit.html
 
 Alias of
 
 ```puppet
-Pattern['^[^/]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope)$']
+Pattern[/^[[a-z][A-Z][0-9]:\-_.\\@]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope)$/]
```
ekohl, line 1814:
```suggestion
Pattern[/^[a-zA-Z0-9:\-_.\\@]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope)$/]
```

PR #292, opened 2022-08-10
add udevadm & udev facts

jhoblitt:
I want to be able to identify / construct the `/dev/disk/by-id/<foo>` paths for nvme devices on Linux to be used in EL kickstart scripts and have been unable to identify an existing fact which makes this data available.

The approach taken with the `udev` fact is to dump the entire udev dev db as it seems likely that there will be other uses for this data.

Example CLI usage of the `udev fact:

```bash
$ FACTERLIB=$(pwd)/lib/facter be facter udev.name.nvme0n1
{
  devnum => "b 259:0",
  devtype => "disk",
  diskseq => 3,
  name => "nvme0n1",
  path => "/devices/pci0000:00/0000:00:01.0/0000:01:00.0/nvme/nvme0/nvme0n1",
  priority => 0,
  property => {
    CURRENT_TAGS => ":systemd:",
    DEVLINKS => [
      "/dev/disk/by-diskseq/3",
      "/dev/disk/by-path/pci-0000:01:00.0-nvme-1",
      "/dev/disk/by-id/nvme-eui.0025385271b166c9",
      "/dev/disk/by-id/nvme-Samsung_SSD_960_EVO_1TB_S3ETNX0J208970A"
    ],
    DEVNAME => "/dev/nvme0n1",
    DEVPATH => "/devices/pci0000:00/0000:00:01.0/0000:01:00.0/nvme/nvme0/nvme0n1",
    DEVTYPE => "disk",
    DISKSEQ => "3",
    ID_MODEL => "Samsung SSD 960 EVO 1TB",
    ID_PART_TABLE_TYPE => "gpt",
    ID_PART_TABLE_UUID => "369c0ca0-758a-44d5-abe3-0d81c7a73b51",
    ID_PATH => "pci-0000:01:00.0-nvme-1",
    ID_PATH_TAG => "pci-0000_01_00_0-nvme-1",
    ID_REVISION => "2B7QCXE7",
    ID_SERIAL => "Samsung_SSD_960_EVO_1TB_S3ETNX0J208970A",
    ID_SERIAL_SHORT => "S3ETNX0J208970A",
    ID_WWN => "eui.0025385271b166c9",
    MAJOR => "259",
    MINOR => "0",
    SUBSYSTEM => "block",
    TAGS => ":systemd:",
    USEC_INITIALIZED => "17952026"
  },
  subsystem => "block",
  symlink => [
    "disk/by-diskseq/3",
    "disk/by-path/pci-0000:01:00.0-nvme-1",
    "disk/by-id/nvme-eui.0025385271b166c9",
    "disk/by-id/nvme-Samsung_SSD_960_EVO_1TB_S3ETNX0J208970A"
  ],
  sysname => "nvme0n1",
  sysnum => "1"
}
$ FACTERLIB=$(pwd)/lib/facter be facter udev.name.nvme0n1.symlink
[
  "disk/by-diskseq/3",
  "disk/by-id/nvme-Samsung_SSD_960_EVO_1TB_S3ETNX0J208970A",
  "disk/by-id/nvme-eui.0025385271b166c9",
  "disk/by-path/pci-0000:01:00.0-nvme-1"
]
$ FACTERLIB=$(pwd)/lib/facter be facter udev.name.nvme0n1.property.ID_SERIAL
Samsung_SSD_960_EVO_1TB_S3ETNX0J208970A
$ FACTERLIB=$(pwd)/lib/facter be facter udev.path./devices/virtual/vc/vcsu
{
  name => "vcsu",
  path => "/devices/virtual/vc/vcsu",
  priority => 0,
  property => {
    DEVNAME => "/dev/vcsu",
    DEVPATH => "/devices/virtual/vc/vcsu",
    MAJOR => "7",
    MINOR => "64",
    SUBSYSTEM => "vc"
  }
}
```


lib/facter/udevadm.rb
```diff
@@ -0,0 +1,26 @@
...
+#   Check if the kernel is linux.
+#
+# Caveats:
+#
+Facter.add(:udevadm) do
+  confine kernel: :linux
+
+  path = Facter::Util::Resolution.which('udevadm')
+
+  if path
```
ekohl, line 18: I think in general it's bad to execute code outside of `setcode`. It doesn't respect the `confine` bits and gets executed on every OS.

spec/unit/facter/udev_spec.rb
```diff
@@ -0,0 +1,163 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'udev fact', type: :fact do
+  subject(:fact) { Facter.fact(:udev) }
+
+  before { Facter.clear }
```
ekohl, line 8: Some day we'll have a specific fact tester in rspec-puppet that does all of this.

spec/unit/facter/udev_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+      allow(Facter::Util::Resolution).to receive(:exec).with('udevadm info -e').and_return(
+        File.read(fixtures('udevdb', udevdb_fixture))
+      )
+    end
+
+    context 'centos8-supermicro-1114S-WN10RT' do
+      let(:udevdb_fixture) { 'centos8-supermicro-1114S-WN10RT.txt' }
+
+      it do
+        expect(Facter.value(:udev)).to include(
```
ekohl, line 23: Is this a better pattern?
```suggestion
        expect(subject.value).to include(
```
Or since it's named
```suggestion
        expect(fact.value).to include(
```

lib/facter/udev.rb
```diff
@@ -0,0 +1,65 @@
...
+      case type
+      when 'N'
+        d[:name] = data
+      when 'P'
+        d[:path] = data
+      when 'S'
+        d[:symlink].nil? ? d[:symlink] = [data] : d[:symlink].push(data)
+      when 'E'
+        (key, value) = data.split('=', 2)
+        d[:property].nil? ? d[:property] = { key => value } : d[:property][key] = value
```
ekohl, line 34: This may be more accurate:
```suggestion
        d.key?(:property) ? d[:property] = { key => value } : d[:property][key] = value
```

Another alternative is to initialize `d` as:
```ruby
d = {
  symlink: [],
  property: {},
}
```

Then this and the symlink code above is greatly simplified.

ekohl (commented):
I wonder if it makes sense to enhance the existing `disks` fact. Also, as I mentioned in chat, this may be a good thing to have in Facter natively.

lib/facter/udevadm.rb
jhoblitt: :facepalm:

lib/facter/udevadm.rb
```diff
@@ -0,0 +1,26 @@
...
+# Purpose:
+#   Provide information about the udevadm utilty.
+#
+# Resolution:
+#   Check if the kernel is linux.
+#
+# Caveats:
+#
+Facter.add(:udevadm) do
+  confine kernel: :linux
```
bastelfreak, line 14: I think a bit cleaner is it to confine onto the existence of `udevadm`:
```
confine { Facter::Core::Execution.which('udevadm') }
```

then we don't need to wrap the setcode into an if block

jhoblitt: OK - I think this is a reasonable approach.  A possible downside is that it doesn't appear that there is any memoization on top of `Facter::Core::Execution::Posix.which` but odds are that inode data for the `stat` calls will be cached by the OS.

spec/unit/facter/udev_spec.rb
jhoblitt: Yes, that is definitely an improvement.

lib/facter/udevadm.rb
ekohl: You can just move everything inside `setcode` and make it return `nil` if `path` is nil. So something like this probably works:
```ruby
setcode do
  path = Facter::Util::Resolution.which('udevadm')
  if path
    {}
  end
end

File: lib/facter/udev.rb (hunk 4 above)
Comment (jhoblitt) [reply lib/facter/udev.rb:34]: I agree that `#key?` is more correct. I'm unsure if initializing the hash keys is the right approach since not all devices will have a symlink.

File: lib/facter/udevadm.rb (hunk 1 above)
Comment (jhoblitt) [reply lib/facter/udevadm.rb:18]: This should be addressed and I've converted to using `Facter::Core::Execution` as well.

File: spec/unit/facter/udev_spec.rb (hunk 3 above)
Comment (jhoblitt) [reply spec/unit/facter/udev_spec.rb:23]: Done.

File: lib/facter/udevadm.rb (hunk 5 above)
Comment (jhoblitt) [reply lib/facter/udevadm.rb:14]: Done.

Comment (jhoblitt) [reply lib/facter/udevadm.rb:14]: Overlapping comments, @ekohl last suggestion hadn't appear when I marked this resolved.  I like that approach even better.  Standby...

Comment (jhoblitt) [reply lib/facter/udevadm.rb:14]: [re]done.

File: lib/facter/udev.rb (hunk 4 above)
Comment (ekohl) [reply lib/facter/udev.rb]: You could argue that an empty list also signifies that there are no symlinks.

File: lib/facter/udevadm.rb [hunk 6]
```diff
@@ -0,0 +1,25 @@
...
+
+  setcode do
+    path = Facter::Core::Execution.which('udevadm')
+    return nil if path.nil?
```
ekohl, line 18: I don't think you can use `return` in blocks. At least, pretty sure I had problems with that in the past.

spec/unit/facter/udevadm_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+require 'spec_helper'
+
+describe 'udevadm fact', type: :fact do
+  subject(:fact) { Facter.fact(:udevadm) }
+
+  before { Facter.clear }
+
+  context 'when kernel is linux and udevadm is present' do
+    before do
+      allow(Facter.fact(:kernel)).to receive(:value).and_return(:linux)
```
ekohl, line 12: I think the real fact is an uppercased string
```suggestion
      allow(Facter.fact(:kernel)).to receive(:value).and_return('Linux')
```

spec/unit/facter/udevadm_spec.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'udevadm fact', type: :fact do
```
ekohl, line 5: Structure wise I I think what you can do is:
```ruby
describe 'udevadm fact', type: :fact do
  context 'on Linux' do
    before { allow(Facter.fact(:kernel)).to receive(:value).and_return('Linux') }

    context 'when udevadm is present' do
      # ...
    end

    context 'when udevadm is absent' do
      # ...
    end
  end

  context 'on Windows' do
    # ...
  end
end

File: lib/facter/udevadm.rb (hunk 6 above)
Comment (jhoblitt) [reply lib/facter/udevadm.rb:18]: Yes, that was a silly mistake. I will change it to break.

File: spec/unit/facter/udevadm_spec.rb (hunk 7 above)
Comment (jhoblitt) [reply spec/unit/facter/udevadm_spec.rb:12]: I copied this from the `systemd` fact and it oddly does seem to be working. I will fix that fact as well.

File: spec/unit/facter/udevadm_spec.rb (hunk 8 above)
Comment (jhoblitt) [reply spec/unit/facter/udevadm_spec.rb:5]: I honestly didn't do that on purpose as I'm not a fan on tests been deeply nested/indented but I will DRY it up.

File: lib/facter/udevadm.rb (hunk 6 above)
Comment (jhoblitt) [reply lib/facter/udevadm.rb]: `return` has been replaced with `nil`.

File: spec/unit/facter/udevadm_spec.rb (hunk 7 above)
Comment (jhoblitt) [reply spec/unit/facter/udevadm_spec.rb]: Fixed for all facts in this mod.

File: spec/unit/facter/udevadm_spec.rb (hunk 8 above)
Comment (jhoblitt) [reply spec/unit/facter/udevadm_spec.rb:5]: The spec has been DRY'd a bit.

**Comment by jcpunk** [comment]
> Should the /dev devices be moved under a sub hash key such as udev::dev::<device name> to allow for all devices to be exposed as udev::path::<device> in the future?

I'd prefer to have the `udev::dev` and `udev::path` options available going forward.

I'd love to have these natively in facter, but it can take a long time for those to get out to folks whereas this would be a lot faster....

**Comment by jhoblitt** [comment]
I'm concerned that adding all of the path entries is going to slow the fact down but I'm willing give it.

**Comment by jhoblitt** [comment]
I've added a test fixture for EL9.

**Review by bastelfreak — APPROVED** [review]

**Comment by jhoblitt** [comment]
It looks like fedora 37 uses more letter codes than EL7/8/9. I'm looking into adding support for the extra codes.

**Comment by jhoblitt** [comment]
@jcpunk @bastelfreak I have added a seperate `name` and `path` index as `udev::{name,path}`.  Is this what you had in mind?

Interestingly, the facter cli doesn't seem to have a mechanism for escaping `.` in fact names. E.g.:

```bash
$ FACTERLIB=$(pwd)/lib/facter be facter udev.name.nvme0n1
{
  name => "nvme0n1",
  path => "/devices/pci0000:00/0000:00:01.0/0000:01:00.0/nvme/nvme0/nvme0n1",
  priority => 0,
  property => {
    CURRENT_TAGS => ":systemd:",
    DEVLINKS => [
      "/dev/disk/by-diskseq/3",
      "/dev/disk/by-path/pci-0000:01:00.0-nvme-1",
      "/dev/disk/by-id/nvme-Samsung_SSD_960_EVO_1TB_S3ETNX0J208970A",
      "/dev/disk/by-id/nvme-eui.0025385271b166c9"
    ],
    DEVNAME => "/dev/nvme0n1",
    DEVPATH => "/devices/pci0000:00/0000:00:01.0/0000:01:00.0/nvme/nvme0/nvme0n1",
    DEVTYPE => "disk",
    DISKSEQ => "3",
    ID_MODEL => "Samsung SSD 960 EVO 1TB",
    ID_PART_TABLE_TYPE => "gpt",
    ID_PART_TABLE_UUID => "369c0ca0-758a-44d5-abe3-0d81c7a73b51",
    ID_PATH => "pci-0000:01:00.0-nvme-1",
    ID_PATH_TAG => "pci-0000_01_00_0-nvme-1",
    ID_REVISION => "2B7QCXE7",
    ID_SERIAL => "Samsung_SSD_960_EVO_1TB_S3ETNX0J208970A",
    ID_SERIAL_SHORT => "S3ETNX0J208970A",
    ID_WWN => "eui.0025385271b166c9",
    MAJOR => "259",
    MINOR => "0",
    SUBSYSTEM => "block",
    TAGS => ":systemd:",
    USEC_INITIALIZED => "17952026"
  },
  symlink => [
    "disk/by-diskseq/3",
    "disk/by-path/pci-0000:01:00.0-nvme-1",
    "disk/by-id/nvme-Samsung_SSD_960_EVO_1TB_S3ETNX0J208970A",
    "disk/by-id/nvme-eui.0025385271b166c9"
  ]
}
$ FACTERLIB=$(pwd)/lib/facter be facter udev.path./devices/virtual/vc/vcsu
{
  name => "vcsu",
  path => "/devices/virtual/vc/vcsu",
  priority => 0,
  property => {
    DEVNAME => "/dev/vcsu",
    DEVPATH => "/devices/virtual/vc/vcsu",
    MAJOR => "7",
    MINOR => "64",
    SUBSYSTEM => "vc"
  }
}
$ FACTERLIB=$(pwd)/lib/facter be facter udev.path./devices/pci0000:00/0000:00:01.0/0000:01:00.0/nvme/nvme0/nvme0n1

$ FACTERLIB=$(pwd)/lib/facter be facter udev.path./devices/pci0000:00/0000:00:01\\.0/0000:01:00\\.0/nvme/nvme0/nvme0n1


```

jhoblitt:
I have added support for the all of the additional letter codes present on my f37 desktop.  There are still a few unsupported code but I'd like to have a real world test fixture for them before adding support.

lib/facter/udevadm.rb
```diff
@@ -0,0 +1,25 @@
...
+# Resolution:
+#   Check if the kernel is linux.
+#
+# Caveats:
+#
+Facter.add(:udevadm) do
+  confine(kernel: 'Linux')
+
+  setcode do
+    path = Facter::Core::Execution.which('udevadm')
```
bastelfreak, line 17: I prefer to confine on the file, but thats personal preference.

bastelfreak (approved):

lib/facter/udevadm.rb
jhoblitt: Uh, is the path `/usr/bin/udevadm` guaranteed to exist?

deemon87:
I have question. What the reason to get udevadm facts? For example on some hosts count of facts increase on 25k+

PR #295, opened 2022-09-21
Unit_file: Add notify_service

fe80:
Hello,

I had add new parameters for Unit_file define, `notify_service`. This params I can use for disable the service notify. Very useful for not restart an in memory service for space or a new comment.

I've try to add spec but that not work. I've try something like that:
```ruby
context 'notify_service => false' do
  let(:params) { { enable: true, notify_service: false } }
  
  it do
    expect(subject).to contain_service(title).
      without_subscribes_to("File[/etc/systemd/system/#{title}]")
  end
end
```
or

```ruby
context 'notify_service => false' do
  let(:params) { { enable: true, notify_service: false } }
  
  it do
    expect(subject).to contain_file("/etc/systemd/system/#{title}").
      without_notifies("Service[#{title}]")
  end
end
```

Can you help me for this part ?

Regards,

### DIFF
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -64,6 +64,9 @@
 # @param daemon_reload
 #   call `systemd::daemon-reload` to ensure that the modified unit file is loaded
 #
+# @param notify_service
+#   Notify service if systemd file is changed
+#
 # @example manage unit file + service
 #   systemd::unit_file { 'foo.service':
 #     content => file("${module_name}/foo.service"),
@@ -88,7 +91,8 @@ define systemd::unit_file (
   Optional[Boolean]                        $hasstatus = undef,
   Boolean                                  $selinux_ignore_defaults = false,
   Hash[String[1], Any]                     $service_parameters = {},
-  Boolean                                  $daemon_reload = true
+  Boolean                                  $daemon_reload = true,
+  Boolean                                  $notify_service = true,
 ) {
   include systemd
 
@@ -152,7 +156,11 @@ define systemd::unit_file (
       }
       Service[$name] -> File["${path}/${name}"]
     } else {
-      File["${path}/${name}"] ~> Service[$name]
+      if $notify_service {
+        File["${path}/${name}"] ~> Service[$name]
+      } else {
+        File["${path}/${name}"] -> Service[$name]
+      }
 
       if $daemon_reload {
         Systemd::Daemon_reload[$name] ~> Service[$name]
--- a/spec/defines/unit_file_spec.rb
+++ b/spec/defines/unit_file_spec.rb
@@ -267,6 +267,16 @@ describe 'systemd::unit_file' do
             expect(subject).not_to create_systemd__daemon_reload(title)
           }
         end
+
+        context 'notify_service => false' do
+          let(:params) { { enable: true, notify_service: false } }
+
+          it do
+            is_expected.not_to contain_service(title).with_subscribes_to(
+              "File[/etc/systemd/system/#{title}]"
+            )
+          end
+        end
       end
     end
   end


ekohl:
I think there's no such thing as `without_notifies` or `without_subscribes_to`. The closest I can think of is
```ruby
it { is_expected.not_to contain_service(title).with_subscribes_to("File[/etc/systemd/system/#{title}]")
```

Slight nuance is that it's also true if the service is not contained in the catalog. Perhaps a possible enhancement for https://github.com/puppetlabs/rspec-puppet?

fe80:
Hello @ekohl 

Thank it's look good !

I've also create an issue here https://github.com/puppetlabs/rspec-puppet/issues/28

Regards,

smortex (approved):

spec/defines/unit_file_spec.rb
```diff
@@ -267,6 +267,16 @@
             expect(subject).not_to create_systemd__daemon_reload(title)
           }
         end
+
+        context 'notify_service => false' do
+          let(:params) { { enable: true, notify_service: false } }
+
+          it do
+            is_expected.not_to contain_service(title).with_subscribes_to(
```
ekohl, line 275: `with_subscribes_to` expects a property to be set, like
```puppet
service { $title:
  subscribes_to => File[$file],
}
```

That's never going to be true. This checks the compiled catalog (which mirrors the check on line 210 for example):
```suggestion
            is_expected.not_to contain_service(title).that_subscribes_to(
```

ekohl (changes requested):

spec/defines/unit_file_spec.rb
fe80: That not work, but something like this look ok:
```diff
           let(:params) { { enable: true, notify_service: false } }
 
           it do
-            is_expected.not_to contain_service(title).with_subscribes_to(
-              "File[/etc/systemd/system/#{title}]"
+            is_expected.not_to contain_file("/etc/systemd/system/#{title}").with(
+              notify: /Service\[#{title}\]/
             )
           end
         end
```

It's good for you ?

manifests/unit_file.pp
```diff
@@ -152,7 +156,9 @@
       }
       Service[$name] -> File["${path}/${name}"]
     } else {
-      File["${path}/${name}"] ~> Service[$name]
+      if $notify_service {
+        File["${path}/${name}"] ~> Service[$name]
+      }
```
ekohl, line 161: You still want correct ordering:
```suggestion
      } else {
        File["${path}/${name}"] -> Service[$name]
      }
```

spec/defines/unit_file_spec.rb
fe80: I've try before, with my path when it's true that fail correctly:
```diff
diff --git a/spec/defines/unit_file_spec.rb b/spec/defines/unit_file_spec.rb
index a0521aa..6dde40a 100644
--- a/spec/defines/unit_file_spec.rb
+++ b/spec/defines/unit_file_spec.rb
@@ -272,8 +272,18 @@ describe 'systemd::unit_file' do
           let(:params) { { enable: true, notify_service: false } }
 
           it do
-            is_expected.not_to contain_service(title).with_subscribes_to(
-              "File[/etc/systemd/system/#{title}]"
+            is_expected.not_to contain_file("/etc/systemd/system/#{title}").with(
+              notify: /Service\[#{title}\]/
+            )
+          end
+        end
+
+        context 'notify_service => true' do
+          let(:params) { { enable: true, notify_service: true } }
+
+          it do
+            is_expected.not_to contain_file("/etc/systemd/system/#{title}").with(
+              notify: /Service\[#{title}\]/
             )
           end
         end
```

```
...
  21) systemd::unit_file supported operating systems on centos-8-x86_64 notify_service => true is expected not to contain File[/etc/systemd/system/test.service] with notify =~ /Service\[test.service\]/
      Failure/Error:
        is_expected.not_to contain_file("/etc/systemd/system/#{title}").with(
          notify: /Service\[#{title}\]/
        )
      
        expected that the catalogue would not contain File[/etc/systemd/system/test.service]
      # ./spec/defines/unit_file_spec.rb:285:in `block (6 levels) in <top (required)>'

Finished in 1 minute 49.83 seconds (files took 54.25 seconds to load)
609 examples, 21 failures
...
```

fe80: Hello,

Any update/idea about this test ?

TheMeier:
implemented in https://github.com/voxpupuli/puppet-systemd/pull/433

PR #303, opened 2022-12-20
modulesync 5.4.0

bastelfreak:
modulesync 5.4.0


ekohl (approved):

ekohl:
Debian 8 fails because it's EOL and has moved away from the mirrors.

PR #306, opened 2023-01-26
drop debian 8 support (EOL)

jhoblitt:


ekohl:
https://github.com/voxpupuli/puppet-systemd/pull/305 is more complete.

PR #316, opened 2023-01-30
Revert udevadm and udev facts from #292

### LINKED ISSUE #314: Udev fact exceeds the number of facts limit
The new [udev fact](https://github.com/voxpupuli/puppet-systemd/blob/master/lib/facter/udev.rb) introduced in v4.0.0 of the module seems to create a high number of nested facts, which on my nodes exceeds the default limit of 2048 by a lot. Is this actually intended behaviour?
```
Info: Loading facts
Warning: The current total number of facts: 11210 exceeds the number of facts limit: 2048
```

jhoblitt:
Per discussion on the #voxpupuli slack channel and #314, it is clear that the `udev` fact is causing problems beyond merely triggering the default puppet agent max fact warning.

The short term plan is to migrate these facts into a dedicated module: https://github.com/voxpupuli/puppet-udevdb

Resolves #314


jhoblitt:
The fixed will be included in the #317 release.

PR #322, opened 2023-03-16
Add timer_entry to manage_{dropin,unit}

ekohl:
This makes it easier to deal with timer units. Right now it only implements OnCalendar, but it should be easy to expand to all types.


bastelfreak (approved):

jhoblitt (approved):
LGTM.

PR #329, opened 2023-04-14
Add AmbientCapabilities to Systemd::Unit::Service

ekohl:


types/unit/service.pp
```diff
@@ -4,6 +4,7 @@
 #
 type Systemd::Unit::Service = Struct[
   {
+    Optional['AmbientCapabilities']       => Variant[Pattern[/^CAP_[A-Z_]+$/],Array[Pattern[/^CAP_[A-Z_]+$/],1]],
```
ekohl, line 7: This doesn't support multiple capabilities separated by whitespace. I thought about that but it became complicated in the template. I don't need it now and if needed you can specify it multiple times. Quoting `man systemd.exec`:
> This option may appear more than once in which case the ambient capability sets are merged.

bastelfreak (approved):

PR #331, opened 2023-04-18
Group spec expectations in a single example

ekohl:
The catalog is compiled for every example, so testing the same resource in a single example is faster.


traylenator (approved):
And I'll rebase the mine after

traylenator:
Must get into habbit of doing this.

kenyon (approved):

PR #333, opened 2023-04-19
Add comment in manage_unit deployed files

traylenator:

It's useful to see clearly see that files deployed with puppet are deployed with puppet.

Add a comment to top of files deployed by `systemd::manage_unit` and  `systemd::manage_dropin`.

e.g 
```puppet
systemd::manage_unit{'redis@.service':
  ensure          => present,
  unit_entry    => {
    'Description' => 'redis server instance on port %i',
   },
  service_entry => {
    'Type'      => 'simple',
    'User'      => 'redis',
    'Group'     => 'redis',
    'ExecStart' => '/usr/bin/redis-server /etc/redis/redis_%i.conf',
    'ExecStop'  => '/usr/bin/redis-cli -p %i shutdown',
    'PIDFile'   => '/run/redis/%i.pid',
  },
}
```
will now deploy `/etc/systemd/system/@redis@.service` with a first line of:

```
# Deployed with puppet
```

### DIFF
--- a/templates/unit_file.epp
+++ b/templates/unit_file.epp
@@ -19,6 +19,8 @@
 ]
 
 -%>
+# Deployed with puppet
+#
 <%-
 $_unit_sections.each | $_section | {
   $_values = getvar("${downcase($_section)}_entry")
--- a/spec/defines/manage_dropin_spec.rb
+++ b/spec/defines/manage_dropin_spec.rb
@@ -17,6 +17,9 @@ describe 'systemd::manage_dropin' do
             }
           end
 
+          it { is_expected.to compile.with_all_deps }
+          it { is_expected.to contain_systemd__dropin_file('foobar.conf').with_content(%r{^# Deployed with puppet$}) }
+
           context 'setting some parameters simply' do
             let(:params) do
               super().merge(


templates/unit_file.epp
```diff
@@ -5,6 +5,7 @@
  Optional[Hash] $timer_entry,
 | -%>
 <% if $unit_entry { -%>
+# Deployed with puppet
```
bastelfreak, line 8: I'm wondering if this comment should be above line 7? but with all the braces the template is hard to read :D

traylenator: Should be it and it clearly needed a test.

traylenator:
We could put the comment as an additional description:

```config
[Unit]
Docs=This unit is at least partially manged by puppet
```
And it would show up in a status, something like:

```
$ systemctl status sshd
○ sshd.service - OpenSSH server daemon
     Loaded: loaded (/usr/lib/systemd/system/sshd.service; disabled; preset: disabled)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: inactive (dead)
       Docs: man:sshd(8)
             man:sshd_config(5)
             This unit is at least partially manged by puppet
            
```

cool but probably a bit intrusive.

traylenator:
If possible and then I'll make a release.

bastelfreak:
I would be fine with a comment or a description, whatever you prefer.

traylenator:
Morning,
I think stick with the comment as is. If we add to the description it gets ambiguous when the description is reset with an `''` value.

smortex (approved):

PR #341, opened 2023-04-27
Allow LimitCORE in [Service] for manage_unit/dropin

traylenator:

Addition of LimitCOR to `[service]` section of
`systemd::manage_unit` and `systemd::manage_dropin`

Added as penance for giving a wrong guess twice to someone how to do this. Always confuses me since while:

`systemctl show sshd.service -t LimitCORESoft` has a value it cannot be set as I understand it.

There is some duplication here with the existing `systemd::service_limits` but that is fine.

* https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Process%20Properties

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -794,6 +794,18 @@ systemd::manage_dropin { 'triggerlimit.conf':
 }
 ```
 
+##### drop in file to override the LimitCORE for a service
+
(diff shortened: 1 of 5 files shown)


traylenator:
Test failure unrelated - tomorrow.

PR #345, opened 2023-05-03
No insistence on unit_entry ever or service_entry with absent manage_unit

traylenator:

Removing a unit file with `systemd::manage_unit` with the following:

```
systemd::manage_unit {'foo.service':
  ensure     => 'absent',
}
```

This is not possible and a `service_entry` and `unit_entry` must be specified as the unit name
is in the service namespace.

Remove insistence for `unit_entry` and  `service_name` parameter for the absent case.

For the `ensure => 'present`` the `unit_entry` is no longer required since it is perfectly possible to create
units without any `[Unit]` section. 

Also improve failure message when `service_entry` is not specified for service unit.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1037,6 +1037,13 @@ systemd::manage_unit{'arcd@.service':
     'ExecStart'     => /usr/sbin/arcd /usr/libexec/arcd/arcd.pl,
     'StandardInput' => 'socket',
   }
+```
+
+##### Remove a unit file
+
+```puppet
+systemd::manage_unit { 'my.service':
+  ensure     => 'absent',
 }
 ```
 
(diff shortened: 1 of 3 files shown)


Geod24:
Should we require a `unit_entry` as well ? All fields in `types` are optional, and you can have a systemd unit file without a `[Unit]` entry (e.g. for timers).

traylenator:
> systemd::manage_unit {'foo.service':
>   ensure     => 'absent',
> }

So both:

```puppet
systemd::manage_unit {'foo.service':
  ensure     => 'absent',
}
```
and

```puppet
systemd::manage_unit {'foo.service':
  ensure     => 'present',
  service_entry => {
     ....
  },
}
```
are now both valid.

smortex (approved):
LGTM

Issue #299, opened 2022-10-12
Systemd::Unit::Service: missing User and Group


b4ldr:
a per the description we need to support User and Group in Systemd::Unit::Service

thiagolinhares:
Hi @b4ldr do you plan on merging that piece of code you're working soon?
I ask it because I came across this dependency with a project I'm working on.

Also, if you'd need help evaluating and testing, I would love to help.

b4ldr:
@thiagolinhares in on honestly the patch i was working grew out of control i thikn i need to scrap that and confine myself to just this secific bug, however im not sure when ill have the time to do that

PR #352, opened 2023-06-15
Allow WorkingDirectory to be specified in [Service]

traylenator:

Support use of `WorkingDirectory` directive in `[Service]` section of `systemd::manage_unit` and `systemd::manage_dropin`

https://www.freedesktop.org/software/systemd/man/systemd.exec.html#WorkingDirectory=

Fixes: #299

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -2316,6 +2316,7 @@ Struct[{
     Optional['AmbientCapabilities']       => Variant[Pattern[/^CAP_[A-Z_]+$/],Array[Pattern[/^CAP_[A-Z_]+$/],1]],
     Optional['User']                      => String[1],
     Optional['Group']                     => String[1],
+    Optional['WorkingDirectory']          => String[0],
     Optional['Type']                      => Enum['simple', 'exec', 'forking', 'oneshot', 'dbus', 'notify', 'idle'],
     Optional['ExitType']                  => Enum['main', 'cgroup'],
     Optional['RemainAfterExit']           => Boolean,
--- a/spec/type_aliases/systemd_unit_service_spec.rb
+++ b/spec/type_aliases/systemd_unit_service_spec.rb
@@ -44,4 +44,9 @@ describe 'Systemd::Unit::Service' do
   it { is_expected.to allow_value({ 'Group' => 'root' }) }
   it { is_expected.to allow_value({ 'StandardOutput' => 'null' }) }
   it { is_expected.to allow_value({ 'StandardError' => 'null' }) }
+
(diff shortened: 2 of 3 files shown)


REFERENCE.md
```diff
@@ -2316,6 +2316,7 @@ Struct[{
     Optional['AmbientCapabilities']       => Variant[Pattern[/^CAP_[A-Z_]+$/],Array[Pattern[/^CAP_[A-Z_]+$/],1]],
     Optional['User']                      => String[1],
     Optional['Group']                     => String[1],
+    Optional['WorkingDirectory']          => String,
```
smortex, line 2319: Without an explicit minimum length, I trend to think that it was forgotten.  So I suggest:

```suggestion
    Optional['WorkingDirectory']          => String[0],
```

spec/type_aliases/systemd_unit_service_spec.rb
```diff
@@ -39,4 +39,10 @@
   it { is_expected.to allow_value({ 'EnvironmentFile' => ['/etc/sysconfig/foo', '-/etc/sysconfig/foo-bar'] }) }
   it { is_expected.not_to allow_value({ 'EnvironmentFile' => '-/' }) }
   it { is_expected.not_to allow_value({ 'EnvironmentFile' => 'relative-path.sh' }) }
+
+  it { is_expected.to allow_value({ 'WorkingDirectory' => '/var/lib/here' }) }
+  it { is_expected.to allow_value({ 'WorkingDirectory' => '-/var/lib/here' }) }
+  it { is_expected.to allow_value({ 'WorkingDirectory' => '~' }) }
+  it { is_expected.to allow_value({ 'WorkingDirectory' => '' }) }
+
```
smortex, line 47: Removing this blank line should fix CI.

smortex (approved):
Looks good, fixing CI should be straightforward, and while here maybe we can put an explicit 0-sized string as minimum length.

REFERENCE.md
traylenator: You need to allow `WorkingDirectory=''` to allow people to set back to the default.
Folk could explicitly set back to the default I guess but other similar directives support `''`.

traylenator: Sorry misinterpreted - Explicit 0 ?  that's the default anyway surely?

smortex:
> ```
> Failure/Error: configure_beaker
> RuntimeError:
>   No repository installation step for archlinux yet...
> ```
> 
> So that's unrelated, but something we should fix.

I opened this today: https://github.com/voxpupuli/beaker_puppet_helpers/pull/31 :wink:

PR #357, opened 2023-06-27
Support Debian 12

traylenator:

In particular with Debian 11->12 systemd-resolved is now a sub package
that must be installed.

Clearly there are no tests running for Debian 12 yet.

### DIFF
--- a/spec/acceptance/resolved_spec.rb
+++ b/spec/acceptance/resolved_spec.rb
@@ -3,6 +3,8 @@
 require 'spec_helper_acceptance'
 
 describe 'systemd with manage_resolved true' do
+  has_package = (fact('os.family') == 'RedHat' && fact('os.release.major') != '8') || (fact('os.name') == 'Debian' && fact('os.release.major').to_i >= 12)
+
   context 'configure systemd resolved' do
     it 'works idempotently with no errors' do
       pp = <<-PUPPET
@@ -13,10 +15,7 @@ describe 'systemd with manage_resolved true' do
       PUPPET
       apply_manifest(pp, catch_failures: true)
       # RedHat 7, 9, Debian 12 and newer installs package first run before fact $facts['internal_services'] is set
-      if (fact('os.release.major') != '8' && (fact('os.family') == 'RedHat')) ||
-         ((fact('os.name') == 'Debian' && fact('os.release.major').to_i >= 12))
-        apply_manifest(pp, catch_failures: true)
-      end
+      apply_manifest(pp, catch_failures: true) if has_package
       apply_manifest(pp, catch_changes: true)
     end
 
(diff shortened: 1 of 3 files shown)


traylenator:
Needs a bit more testing as enabling with:

```puppet
class{'systemd':
  manage_resolved   => true,
  use_stub_resolver => true,
}
```
I lost DNS. ( fixed by reboot and now running and  using systemd-resolved )

traylenator:
> Needs a bit more testing as enabling with:
> 
> ```puppet
> class{'systemd':
>   manage_resolved   => true,
>   use_stub_resolver => true,
> }
> ```
> 
> I lost DNS. ( fixed by reboot and now running and using systemd-resolved )

Understood - not relevent.

rh-org:
LGTM

saz:
LGTM.

We should think about, installing `systemd-resolved` and `systemd-timesyncd` by default and just disable it on Debian 10(?) and 11, if required. Currently, any newer Debian release might "break", as the packages are missing, no configuration is happening, but at the same time, there won't be any error message.

anarcat (approved):
LGTM.

tuxmea (approved):

traylenator:
@tuxmea Did you try migrating with this - I do loose initially which I never really understood?

ekohl:
FYI: https://github.com/voxpupuli/puppet_metadata/pull/102 would add Debian 12 to acceptance tests using distro packages so you at least have Puppet 7 to test with.

ekohl (changes requested):
If we apply this with https://github.com/voxpupuli/modulesync_config/pull/855 we should have at least some acceptance test.

ekohl:
I took the liberty of rebasing this.

spec/acceptance/resolved_spec.rb
```diff
@@ -12,8 +12,11 @@
       }
       PUPPET
       apply_manifest(pp, catch_failures: true)
-      # RedHat 7, 9 and newer installs package first run before fact $facts['internal_services'] is set
-      apply_manifest(pp, catch_failures: true) if fact('os.release.major') != '8' && (fact('os.family') == 'RedHat')
+      # RedHat 7, 9, Debian 12 and newer installs package first run before fact $facts['internal_services'] is set
+      if (fact('os.release.major') != '8' && (fact('os.family') == 'RedHat')) ||
+         ((fact('os.name') == 'Debian' && fact('os.release.major').to_i >= 12))
+        apply_manifest(pp, catch_failures: true)
+      end
```
ekohl, line 19: For systemd-networkd I extracted this to a separate variable:
https://github.com/voxpupuli/puppet-systemd/blob/5f64bc177c6c1a39b557292840a46c7954dc9f05/spec/acceptance/networkd_spec.rb#L6

Then I reused it to check if the package is installed:
https://github.com/voxpupuli/puppet-systemd/blob/5f64bc177c6c1a39b557292840a46c7954dc9f05/spec/acceptance/networkd_spec.rb#L32

ekohl (approved):

bastelfreak (approved):

PR #358, opened 2023-07-03
Allow SupplementaryGroups and DynamicUser

traylenator:

Support `service_entries` for `DynamicUser` and `SupplementaryGroups`

* https://www.freedesktop.org/software/systemd/man/systemd.exec.html#DynamicUser=
* https://www.freedesktop.org/software/systemd/man/systemd.exec.html#SupplementaryGroups=


smortex (approved):
LGTM!

PR #359, opened 2023-07-03
Add Puppet 8 support

bastelfreak:


ekohl:
https://github.com/voxpupuli/puppet_metadata/pull/91 is intended to fix the Archlinux issue by using the none collection recently added. See https://github.com/voxpupuli/puppet-systemd/pull/361 for an example run.

smortex (approved):
This is fine. Do you want we wait for the Arch fix to be released before merging this?

ekohl:
> Do you want we wait for the Arch fix to be released before merging this?

Not really. It's broken everywhere so we can continue without it.

PR #363, opened 2023-07-12
Nice, IOSchedulingPriority and IOSchedulingClass

traylenator:

Allow three extra settings for `[Service]` sections:

* Nice https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Nice=
* IOSchedulingClass https://www.freedesktop.org/software/systemd/man/systemd.exec.html#IOSchedulingClass=
* IOSchedulingPriority https://www.freedesktop.org/software/systemd/man/systemd.exec.html#IOSchedulingPriority=


types/unit/service.pp
```diff
@@ -25,6 +25,9 @@
     Optional['ExecStopPost']              => Variant[Systemd::Unit::Service::Exec,Array[Systemd::Unit::Service::Exec,1]],
     Optional['KillSignal']                => Pattern[/^SIG[A-Z]+$/],
     Optional['KillMode']                  => Enum['control-group', 'mixed', 'process', 'none'],
+    Optional['Nice']                      => Variant[String[0,0],Integer[-20,19]],
```
ekohl, line 28: It took me a second to realize `String[0,0]` equals `Enum['']`. Thoughts on what should be preferred?

ekohl (approved):

types/unit/service.pp
traylenator: * `Enum[''] when part of a  longer `Enum` clearly makes sense.
* `String[0,0]` reads very obviously as the empty string.

Expect `String[0,0]` is quicker though that's nothing.

ekohl: I've gotten used to writing `Enum['']` because I just never thought about `String[0,0]` before. In support of`Enum['']`: it's shorter.

I'm fine either way, just wanted to write down my thoughts here.

PR #367, opened 2023-07-13
add ability to set limits with the systemd::manage_unit resource

oOHenry:
currently it is not allowed to set limits with the `systemd::manage_unit` ressource, to set limits you need to create your service and then create a dropin file with `systemd::service_limits`. 
Add limit parameters to the `Systemd::Unit::Service`type


manifests/manage_unit.pp
```diff
@@ -142,7 +147,7 @@
     daemon_reload           => $daemon_reload,
     content                 => epp('systemd/unit_file.epp', {
         'unit_entry'    => $unit_entry,
-        'service_entry' => $service_entry,
+        'service_entry' => merge($service_entry, $limit_entry),
```
traylenator, line 150: The current `service_entry` , `socket_entry`, ...  parameters correspond to the  `[Service]`, `[Socket]` entries of the unit file.

A `LimitNOFILE` is just a attribute of the existing `[Service]` entry so new parameter for the existing `Systemd::Unit::Service` type ?

I don't believe this should be added like this.

traylenator (changes requested):

manifests/manage_unit.pp
traylenator: Exactly the same as the existing [LimitCORE](https://github.com/voxpupuli/puppet-systemd/blob/master/types/unit/service.pp#L33)

oOHenry: If I add all limits to the `Systemd::Unit::Service` it would be possible to set limits with `systemd::manage_dropin` instead of `systemd::service_limits`.
If that is OK I can copy all limits from [Systemd::ServiceLimits](https://github.com/voxpupuli/puppet-systemd/blob/master/types/servicelimits.pp ) to [Systemd::Unit::Service](https://github.com/voxpupuli/puppet-systemd/blob/master/types/unit/service.pp)

traylenator: Could also change the existing `systemd::servicelimits` to just use `systemd::manage_dropin` and drop that duplication. This would be good.

traylenator: Could consider `systemd::servicelimits` for deprecation to be honest. Not as part of this however.

oOHenry: @traylenator I added all Limits to `Systemd::Unit::Service`, is this PR now good for merge or do I need to change something else?

REFERENCE.md
```diff
@@ -2352,7 +2352,48 @@ Struct[{
     Optional['IOSchedulingPriority']      => Variant[String[0,0],Integer[0,7]],
     Optional['SyslogIdentifier']          => String,
     Optional['LogLevelMax']               => Enum['emerg','alert','crit','err','warning','notice','info','debug'],
+    Optional['LimitCPU']                  => Pattern['^\d+(s|m|h|d|w|M|y)?(:\d+(s|m|h|d|w|M|y)?)?$'],
```
traylenator, line 2355: Just need some type checking in `spec/type_units/systemd_unit_service_spec.rb` copying some from the existing type makes sense.

Once this is merged (and released) am happy to do deprecating merge of `systemd::servicelimits` it can sit there till next major release.

traylenator (changes requested):

REFERENCE.md
oOHenry: :+1:  took tests from `systemd_servicelimits_spec.rb`

traylenator (approved):

PR #371, opened 2023-08-05
Implement DNSStubListenerExtra for resolved.conf

ekohl:
Untested now, but something I need in my setup and having a PR makes testing easier.


ekohl:
OracleLinux 7  is just messed up in acceptance tests. I think we should drop it.

Issue #318, opened 2023-02-06
EL8 and EL9 need the `systemd-networkd` package installed for `systemd::network`


hboetes:
As far as I can tell `/etc/systemd/network` is not created by  `systemd::networkd` forcing me to do something like this.

```puppet
      systemd::network{"${interface[identifier]}.network":
        content         => $content,
        restart_service => true,
        require         => File['/etc/systemd/network'],
      }
```

Wouldn't it be nice for `systemd::networkd` to — optionally or not — create `/etc/systemd/network`

kenyon:
`/etc/systemd/network` is normally created by the `systemd` package. What OS are you on?

hboetes:
I created this recipe for Rocky Linux 8.7. So to see what is going on  I downloaded the rocky rpm from https://download.rockylinux.org/pub/rocky/8/BaseOS/x86_64/os/Packages/s/systemd-239-68.el8_7.2.x86_64.rpm and compared the output. (I doubled checked the output, this is a short summary)

```
han@it1notebook ~ % rpm -ql systemd-251.11-1.fc37.x86_64|grep /etc/systemd/network
/etc/systemd/network
han@it1notebook ~ % rpm -ql systemd-239-68.el8_7.2.x86_64.rpm |/usr/bin/grep -q /etc/systemd/network || echo 'No /etc/systemd/network'
warning: systemd-239-68.el8_7.2.x86_64.rpm: Header V4 RSA/SHA256 Signature, key ID 6d745a60: NOKEY
No /etc/systemd/network
```

So the Fedora Core 37 rpm does indeed contain `/etc/systemd/network`, whereas the Rocky Linux rpm does not.

kenyon:
Interesting. To be fair, this module doesn't yet declare support for Rocky Linux: https://github.com/voxpupuli/puppet-systemd/blob/d785f5d8217d6f34c66524b61d51da18e254345a/metadata.json

jhoblitt:
@kenyon We can fix that... :) #319

kenyon:
Since the #319 tests passed, looks like the problem now is that we are missing test coverage. I don't see any acceptance tests applying `systemd::network` resources: https://github.com/voxpupuli/puppet-systemd/tree/master/spec/acceptance

ekohl:
systemd-networkd is not included on EL8 (at least RHEL 8 and CentOS Stream 8), but it is in EPEL8. Then:
```console
# rpm -qv systemd-networkd
[...]

PR #380, opened 2023-10-18
Install systemd-networkd package, if any

ekohl:
On Red Hat systems networkd is packaged as systemd-networkd. In particular:

* EL7
* EL8/EL9 shipped in EPEL
* Fedora since 33


spec/acceptance/networkd_spec.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+require 'spec_helper_acceptance'
+
+describe 'systemd with manage_networkd true' do
```
ekohl, line 5: Should we skip this on EL8, since we don't install EPEL?

Also, OracleLinux 7 looks messed up.

kenyon:
#318 is the issue for this.

spec/acceptance/networkd_spec.rb
ekohl: Implemented: it now installs EPEL

PR #383, opened 2023-10-26
(#382) Update types to allow for % where appropriate

### LINKED ISSUE #382: systemd::journald SystemMaxUse=, SystemKeepFree=,RuntimeMaxUse=, RuntimeKeepFree=, support using %'s instead of explicit sizing
Journald.conf supports using %'s in certain settings, but your types only allow for (K|M|G|T|P|E). Using a % in an expected value throws an error.
"parameter 'journald_settings' entry 'SystemMaxUse' expects a match for Variant[Pattern[/^[0-9]+(K|M|G|T|P|E)?$/], Systemd::JournaldSettings::Ensure = Struct[{'ensure' => Enum['absent', 'present']}]], got '10%"

- Puppet: all
- Ruby: all
- Distribution: all
- Module version: all

Set a value to a '%' in hiera and puppet fails to apply

## What are you seeing
"parameter 'journald_settings' entry 'SystemMaxUse' expects a match for Variant[Pattern[/^[0-9]+(K|M|G|T|P|E)?$/], Systemd::JournaldSettings::Ensure = Struct[{'ensure' => Enum['absent', 'present']}]], got '10%"
For it to update the /etc/systemd/journald.conf file appropriately
## Output log
"parameter 'journald_settings' entry 'SystemMaxUse' expects a match for Variant[Pattern[/^[0-9]+(K|M|G|T|P|E)?$/], Systemd::JournaldSettings::Ensure = Struct[{'ensure' => Enum['absent', 'present']}]], got '10%"

rgardner4012:
Updates systemd::journald  RuntimeMaxUse=, RuntimeKeepFree=,SystemMaxUse=, SystemKeepFree=, types to support using percentages

Fixes #382


kenyon:
Are percentages really allowed? I see https://github.com/systemd/systemd/issues/9548 and https://bugzilla.redhat.com/show_bug.cgi?id=874631

rgardner4012:
> Are percentages really allowed? I see [systemd/systemd#9548](https://github.com/systemd/systemd/issues/9548) (a request for enhancement that is still open) and https://bugzilla.redhat.com/show_bug.cgi?id=874631
> 
> The documentation doesn't actually say that you can use percentages: https://www.freedesktop.org/software/systemd/man/latest/journald.conf.html#SystemMaxUse=

I was going off of this in your second link

"The first pair defaults to 10% and the second to 15% of the size of the respective file system, but each value is capped to 4G. If the file system is nearly full and either SystemKeepFree= or RuntimeKeepFree= are violated when systemd-journald is started, the limit will be raised to the percentage that is actually free. This means that if there was enough free space before and journal files were created, and subsequently something else causes the file system to fill up, journald will stop using more space, but it will not be removing existing files to reduce the footprint again, either. Also note that only archived files are deleted to reduce the space occupied by journal files. This means that, in effect, there might still be more space used than SystemMaxUse= or RuntimeMaxUse= limit after a vacuuming operation is complete."

kenyon:
Right, but that doesn't say that you can use percentages. Also https://lists.freedesktop.org/archives/systemd-devel/2014-October/023596.html which was in reply to https://lists.freedesktop.org/archives/systemd-devel/2014-September/023262.html.

rgardner4012:
You are correct, it reads the values without complaining but doesn't actually use them. What is the best way to get this backed out? Do I need to open a PR with the change reverted?

kenyon:
The revert button here does it: https://github.com/voxpupuli/puppet-systemd/pull/383#event-10784048945

PR #392, opened 2023-11-16
Add `StateDirectory` to `Systemd::Unit::Service`

henrixh:
Allow for the `StateDirectory` service option to be used.

### DIFF
--- a/types/unit/service.pp
+++ b/types/unit/service.pp
@@ -103,6 +103,7 @@ type Systemd::Unit::Service = Struct[
     Optional['PrivateTmp']                => Boolean,
     Optional['RuntimeDirectory']          => String,
     Optional['RuntimeDirectoryMode']      => Stdlib::Filemode,
+    Optional['StateDirectory']            => String,
     Optional['LogsDirectory']             => String,
     Optional['LogsDirectoryMode']         => Stdlib::Filemode,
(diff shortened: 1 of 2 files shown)


types/unit/service.pp
```diff
@@ -103,6 +103,7 @@
     Optional['PrivateTmp']                => Boolean,
     Optional['RuntimeDirectory']          => String,
     Optional['RuntimeDirectoryMode']      => Stdlib::Filemode,
+    Optional['StateDirectory']            => String,
```
bastelfreak, line 106: should we enforce an absolute path here with Stdlib::Absolutepath?

henrixh: No, it's not an absolute path. I'm not sure it would even qualify as a relative path. Possibly an array of strings?  See docs on https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html

> These options take a whitespace-separated list of directory names. The specified directory names must be relative, and may not include "..". If set, when the unit is started, one or more directories by the specified names will be created (including their parents) below the locations defined in the following table.

henrixh: If a better type is desired, at least these fields should also be updated to reflect that, as they have the same semantics for their value(s):

- `RuntimeDirectory`
- `StateDirectory`
- `CacheDirectory`
- `LogsDirectory`
- `ConfigurationDirectory`

Personal preference - leave it at `String`.

PR #399, opened 2023-12-05
Correct coredump parameter from MaxFree to KeepFree

### LINKED ISSUE #398: Invalid option `MaxFree` in `cordedump.conf`
The option `MaxFree` does not show up in [the docs](https://www.freedesktop.org/software/systemd/man/latest/coredump.conf.html#MaxUse=). But there is an option `KeepFree`. Is that an old option or is it a bug?

traylenator:

Incorrect coredump paramters were added.

It should be `KeepFree` and `MaxFree` was never valid.

https://www.freedesktop.org/software/systemd/man/latest/coredump.conf.html#MaxUse=

* Fixes #398

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,7 +452,7 @@ class{'systemd':
     'ExternalSizeMax' => '10G',
     'JournalSizeMax'  => '20T',
     'MaxUse'          => '1E',
-    "MaxFree'         => '1P',
+    "KeepFree'        => '1P',
   }
 }
 ```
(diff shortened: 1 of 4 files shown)


traylenator:
> Another occurrence here with a quoting typo:
> 
> https://github.com/voxpupuli/puppet-systemd/blame/155ed514d332c350e16f7c26072a98bd653412db/README.md#L455C6-L455C13

Good catch.

kenyon (approved):

PR #401, opened 2023-12-14
Allow percent (%) character in unit names.

traylenator:

This allows relations to be set up between similar instances of two templated units.

For example:

```puppet
systemd::manage_dropin{'user-aklog.conf':
  ensure     => present,
  unit       => 'user@.service',
  unit_entry => {
    'Documentation' => 'Run an aklog before we start systemd --user',
    'After'         => ['user-aklog@%i.service'],
    'Requires'      => ['user-aklog@%i.service'],
  },
}
```

Previously the `%` character was considered illegal in a unit name.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -837,6 +837,18 @@ systemd::manage_dropin { 'maxloglevel.conf':
 }
 ```
 
+##### have a unit instance auto run before user-<uid>.service
+
+```puppet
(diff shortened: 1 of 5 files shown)


bastelfreak (approved):

smortex (approved):

traylenator:
A brilliant ratio of code to docs and tests. 🥱

PR #405, opened 2024-01-22
`systemd::unit_file`: Ensure link gets removed on `ensure => absent`

baurmatt:
Currently the `systemd::unit_file` type ignores the `ensure` parameter if `target` is defined. This allows the `ensure` parameter to be absent and thus removing the unit file if `target` is set.

### DIFF
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -93,7 +93,10 @@ define systemd::unit_file (
   }
 
   if $_target {
-    $_ensure = 'link'
+    $_ensure = $ensure ? {
+      'absent' => 'absent',
+      default  => 'link',
+    }
   } else {
     $_ensure = $ensure ? {
       'present' => 'file',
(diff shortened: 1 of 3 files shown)


TheMeier (approved):

manifests/unit_file.pp
```diff
@@ -93,7 +93,10 @@
   }
 
   if $_target {
-    $_ensure = 'link'
+    $_ensure = $ensure ? {
+      'absent' => 'absent',
+      default  => 'link',
+    }
```
kenyon, line 99: There is a [stdlib function](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/functions/ensure.pp) for this:

```suggestion
    $_ensure = stdlib::ensure($ensure, link)
```

kenyon (changes requested):

manifests/unit_file.pp
kenyon: The stdlib requirement in `metadata.json` would also need to be set to minimum of 6.6.0 to use `stdlib::ensure()`.

baurmatt: Changed, thanks! :)

kenyon (approved):

PR #406, opened 2024-01-29
fix: refresh service only based on drop-in file changes

shieldwed:

Since multiple drop-in files of the same unit notify the same `Systemd::Daemon_reload` resource, changing a drop in file without `notify_service` still triggers a service refresh if at least one other drop-in file has `notify_service` set (default) regardless of it being changed.

Therefore, the `Systemd::Daemon_reload` should only be ordered before the `Service` but not notify it. The `service` will already be refreshed if the drop-in file changes and `notify_service` is set (default).

#### This Pull Request (PR) addresses the following issues

Consider the following Puppet code snippet with two drop-in files for the service `nginx.service`:

```Puppet
  systemd::dropin_file {
    default:
      unit => 'nginx.service',
      ;
    "restart.conf":
      notify_service => false,
      content        => @(EOT)
        [Service]
        Restart=on-failure
        | EOT
      ,
    ;
    "ulimit.conf":
      content => @(EOT)
        [Service]
        LimitNOFILE=4096
        | EOT
      ,
  }

```
On the first Puppet-run, this code will correctly refresh the service, as the `ulimit.conf`-file will be written. Once the drop-in files are in sync with this configuration, no more SystemD daemon reloads and service refreshes happen.

Now, changing the `Restart` configuration of `nginx.service` doesn't need the service itself to be restarted.

However, due to the second drop-in file specifying `notify_service => true` (default), this will lead to `Systemd::Daemon_reload[$unit]` being triggered (expected due to `daemon_reload => true` (default)), which then **in turn** triggers `Service[nginx.service]` and `Service[nginx]` if they exist.

### DIFF
--- a/manifests/dropin_file.pp
+++ b/manifests/dropin_file.pp
@@ -81,7 +81,7 @@ define systemd::dropin_file (
     File[$full_filename] ~> Service <| title == $unit or name == $unit |>
 
     if $daemon_reload {
-      Systemd::Daemon_reload[$unit] ~> Service <| title == $unit or name == $unit |>
+      Systemd::Daemon_reload[$unit] -> Service <| title == $unit or name == $unit |>
     }
 
     if $unit =~ /\.service$/ {
@@ -89,7 +89,7 @@ define systemd::dropin_file (
       File[$full_filename] ~> Service <| title == $short_service_name or name == $short_service_name |>
 
       if $daemon_reload {
-        Systemd::Daemon_reload[$unit] ~> Service <| title == $short_service_name or name == $short_service_name |>
+        Systemd::Daemon_reload[$unit] -> Service <| title == $short_service_name or name == $short_service_name |>
       }
     }
   }
--- a/spec/defines/dropin_file_spec.rb
+++ b/spec/defines/dropin_file_spec.rb
@@ -69,8 +69,30 @@ describe 'systemd::dropin_file' do
 
             it { is_expected.to compile.with_all_deps }
             it { is_expected.to contain_service('myservice').that_subscribes_to("File[#{filename}]") }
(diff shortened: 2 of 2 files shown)


manifests/dropin_file.pp
```diff
@@ -81,15 +81,15 @@
     File[$full_filename] ~> Service <| title == $unit or name == $unit |>
 
     if $daemon_reload {
-      Systemd::Daemon_reload[$unit] ~> Service <| title == $unit or name == $unit |>
+      Systemd::Daemon_reload[$unit] -> Service <| title == $unit or name == $unit |>
```
ekohl, line 84: If this isn't a notify anymore, should this be around line 77 instead? I think the ordering is still correct. Same for the other case.

ekohl (commented):
This probably impossible to test due to rspec-puppet having different transitive dependencies than real Puppet: https://github.com/puppetlabs/rspec-puppet/pull/35 and also missing matchers (https://github.com/puppetlabs/rspec-puppet/issues/28).

manifests/dropin_file.pp
shieldwed: Thanks for taking a look at this @ekohl. The conditions stay the same. The ordering of the `Daemon_reload` in relation to the `Service` still only needs to happen when both `$notify_service` and `$daemon_reload` are true.

shieldwed:
> This probably impossible to test due to rspec-puppet having different transitive dependencies than real Puppet ...
I've implemented some tests for the changed cases. Can you please look at them again?

shieldwed:
@ekohl can you please have another look at this PR? Thank you

shieldwed:
Can someone please have a look at this again (e.g. @bastelfreak or @TheMeier perhaps)? Thank you.

TheMeier (approved):

PR #409, opened 2024-02-02
Support multiple Environment Settings

### LINKED ISSUE #404: manage_unit service_entry Environment can't have multiple values

- Puppet: 8.3.1
- Ruby: 
- Distribution: Debian Bookworm
- Module version: 6.3.0

```
  systemd::manage_unit { 'test.service':
    unit_entry    => {
      'Description' => 'Bla',
    },
    service_entry => {
      'Type'        => 'oneshot',
      'Environment' => [
        "bla=foo",
        "bar=bla",
      ],
      'ExecStart'   => '/usr/bin/true',
    },
    enable        => false,
    active        => false,
  }
```

## What are you seeing

An error during the puppet run (see output log)

A systemd service that contains multiple "Environment=" lines (Which is valid as per https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Environment)

## Output log

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Systemd::Manage_unit[test.service]: parameter 'service_entry' entry 'Environment' expects a String value, got Tuple (file: /etc/puppetlabs/code/environments/j_test/modules/profile/manifests/test/test.pp, line: 23) on node xxx
```

traylenator:

It was previously impossible to set more than one environment variable:

e.g.

```puppet
systemd::manage_unit { 'test.service':
  unit_entry    => {
    'Description' => 'Bla',
  },
  service_entry => {
    'Type'        => 'oneshot',
    'Environment' => [
      'bla=foo',
      'bar=bla',
    ],
    'ExecStart'   => '/usr/bin/true',
  },
  enable        => false,
  active        => false,
}
```

Fixes #404

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -2444,7 +2444,7 @@ Struct[{
     Optional['NotifyAccess']              => Enum['none', 'default', 'main', 'exec',  'all'],
     Optional['OOMPolicy']                 => Enum['continue', 'stop','kill'],
     Optional['OOMScoreAdjust']            => Integer[-1000,1000],
-    Optional['Environment']               => String,
+    Optional['Environment']               => Variant[String[0],Array[String[1],1]],
     Optional['EnvironmentFile']           => Variant[
       Stdlib::Unixpath,Pattern[/-\/.+/],
       Array[Variant[Stdlib::Unixpath,Pattern[/-\/.+/]],1],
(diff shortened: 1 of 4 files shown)


moritzdietz:
Would you believe that I was *just* about to open a PR to fix this very thing? I wasn't expecting anyone to pay attention to this issue so quickly hah! Thanks for that

ekohl (approved):

ekohl:
Test failure is unreadable on the mobile app, but I suspect it's the timeout again

moritzdietz:
> Test failure is unreadable on the mobile app, but I suspect it's the timeout again

Yup, the CI job says "read timeout reached" as the error message

ekohl:
I wonder why it happens so often. Perhaps it's quay.io that's unreliable. Maybe also beaker-docker that could do something smarter, like retry. Right now I don't even know where it exactly fails

PR #410, opened 2024-02-05
Add loginctl_linger_users fact

jcpunk:
Add a fact listing users with linger enabled.

As near as I can tell from the systemd source, `root` is always able to linger even it not setup explicitly.

### DIFF
--- a/spec/unit/facter/loginctl_linger_users_spec.rb
+++ b/spec/unit/facter/loginctl_linger_users_spec.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Facter.fact(:loginctl_linger_users) do
+  before do
+    Facter.clear
+  end
+
+  describe 'loginctl_linger_users' do
+    context 'returns root when nothing is present' do
+      it do
+        allow(Dir).to receive(:exist?).and_return(false)
+        allow(Dir).to receive(:exist?).with('/var/lib/systemd/linger').and_return(false)
+
+        expect(Facter.value(:loginctl_linger_users)).to eq(['root'])
+      end
+    end
+
+    context 'returns list with root others are present' do
+      it do
+        allow(Dir).to receive(:exist?).and_return(false)
+        allow(Dir).to receive(:exist?).with('/var/lib/systemd/linger').and_return(true)
+        allow(Dir).to receive(:entries).with('/var/lib/systemd/linger').and_return(%w[. .. abc test])
+
+        expect(Facter.value(:loginctl_linger_users)).to eq(%w[abc test root])
+      end
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


spec/unit/facter/loginctl_linger_users_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    context 'returns list with root others are present' do
+      before do
+        allow(Dir).to receive(:exist?).with('/var/lib/systemd/linger').and_return(true)
+        allow(Dir).to receive(:entries).with('/var/lib/systemd/linger').and_return(['.', '..', 'abc', 'test'])
+      end
+
+      it do
+        expect(Facter.value(:loginctl_linger_users)).to eq(['abc', 'test', 'root'])
```
zilchms, line 28: try %w[abc test root] instead of ['abc', 'test', 'root']

zilchms (changes requested):

zilchms (approved):
I am willing to let it pass with the tests written the way they are, but i would suggest we find someone with better rspec knowledge to go over it too before we merge

h0tw1r3:
Generally not in favor of adding facts like this, particularly on widely used modules. A function or resource may be more appropriate.

That said, what's an sample use case?

h0tw1r3 (changes requested):
Generally not in favor of adding facts like this, particularly on widely used modules. A function or resource may be more appropriate.

That said, what's an sample use case?

jcpunk:
My overall plan is to feed this value back into
```
systemd::logind_settings:
    KillExcludeUsers:
```
So my users with `linger` will have the ability to trivially persist processes.

Pat

h0tw1r3:
I'm confused, isn't that redundant if touching a /var/lib/systemd/linger/<user> file is the same as adding <user> to KillExcludeUsers in the logind conf file?

traylenator:
> I'm confused, isn't that redundant if touching a /var/lib/systemd/linger/ file is the same as adding to KillExcludeUsers in the logind conf file?

Maybe a slight difference:

I guess enable linger will stop things being killed that are running in the `systemd --user` scope however all other user processes would be killed. Better to encourage users to run long lived processes  in the correct context if they want them to survive.

I'm kind of worried about this fact as it would be pretty large in our case.

jcpunk:
If this fact looks to be too much, I can always run it out of one of my private modules.  Just figured I'd default to "share with the community".

h0tw1r3:
> I'm kind of worried about this fact as it would be pretty large in our case.

Agree. My basic objection to a fact like this.

jcpunk:
No worries, I'll close this out and run the fact privately.

traylenator:
If I was going to do this and am actually thinking about it I would not rely on puppet to maintain that list. I'd set up a path unit and service and sync `/var/lib/systemd/linger` entries to `KillExcludeUsers` settings.

PR #416, opened 2024-02-28
Support reload of `systemd --user` for non-zero UIDs

traylenator:

This change increased the minimum required Puppet version to 6.24.0 or 7.9.0 [PUP-5704](https://puppet.atlassian.net/browse/PUP-5704) to support arrays to the command attribute of the exec type.

Support calling the `systemd --user` for a particular uid.

A `uid` must be used rather than the probably more convenient username since the `$XDG_RUNTIME_DIR` for the user must be known.

This change assumes that `XDG_RUNTIME_DIR` is `/run/user/<uid>` for all uids.

Example run:
```puppet
notify{'junk':
 notify => Systemd::Daemon_reload['foobar'],
}

systemd::daemon_reload{'foobar':
  uid => 1000,
}
```

results in

```
Notice: Compiled catalog for fedora. in environment production in 0.05 seconds
Notice: junk
Notice: /Stage[main]/Main/Notify[junk]/message: defined 'message' as 'junk'
Notice: /Stage[main]/Main/Systemd::Daemon_reload[foobar]/Exec[systemd-foobar-systemctl-user-1000-daemon-reload]: Triggered 'refresh' from 1 event
Notice: Applied catalog in 0.17 seconds
```
and a journal (debug on) of

```
Feb 28 11:44:34 fedora systemd[1]: user@1000.service: Got notification message from PID 1877 (RELOADING=1, MONOTONIC_USEC=5874581103)
Feb 28 11:44:34 fedora systemd[1]: user@1000.service: Changed running -> reload-notify
Feb 28 11:44:34 fedora systemd[1]: user@1000.service: Installed new job user@1000.service/nop as 6151
...

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -658,12 +658,35 @@ Default value: `['create']`
 
 Run systemctl daemon-reload
 
+#### Examples
+
+##### Force reload the system systemd
+
+```puppet
+notify{ 'fake event to notify from':
+  notify => Systemd::Daemon_reload['special']
+}
+systemd::daemon_reload{ 'special': }
+```
+
+##### Force reload a systemd --user
+
+```puppet
+notify{ 'fake event to notify from':
+  notify => Systemd::Daemon_reload['user']
+}
+systemd::daemon_reload{ 'user':
+  uid => 1234,
+}
+```
+
 #### Parameters
 
 The following parameters are available in the `systemd::daemon_reload` defined type:
 
 * [`name`](#-systemd--daemon_reload--name)
 * [`enable`](#-systemd--daemon_reload--enable)
+* [`uid`](#-systemd--daemon_reload--uid)
 
 ##### <a name="-systemd--daemon_reload--name"></a>`name`
 
@@ -674,10 +697,19 @@ A globally unique name for the resource
 Data type: `Boolean`
 
 Enable the reload exec
-
 * Added in case users want to disable the reload globally using a resource collector
 
 Default value: `true`
+
+##### <a name="-systemd--daemon_reload--uid"></a>`uid`
+
+Data type: `Optional[Integer[1]]`
+
+Specify uid of `systemd --user` to reload. When `uid` is left `undef` the system
+systemd instance will be reloaded. It is assumed that the `XDG_RUNTIME_DIR` for
+the user is `/run/user/<uid>`.
+
+Default value: `undef`
 
 ### <a name="systemd--dropin_file"></a>`systemd::dropin_file`
 
--- a/manifests/daemon_reload.pp
+++ b/manifests/daemon_reload.pp
@@ -7,15 +7,51 @@
 #
 # @param enable
 #   Enable the reload exec
-#
 #   * Added in case users want to disable the reload globally using a resource collector
+#
+# @param uid
+#   Specify uid of `systemd --user` to reload. When `uid` is left `undef` the system
+#   systemd instance will be reloaded. It is assumed that the `XDG_RUNTIME_DIR` for
+#   the user is `/run/user/<uid>`.
+#
+# @example Force reload the system systemd
+#   notify{ 'fake event to notify from':
+#     notify => Systemd::Daemon_reload['special']
+#   }
+#   systemd::daemon_reload{ 'special': }
+#
+# @example Force reload a systemd --user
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: REFERENCE.md [hunk 1]
```diff
@@ -622,12 +622,35 @@ Default value: `['create']`
...
 
+#### Examples
+
+##### Force reload the system systemd
+
+```puppet
+notify{ 'fake event to notify from':
+  notify => Systemd::Daemon_reload['special']
+}
+systemd::daemon_reload {'special':}
```
kenyon, line 633:
```suggestion
notify { 'fake event to notify from':
  notify => Systemd::Daemon_reload['special'],
}
systemd::daemon_reload { 'special': }
```

kenyon (approved):
Style nits, otherwise 🚀

manifests/daemon_reload.pp
```diff
@@ -7,15 +7,51 @@
...
-    exec { "${module_name}-${name}-systemctl-daemon-reload":
-      command     => 'systemctl daemon-reload',
+    # For a `systemd --user` instance XDG_RUNTIME_DIR must be set so dbus
+    # can be found.
+
+    if $uid {
+      $_title   = "${module_name}-${name}-systemctl-user-${uid}-daemon-reload"
+      $_user    = String($uid)  # exec seems unhappy with integers.
+      $_env     = "XDG_RUNTIME_DIR=/run/user/${uid}"
+      $_command = 'systemctl --user daemon-reload'
```
bastelfreak, line 43: can you switch the commands to an array: https://www.puppet.com/docs/puppet/7/types/exec.html#exec-attribute-command

traylenator:
It is going to be a pain passing in the uid since you often do not know it.

Open to alternatives?

Having a fact that provides details on the running `systemd --user`s seems bad with terrible race conditions.

traylenator:
> @traylenator are you fine with this for now or should we wait for a solution to get the UID?

I am fine with this as is. It unlocks quite a few things for Podman in particular and if we can support a user attribute all of this is still correct.

traylenator:
> > It is going to be a pain passing in the uid since you often do not know it.
> 
> Would a deferred function to resolve a username to uid be possible?

You can certainly do a deferred inline template for the exec command - less sure about the user parameter. Will have a play after this. 

A [getpwnam](https://github.com/wcooley/puppet-name_service_lookups/blob/master/lib/puppet/parser/functions/getpwnam.rb)  is a very generic function so would probably add it to extlib anyway.

traylenator:
> I don't mind the array, but https://www.puppet.com/docs/puppet/7/types/exec.html#exec-attribute-command should really document that it was [PUP-5704](https://puppet.atlassian.net/browse/PUP-5704) that implemented it and you need at least Puppet 6.24.0 or Puppet 7.9.0.
> 
> Because of that we should raise the minimum version in `metadata.json`.

min puppet version bumped.

traylenator:
[getpwnam](https://github.com/wcooley/puppet-name_service_lookups/blob/master/lib/puppet/parser/functions/getpwnam.rb)  is a very generic function so would probably add it to extlib anyway.

Wrong function:

It would be
 https://github.com/wcooley/puppet-name_service_lookups/blob/master/lib%2Fpuppet%2Fparser%2Ffunctions%2Fgetpwuid.rb

of course.

bastelfreak:
@traylenator oh having those in extlib would be awesome! (and maybe be converted to the modern API :eyes: )

ekohl:
Almost feels generic enough to belong in stdlib.

traylenator:
Rebased - since we crossed the backwards incompatible threshold - would be good.

bastelfreak (approved):

traylenator:
> the code looks fine to me. But is it actually a backwards-incompatible change?

Only for exec.command = Array , so increase in puppet version.

traylenator:
I made the mistake at looking at this again:

```
machinectl shell mycentos@.host /usr/bin/systemctl --user daemon-reload
```
ran as root avoids the need the UID and knowing where things are.

traylenator:
#443 is a much better solution and avoids all the user -> uid hassles.

traylenator:
Thanks for all the reviews - I've hopefully taken over all the comments to the new one.

PR #419, opened 2024-03-02
feat: add a cron like wrapper for timers

### LINKED ISSUE #374: add a cron-like systemd::timer interface
I'm looking at converting my `cron` resources into a `systemd::timer` but the interface to that resources is rather unwieldy. It assumes prior and rather intricate knowledge of the format of systemd unit files and makes the cron resource look extremely intuitive and fun in comparison (which is not a compliment).

I think there's room for "cron-like" timer interface. We're not talking about reinventing the wheel here, probably just a little template will do. An interface such as:

```
define systemd::unit::job(
    $command,
    $user => root,
    $interval => # [...] that is of course the hard part
) {

    systemd::timer { $name:
        # ....
    }
}
```

Wikimedia operations people actually did this in their own `systemd` module which, strangely, is not related to this module but shares a lot of the interfaces, see their [systemd::timer::job resource](https://gerrit.wikimedia.org/r/plugins/gitiles/operations/puppet/+/fd8b59c7f0be1bcd7a4927d5d572986fe8c9cb47/modules/systemd/manifests/timer/job.pp). It does a *lot* more than just create a timer though: it includes things like logging, monitoring and so on, probably not something we want to merge as-is just now. I mostly cite it as an example of how things could be done.

It makes conversions much easier, here's a [patch of one conversion](https://gerrit.wikimedia.org/r/plugins/gitiles/operations/puppet/+/a22f0afa89e641a792cea13475efff256029fb36%5E%21/). I particularly like the `> /dev/null 2>&1` removal, personally. :)

I mostly open this to get feedback, I'm likely going to implement this on our side anyway, but I figured it could be useful for folks here if I send it as a PR instead of just our secret sauce. See [our internal ticket as well](https://gitlab.torproject.org/tpo/tpa/team/-/issues/41303).

TheMeier:
This helper will allow to define a timer in a cron like way, and will setup a corresponding timer and service.
In basic usage it only requires a command, user, timerspec and an ensure parameter.
For additional feaures the parameter `service_overrides`, `timer_overrides`, `service_unit_overrides` and `timer_unit_overrides` can be used to define additional aspects of the create units.
The correct order of the units is also ensured.

Fixes #374
Related to https://github.com/voxpupuli/puppet-systemd/pull/287


types/unit/timer/timerspec.pp
```diff
@@ -0,0 +1,4 @@
+# @summary Timer specification for systemd timers
+# @see https://www.freedesktop.org/software/systemd/man/timer.unit.html
```
kenyon, line 2: I get 404 for this. Did you mean https://www.freedesktop.org/software/systemd/man/latest/systemd.timer.html?

manifests/timer_wrapper.pp
```diff
@@ -0,0 +1,92 @@
...
+  Optional[Systemd::Unit::Timer::TimerSpec] $on_start_up_sec = undef,
+  Optional[Systemd::Unit::Timer::TimerSpec] $on_unit_active_sec = undef,
+  Optional[Systemd::Unit::Timer::TimerSpec] $on_unit_inactive_sec = undef,
+  Optional[Systemd::Unit::Timer::TimerSpec] $on_calendar = undef,
+  Systemd::Unit::Service                    $service_overrides         = {},
+  Systemd::Unit::Timer                      $timer_overrides           = {},
+  Systemd::Unit::Unit                       $timer_unit_overrides      = {},
+  Systemd::Unit::Unit                       $service_unit_overrides    = {},
+) {
+  $_timer_spec = delete_undef_values({
```
kenyon, line 35: I think using built-in functions would be preferable, as documented in https://forge.puppet.com/modules/puppetlabs/stdlib/reference#delete_undef_values

manifests/timer_wrapper.pp
```diff
@@ -0,0 +1,92 @@
...
+        on_unit_inactive_sec,
+        or on_calendar must be set'
+      )
+    }
+    if ! $command {
+      fail('command must be set')
+    }
+  }
+
+  $service_ensure = $ensure ? { 'absent' => false,  default  => true, }
```
kenyon, line 59: For these situations we have a nice function `stdlib::ensure`: https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/REFERENCE.md#stdlib--ensure

But then I guess you still have to do something like this selector for the service `enable`.

TheMeier: Right. This would make this one line nicer, but add another line with basically the same content. So I think the current implementation is justified.

manifests/timer_wrapper.pp
TheMeier: Changed. Although the readability is better when using `delete_undef_values()` IMHO

manifests/timer_wrapper.pp
```diff
@@ -0,0 +1,92 @@
...
+#   example: `'service_overrides' => { 'Wants' => 'network-online.target' }`
+# @param timer_overrides override for the`[Timer]` section of the timer
+#   example: `'timer_overrides' => { 'OnBootSec' => '10' }`
+# @param service_unit_overrides override for the`[Unit]` section of the service
+#   example: `'service_unit_overrides' => { 'Wants' => 'network-online.target' }`
+# @param timer_unit_overrides override for the `[Unit]` section of the timer
+#   can not be used in combination with `additional_timer_params`
+define systemd::timer_wrapper (
+  Enum['present', 'absent']                 $ensure,
+  Optional[String[1]]                       $command = undef,
```
traylenator, line 22: As you check below `$command` has to be set. 

Can't you just make it non-optional here in the first place. Also there is an existing type
`Systemd::Unit::Service::Exec` which presumably makes sense here.

Could also default this to `$name`  so e.g

```puppet
systemd::timer_wrapper('/usr/bin/rm -rf /':
   on_boot_sec => '1h',
 }
```
could be done.

manifests/timer_wrapper.pp
```diff
@@ -0,0 +1,92 @@
...
+  Systemd::Unit::Unit                       $service_unit_overrides    = {},
+) {
+  $_timer_spec = {
+    'OnActiveSec'       => $on_active_sec,
+    'OnBootSec'         => $on_boot_sec,
+    'OnStartUpSec'      => $on_start_up_sec,
+    'OnUnitActiveSec'   => $on_unit_active_sec,
+    'OnUnitInactiveSec' => $on_unit_inactive_sec,
+    'OnCalendar'        => $on_calendar,
+  }.filter |$k, $v| { $v =~ NotUndef }
```
traylenator, line 42:
```suggestion
  }.delete_undef_values
```
Not wrong what you did, just an alternative - you decide.

manifests/timer_wrapper.pp
```diff
@@ -0,0 +1,92 @@
...
+    if $_timer_spec == {} {
+      fail('At least one of on_active_sec,
+        on_boot_sec,
+        on_start_up_sec,
+        on_unit_active_sec,
+        on_unit_inactive_sec,
+        or on_calendar must be set'
+      )
+    }
+    if ! $command {
```
traylenator, line 54: This can go if the parameter is not optional.

manifests/timer_wrapper.pp
```diff
@@ -0,0 +1,92 @@
...
+      fail('command must be set')
+    }
+  }
+
+  $service_ensure = $ensure ? { 'absent' => false,  default  => true, }
+  $unit_name = regsubst($title, '/' , '_', 'G')
```
traylenator, line 60:
```suggestion
  $unit_name = systemd::escape($title)
```
may as well use the supplied function which covers even stranger cases.

manifests/timer_wrapper.pp
```diff
@@ -0,0 +1,92 @@
...
+  systemd::manage_unit { "${unit_name}.service":
+    ensure        => $ensure,
+    unit_entry    => $service_unit_overrides,
+    service_entry => {
+      'ExecStart' => $command, # if ensure present command is defined is checked above
+      'User'      => $user, # defaults apply
+      'Type'      => 'oneshot',
+    }.filter |$key, $val| { $val =~ NotUndef } + $service_overrides,
+  }
+  systemd::manage_unit { "${unit_name}.timer":
```
traylenator, line 71: The timer must an have `[Install]` section or else it is impossible to enable. For info puppet just skips over such units for enable that can't be enabled.

manifests/timer_wrapper.pp
```diff
@@ -0,0 +1,92 @@
...
+    ensure      => $ensure,
+    unit_entry  => $timer_unit_overrides,
```
traylenator, line 73: This will be undef if there is no `$timer_unit_overrides`  . I think would be good to always have a 
```config
[Unit]
Description=Puppet Generated Timer from $title
```
or something.

manifests/timer_wrapper.pp
```diff
@@ -0,0 +1,92 @@
...
+      fail('command must be set')
+    }
+  }
+
+  $service_ensure = $ensure ? { 'absent' => false,  default  => true, }
+  $unit_name = regsubst($title, '/' , '_', 'G')
+
+  systemd::manage_unit { "${unit_name}.service":
+    ensure        => $ensure,
+    unit_entry    => $service_unit_overrides,
```
traylenator, line 64: Set a minimum unit_entry with a Description generated from the `$title` .
```config
[Unit]
Description=Puppet generated service $title
```
or something.
Don't think a `Description` is mandatory but it seems like a good idea to always be set.

spec/defines/timer_wrapper_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+
+        context 'ensure absent' do
+          let :params do
+            {
+              ensure: 'absent',
+            }
+          end
+
+          it {
+            is_expected.to contain_Systemd__Manage_unit("#{title}.timer")
```
traylenator, line 98: Would be good to verify the ensure is really absent here - same for all the other service and timers.

types/unit/timer/timerspec.pp
```diff
@@ -0,0 +1,4 @@
+# @summary Timer specification for systemd timers
+# @see https://www.freedesktop.org/software/systemd/man/systemd.timer.html
+#
+type Systemd::Unit::Timer::TimerSpec = Variant[Integer[0],String,Array[Variant[Integer[0],String]]]
```
traylenator, line 4: Should add its own rspec file in `spec/type_aliases` could just copy paste or even move the examples out of the existing timer one. Good for future  if and when its being fixed.

traylenator (changes requested):

manifests/timer_wrapper.pp
```diff
@@ -0,0 +1,92 @@
...
+# @param on_unit_inactive_sec run service relative to when the unit was last deactivated
+# @param on_calendar the calendar event expressions time to run the service
+# @param service_overrides override for the`[Service]` section of the service
+#   example: `'service_overrides' => { 'Wants' => 'network-online.target' }`
+# @param timer_overrides override for the`[Timer]` section of the timer
+#   example: `'timer_overrides' => { 'OnBootSec' => '10' }`
+# @param service_unit_overrides override for the`[Unit]` section of the service
+#   example: `'service_unit_overrides' => { 'Wants' => 'network-online.target' }`
+# @param timer_unit_overrides override for the `[Unit]` section of the timer
+#   can not be used in combination with `additional_timer_params`
```
traylenator, line 19: Would be good to add a couple of `@examples` here I think.

manifests/timer_wrapper.pp
TheMeier: I made the optional so you don't need to specify it when `ensure = 'absent'` the check is in a dedicated if clause because of this.

Using `Systemd::Unit::Service::Exec` is a good idea ... changed.

I am not a fan of putting the command in the title. This can lead to really ugly resource names in the catalog, also it bad for readability IMHO

manifests/timer_wrapper.pp
TheMeier: Please see https://github.com/voxpupuli/puppet-systemd/pull/419#discussion_r1510086533. I specifically changed it from `delete_undef_values` to rely less on stdlib and more on builtins.

manifests/timer_wrapper.pp
TheMeier: As mentioned out above, I really want this optional

manifests/timer_wrapper.pp
TheMeier: Great hint. Implemented, thanks!

manifests/timer_wrapper.pp
TheMeier: Good catch, thanks!

manifests/timer_wrapper.pp
TheMeier: I was considering making `description` a mandatory parameter.  I could not see what the benefit actually is. If you have a policy to require descriptions, it should be easy enough to enforce that otherwise. A generic description reflecting the title has no inherit value IMHO

spec/defines/timer_wrapper_spec.rb
TheMeier: done

manifests/timer_wrapper.pp
TheMeier: done

manifests/timer_wrapper.pp
TheMeier: See https://github.com/voxpupuli/puppet-systemd/pull/419#discussion_r1510351683

types/unit/timer/timerspec.pp
TheMeier: done

manifests/timer_wrapper.pp
traylenator: It appears in the logs as starting Description and stopping Description.

No clue what is logged if you don't set it. As long as it's informative enough to trace fine.

Also appears in `systemctl status` of course.

TheMeier: Example:
```
Feb 15 19:31:19 localhost systemd[1]: Started logrotate.timer - Daily rotation of log files.
```
```
 sudo systemctl status logrotate.timer
● logrotate.timer - Daily rotation of log files
```

I see limited value here, which depends very much on how dilligent users actually are. As this is meant as a "cron like" interface, is it worth it? Might be annoying to some users.

TheMeier: Especially a generic default reiterating the name is not providing any benefit IMHO.
I will resolve this. But please feel free to mark it again. If you really think it is worth it I will add it.

types/unit/timer/timerspec.pp
traylenator, line 4:
```suggestion
type Systemd::Timespan = Variant[Integer[0],String,Array[Variant[Integer[0],String]]]
```
From `man systemd.timer` and `man systemd.time` these values seem to be called `time span`s

Looking around these "Time Spans" are also used in say `Restart` of [Service],  

Not suggesting all of those should be adjusted to this now but it make sense I think to add `Timespan` more generic than the time directory. Makes them less silly long as well seems a good idea.

TheMeier: updated.

traylenator (approved):
Useful.

types/unit/timespan.pp
```diff
@@ -0,0 +1,4 @@
+# @summary Timer specification for systemd timers
+# @see https://www.freedesktop.org/software/systemd/man/systemd.timer.html
```
traylenator, line 2:
```suggestion
# @see https://www.freedesktop.org/software/systemd/man/systemd.time.html
```
Sorry.

types/unit/timespan.pp
```diff
@@ -0,0 +1,4 @@
+# @summary Timer specification for systemd timers
```
traylenator, line 1:
```suggestion
# @summary Timer specification for systemd time spans, e.g. timers.
```

types/unit/timespan.pp
TheMeier: Why sorry? I wish all puppet modules would have this kind of quality(-control) :smiley:

PR #422, opened 2024-03-05
Use Systemd::Unit::Amount, Percent and AmountOrPercent

traylenator:

Use the new types:
* `Systemd::Unit::Amount`
* `Systemd::Unit::Percent`
* `Systemd::Unit::AmountOrPercent`

A few points to note:

* While **some** documentation only mentions up to Tera the systemd code supports up to Peta and Exa. Some man pages are correct.
* The `MemoryLow`, `MemoryHigh`, `MemoryHigh` matchers were incorrect. They do not support a "Soft:Hard" setting like the ulimit settings `LimitFSize` and similar.

* https://github.com/systemd/systemd/blame/main/src/basic/format-util.c


traylenator:
centos failures are not related.

PR #426, opened 2024-03-08
Add bolt task to return unit state in a more parsable way

rwaffen:
```text
bolt task run systemd::systemctl_show unit_name=puppet.service -t controller-0

Started on controller-0...
Finished on controller-0:
  {
    "MainPID": "686",
    "LoadState": "loaded",
    "ActiveState": "active",
    "SubState": "running",
    "UnitFileState": "enabled"
  }
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@
 .*.sw?
 /.yardoc/
 /Guardfile
+bolt-debug.log
+.rerun.json
--- a/tasks/systemctl_show.json
+++ b/tasks/systemctl_show.json
@@ -1,0 +1,20 @@
+{
+  "description": "Get systemctl show output of a unit",
+  "input_method": "stdin",
+  "parameters": {
+    "bin_path": {
+      "description": "Path to systemctl binary",
+      "type": "String[1]",
+      "default": "/usr/bin/systemctl"
+    },
+    "properties": {
+      "description": "Properties to retrieve from the unit",
+      "type": "Array[String[1]]",
+      "default": ["ActiveState", "LoadState", "MainPID", "SubState", "UnitFileState"]
+    },
+    "unit_name": {
+      "description": "Name of the unit",
+      "type": "String[1]"
+    }
+  }
+}
--- a/tasks/systemctl_show.rb
+++ b/tasks/systemctl_show.rb
@@ -1,0 +1,15 @@
+#!/opt/puppetlabs/puppet/bin/ruby
+# frozen_string_literal: true
+
+require 'json'
+require 'open3'
+
+params  = JSON.parse($stdin.read)
+
+systemctl_cmd = params['bin_path']
+systemctl_args = ['show', '--no-pager', '--no-legend', '--property', params['properties'].join(','), params['unit_name']]
+output, status = Open3.capture2(systemctl_cmd, *systemctl_args)
+
+raise "Error running systemctl show: #{output}" unless status.success?
+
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -21,3 +21,5 @@
 .*.sw?
 /.yardoc/
 /Guardfile
+bolt-debug.log
```
bastelfreak, line 24: the file is managed via modulesync_config, can you please a PR there as well? https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.gitignore.erb

bolt-project.yaml
```diff
@@ -0,0 +1,3 @@
+---
+analytics: false
+name: systemd
```
bastelfreak, line 3: I don't think we need the file? This repo isn't supposed to be a bolt project?

tasks/systemctl_show.json
```diff
@@ -0,0 +1,20 @@
+{
+  "description": "Get systemctl show output of a unit",
+  "input_method": "stdin",
+  "parameters": {
+    "bin_path": {
+      "description": "Path to systemctl binary",
+      "type": "Optional[String[1]]",
+      "default": "/bin/systemctl"
```
bastelfreak, line 8: I think on most distros `/bin` is a symlink to `/usr/bin` and `/bin/`  is supposed to go away? Should we use `/usr/bin` directly?
```suggestion
      "default": "/usr/bin/systemctl"
```

rwaffen: updated this just now :)

tasks/systemctl_show.rb
```diff
@@ -0,0 +1,15 @@
+#!/opt/puppetlabs/puppet/bin/ruby
+# frozen_string_literal: true
+
+require 'json'
+require 'open3'
+
+params  = JSON.parse($stdin.read)
+
+systemctl_cmd = params['bin_path']
+systemctl_args = ['show', '--no-pager', '--no-legend', '--property', params['properties'].join(','), params['service']]
```
bastelfreak, line 10: TIL: systemctl has a `--output` option, and you can set it to json, but it won't output json.... https://github.com/systemd/systemd/issues/83

.gitignore
rwaffen: see https://github.com/voxpupuli/modulesync_config/pull/879

tasks/systemctl_show.rb
rwaffen: broken since 2015 🫣

bolt-project.yaml
rwaffen: I removed it

bastelfreak: I figured it out. the module was probably cloned as `puppet-systemd`, but the task name is `systemd::systemctl_show`. So the directory needs to ne renamed to `systemd`.

```
$ cp -r puppet-systemd systemd
$ bolt task show systemd::systemctl_show --modulepath=.
systemd::systemctl_show
  Get systemctl show output of a unit

Usage
  bolt task run systemd::systemctl_show --targets <targets> [bin_path=<value>]
  [properties=<value>] unit_name=<value>

Parameters
  bin_path  String[1]
    Path to systemctl binary
    Default: "/usr/bin/systemctl"

  properties  Array[String[1]]
    Properties to retrieve from the unit
    Default: ["ActiveState", "SubState", "LoadState", "UnitFileState"]

  unit_name  String[1]
    Name of the unit

Module
  /home/bastelfreak/code/modulesync_config/modules/voxpupuli/systemd
```

PR #429, opened 2024-03-09
remove resolved settings from config when changed to `absent`

### LINKED ISSUE #397: Impossible to unset DNSStubListenerExtra= once set.

- Puppet: all
- Ruby: all
- Distribution: all
- Module version: 6.2.0

First set 
```puppet
class {'systemd':
   manage_resolved => true,
   use_stub_resolver => true,
   dns_stub_listener_extra => ['10.0.0.1']',
   ...
}
```
Then try unsetting it either by setting  `dns_stub_listener_extra => ''` or leaving undef. 

## What are you seeing

Neither of these allow you to unset the dnf_stub_listenener_extra. Leaving it undef will do nothing.
Setting it to '' or `['']` is illegal since.

```
Optional[Array[String[1]]] $dns_stub_listener_extra = undef,
```

Should be possible to unset the value. I expect 

```
 dns_stub_listener_extra => ['']
```
should be permitted. 

I expect this applies to some other parameters.

TheMeier:
When `dns_stub_listener` and `dns_stub_listener_extra` are changed from a value to `undef` they don't get removed from the config.

There is plenty more settings  here that could use the same treatment I guess

Fixes #397

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -65,9 +65,11 @@
 #
 # @param dns_stub_listener
 #   Takes a boolean argument or one of "udp" and "tcp".
+#   Setting it to `'absent'` will remove `DNSStubListener` existing entries from the configuration file
 #
 # @param dns_stub_listener_extra
 #   Additional addresses for the DNS stub listener to listen on
+#   Setting it to `'absent'` will remove `DNSStubListenerExtra` existing entries from the configuration file
 #
 # @param manage_resolv_conf
 #   For when `manage_resolved` is `true` should the file `/etc/resolv.conf` be managed.
@@ -204,8 +206,8 @@ class systemd (
   Optional[Variant[Boolean,Enum['allow-downgrade']]]  $dnssec = undef,
   Variant[Boolean,Enum['yes', 'opportunistic', 'no']] $dnsovertls = false,
   Optional[Variant[Boolean,Enum['no-negative']]]      $cache = undef,
-  Optional[Variant[Boolean,Enum['udp','tcp']]]        $dns_stub_listener = undef,
-  Optional[Array[String[1]]] $dns_stub_listener_extra = undef,
+  Optional[Variant[Boolean,Enum['udp','tcp','absent']]]  $dns_stub_listener = undef,
+  Optional[Variant[Array[String[1]],Enum['absent']]] $dns_stub_listener_extra = undef,
   Boolean                                             $manage_resolv_conf = true,
   Boolean                                             $use_stub_resolver = false,
   Boolean                                             $manage_networkd = false,
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -294,17 +294,19 @@ Default value: `undef`
 
 ##### <a name="-systemd--dns_stub_listener"></a>`dns_stub_listener`
 
-Data type: `Optional[Variant[Boolean,Enum['udp','tcp']]]`
+Data type: `Optional[Variant[Boolean,Enum['udp','tcp','absent']]]`
 
 Takes a boolean argument or one of "udp" and "tcp".
+Setting it to `'absent'` will remove `DNSStubListener` existing entries from the configuration file
 
 Default value: `undef`
 
 ##### <a name="-systemd--dns_stub_listener_extra"></a>`dns_stub_listener_extra`
 
-Data type: `Optional[Array[String[1]]]`
+Data type: `Optional[Variant[Array[String[1]],Enum['absent']]]`
 
 Additional addresses for the DNS stub listener to listen on
+Setting it to `'absent'` will remove `DNSStubListenerExtra` existing entries from the configuration file
 
 Default value: `undef`
 
--- a/manifests/resolved.pp
+++ b/manifests/resolved.pp
@@ -57,9 +57,9 @@ class systemd::resolved (
   Optional[Variant[Boolean,Enum['resolve']]] $multicast_dns          = $systemd::multicast_dns,
   Optional[Variant[Boolean,Enum['allow-downgrade']]] $dnssec         = $systemd::dnssec,
   Optional[Variant[Boolean,Enum['yes', 'opportunistic', 'no']]] $dnsovertls = $systemd::dnsovertls,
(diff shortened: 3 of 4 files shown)


traylenator:
The fear is that since undef is the default this is now an action.

It's probably okay with this parameter. 

I always quite like yumrepo's explicit  `absent`.

TheMeier:
> The fear is that since undef is the default this is now an action.
> 
> It's probably okay with this parameter.
> 
> I always quite like yumrepo's explicit `absent`.

I see you point. I can't follow your comparison to yumrepo here. These are class parameters, yumrepo is more 
There is a couple of ways we could handle this:

* go with the path laid out in this MR and mark it breaking
* implement more logic with a custom function that could look something like this:
  ```
  $_dns_stub_listener = $dns_stub_listener ? {
    true    => 'yes',
    false   => 'no',
    default => $dns_stub_listener,
  }

  if ($_dns_stub_listener) {
    ini_setting { 'dns_stub_listener':
      ensure  => new_function($_dns_stub_listener),
      value   => $_dns_stub_listener,
      setting => 'DNSStubListener',
      section => 'Resolve',
      path    => '/etc/systemd/resolved.conf',
      notify  => Service['systemd-resolved'],
    }
  }
  ```
* embrace the breaking and do a major restructuring
  Remove all the 
  ```
  if $manage_* {
    contain systemd::*
  }
  ```
  and change alle the subclasses marked `@api private` to be used directly by the user.
  This would have 2 benefits, less params in `init.pp`, move many test out of `init_spec.rb`
  That would be a drastic change, though. But I think it is worth considering.

traylenator:
> and change alle the subclasses marked @api private to be used directly by the user.

having the interface all on init is good - hard to test X+Y if classes are independet.

>go with the path laid out in this MR and mark it breaking
> implement more logic with a custom function that could look something like this:

Not sure about the custom function. Isn't it just:

```puppet
if dns_stub_listener =~ NotUndef {
  ini_setting { 'dns_stub_listener':
    ensure  => 'present',
    value   => $_dns_stub_listener,
    setting => 'DNSStubListener',
    section => 'Resolve',
    path    => '/etc/systemd/resolved.conf',
    notify  => Service['systemd-resolved'],
  }
}
```

i.e default remains as do nothing and set if set to which ever ?

TheMeier:
> Not sure about the custom function. Isn't it just:
> 
> ```puppet
> if $dns_stub_listener =~ NotUndef {
>   ini_setting { '':
>     ensure  => 'present',
>     value   => $_dns_stub_listener,
>     setting => 'DNSStubListener',
>     section => 'Resolve',
>     path    => '/etc/systemd/resolved.conf',
>     notify  => Service['systemd-resolved'],
>   }
> }
> ```
> 
> i.e default remains as do nothing and set if set to which ever ?

But wasn't the whole point to find a way to set `ensure => 'absent'`. The idea was that one can set `$dns_stub_listener == 'absent'` and only in that case the ensure would be set to absent

TheMeier:
@traylenator I have changed the code for an alternative solution... please have another look

manifests/init.pp
```diff
@@ -204,7 +206,7 @@
   Optional[Variant[Boolean,Enum['allow-downgrade']]]  $dnssec = undef,
   Variant[Boolean,Enum['yes', 'opportunistic', 'no']] $dnsovertls = false,
   Optional[Variant[Boolean,Enum['no-negative']]]      $cache = undef,
-  Optional[Variant[Boolean,Enum['udp','tcp']]]        $dns_stub_listener = undef,
+  Optional[Variant[Boolean,Enum['udp','tcp','absent']]]        $dns_stub_listener = undef,
```
bastelfreak, line 209:
```suggestion
  Optional[Variant[Boolean,Enum['udp','tcp','absent']]]  $dns_stub_listener = undef,
```

manifests/init.pp
```diff
@@ -65,9 +65,11 @@
 #
 # @param dns_stub_listener
 #   Takes a boolean argument or one of "udp" and "tcp".
+#   Setting it to `'absent'` will remove `DNSStubListener` existing entries from the configuration file-
 #
 # @param dns_stub_listener_extra
 #   Additional addresses for the DNS stub listener to listen on
+#   Setting it to `['absent']` will remove `DNSStubListenerExtra` existing entries from the configuration file-
```
bastelfreak, line 72: The `[]` look a bit odd.

jstraw: Maybe `Variant[Array[String[1],Enum['absent']]` ?

TheMeier: well the param is Array only at the moment. Should I add `Enum['variant']` to the variants?

ekohl:
The commit message and PR description no longer match the actual code. Would you mind fixing that up to avoid confusion in the future?

manifests/init.pp
```diff
@@ -204,8 +206,8 @@
   Optional[Variant[Boolean,Enum['allow-downgrade']]]  $dnssec = undef,
   Variant[Boolean,Enum['yes', 'opportunistic', 'no']] $dnsovertls = false,
   Optional[Variant[Boolean,Enum['no-negative']]]      $cache = undef,
-  Optional[Variant[Boolean,Enum['udp','tcp']]]        $dns_stub_listener = undef,
-  Optional[Array[String[1]]] $dns_stub_listener_extra = undef,
+  Optional[Variant[Boolean,Enum['udp','tcp','absent']]]  $dns_stub_listener = undef,
```
ekohl, line 209:
```suggestion
  Optional[Variant[Boolean,Enum['udp','tcp','absent']]] $dns_stub_listener = undef,
```

manifests/init.pp
```diff
@@ -204,8 +206,8 @@
...
+  Optional[Variant[Array[String[1]],Enum['absent']]] $dns_stub_listener_extra = undef,
```
ekohl, line 210: For a bit more readable type you can consider this:
```suggestion
  Variant[Undef, Array[String[1]], Enum['absent']] $dns_stub_listener_extra = undef,
```

ekohl (approved):
Small nits, but overall :+1:

PR #433, opened 2024-03-11
Make service restart upon unit file change optional

### LINKED ISSUE #432: Provide possibility to disable service restart upon unit file change

- Puppet: any
- Ruby: aby
- Distribution: any
- Module version: any

Setup a unit file with systemd::manage_unit or directly with systemd::unit_file. Have some of the contents change.

## What are you seeing

puppet-systemd will restart the service no matter what.

I'd expect to have the possibility to disable puppet-systemd restart the service(s) in question when the unit file changes (while still executing daemon-reload so systemd has the most recent version of the unit file loaded).

This is needed especially when touching multiple services per server across multiple servers. The on-disk unit files as well as systemd's view need to be updated, but no further action should be taken as this must be done outside of Puppet.

schustersv:

Provide the possibility to disable service restart upon unit file change

Fixes #432

### DIFF
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -58,6 +58,9 @@
 # @param daemon_reload
 #   call `systemd::daemon-reload` to ensure that the modified unit file is loaded
 #
+# @param service_restart
+#   restart (notify) the service when unit file changed
+#
 # @example manage unit file + service
 #   systemd::unit_file { 'foo.service':
 #     content => file("${module_name}/foo.service"),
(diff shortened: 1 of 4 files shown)


manifests/unit_file.pp
```diff
@@ -80,7 +83,8 @@
   Optional[String]                         $restart   = undef,
   Boolean                                  $selinux_ignore_defaults = false,
   Hash[String[1], Any]                     $service_parameters = {},
-  Boolean                                  $daemon_reload = true
+  Boolean                                  $daemon_reload = true,
+  Boolean                                  $service_restart = true
```
kenyon, line 87: Let's add the trailing comma so that we don't keep having unnecessary diffs.

```suggestion
  Boolean                                  $service_restart = true,
```

kenyon (approved):

manifests/unit_file.pp
schustersv: Ah right, should've done that in the first place as it's already this way in manage_unit.pp

PR #434, opened 2024-03-11
Don't allow ensure=file anymore for systemd::unit_file

baurmatt:
@cocker-cc correctly mentioned in
https://github.com/voxpupuli/puppet-systemd/commit/7f63dfa49850a860b24a93874fe5c5f4430f9edc#commitcomment-139614892 that after #405 `ensure=file` breaks due to `stdlib::ensure` not being able to handle that. As `file` is any way only a synonym for `present`, let's just remove it from the possible values.

https://github.com/voxpupuli/puppet-systemd/commit/7f63dfa49850a860b24a93874fe5c5f4430f9edc#commitcomment-139614892

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -2073,7 +2073,7 @@ The target unit file to create
 
 ##### <a name="-systemd--unit_file--ensure"></a>`ensure`
 
-Data type: `Enum['present', 'absent', 'file']`
+Data type: `Enum['present', 'absent']`
 
 The state of the unit file to ensure
 
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -66,7 +66,7 @@
 #   }
 #
 define systemd::unit_file (
-  Enum['present', 'absent', 'file']        $ensure    = 'present',
+  Enum['present', 'absent']                $ensure    = 'present',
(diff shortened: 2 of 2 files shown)


kenyon (approved):
REFERENCE.md is outdated, but otherwise LGTM. Sad that we missed this in #405, I guess there isn't test coverage.

kenyon:
Might also be better to preserve backwards compatibility by just treating `$ensure=file` the same as `$ensure=present`.

baurmatt:
@kenyon Thanks for the review! I've update `REFERENCE.md`.

Regarding the backwards compatibility: I thought a lot about it. But i feels like a bad UX to have `ensure` allow `file`, but not `link` - which is automatically set if `target` is defined. Thus I prefer to remove the `file` option.

I've also git blamed it back to the source -> https://github.com/voxpupuli/puppet-systemd/blame/b80984882b79c4edb48486513be1436dfa6736f2/manifests/unit_file.pp
It seems it was `file` back than because `systemd::unit_file` was only a simple wrapper around a `file` resource but as of now, it's way more.

kenyon:
@baurmatt I agree with your analysis, this is a good change.

TheMeier (approved):

PR #436, opened 2024-03-12
Create manage_unit, manage_dropin types from hiera

### LINKED ISSUE #435: Add hiera-friendly option to manage dropins
As per recently closed #420 we have a nice option to manage user slices etc, but the only way to make real use of it is via different module.
I would like to use this capability via Hiera - i.e. the way I'd expect it to work is something like this:

```yaml
systemd::dropin_files:
  'user-limits.conf':
    unit: 'user-.slice'
    slice_entry:
        'MemoryMax': '17G'
```

but this obviously does not work as `systemd::dropin_files` does not load `systemd::manage_dropin` class and even if it did, the class does not create resources from hiera.

This is a friendly request improve the code to make it overall more hiera-friendly.

traylenator:

Units and dropins can now be created from hiera records:

e.g.

```yaml
systemd::manage_dropins:
  myconf.conf:
    ensure: present
    unit: myservice.service
    service_entry:
      Type: oneshot
      ExecStart:
        - ''
        - '/usr/bin/doit.sh'

systemd::manage_units:
  myservice.service:
    unit_entry:
      Description: My Customisation
    service_entry:
      CPUWeight: 2000
```
Fixes #435

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -149,7 +149,13 @@
 #   `loginctl_user`.
 #
 # @param dropin_files
-#   Configure dropin files via hiera with factory pattern
+#   Configure dropin files via hiera and `systemd::dropin_file` with factory pattern
+#
+# @param manage_units
+#   Configure units via hiera and `systemd::manage_unit` with factory pattern
+#
+# @param manage_dropins
+#   Configure dropin files via hiera and `systemd::manage_dropin` with factory pattern
 #
 # @param manage_all_network_files
 #
@@ -234,6 +240,8 @@ class systemd (
   Stdlib::Absolutepath                                $network_path = '/etc/systemd/network',
   Hash                                                $loginctl_users = {},
   Hash                                                $dropin_files = {},
+  Hash[String[1], Hash[String[1], Any]]               $manage_units = {},
+  Hash[String[1], Hash[String[1], Any]]               $manage_dropins = {},
   Hash                                                $udev_rules = {},
   Boolean                                             $manage_coredump = false,
   Systemd::CoredumpSettings                           $coredump_settings = {},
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -234,6 +240,8 @@
   Stdlib::Absolutepath                                $network_path = '/etc/systemd/network',
   Hash                                                $loginctl_users = {},
   Hash                                                $dropin_files = {},
+  Hash                                                $manage_units = {},
+  Hash                                                $manage_dropins = {},
```
ekohl, line 244: Others don't have it either, but shouldn't this be narrowed down to `Hash[String[1], Hash]`?

ekohl (approved):

kenyon (approved):

manifests/init.pp
ekohl: Thinking more about it: almost feels like a `Stdlib::CreateResources` type would make sense.

traylenator: Also crossed my mind.

We'd have to bump minimum stdlib from 4.13.1 to 8.5.0 https://github.com/puppetlabs/puppetlabs-stdlib/commit/e8ba623bbd7facbad726e9c56b39a8c5341c16c8

Suggest follow up with a backward-incompatible MR that changes the 3(?) of them.

TheMeier: Though I really wonder who would such an old stdlib, is this really breaking?

TheMeier (approved):

manifests/init.pp
traylenator: Change the two new ones.
Will follow up with the rest and it can sit there for rainy day.

ekohl: > We'd have to bump minimum stdlib from 4.13.1 to 8.5.0 [puppetlabs/puppetlabs-stdlib@e8ba623](https://github.com/puppetlabs/puppetlabs-stdlib/commit/e8ba623bbd7facbad726e9c56b39a8c5341c16c8)

I'll admit: I completely forgot I already wrote that and was just theorizing "it would be nice", but then thought "it'll be a while before everyone has it". Looks like I already did the hard parts long enough ago.

ekohl (approved):

PR #437, opened 2024-03-12
Deprecate `systemd::service_limits`

traylenator:

The `systemd::service_limits` and corresponding type `Systemd::Servicelimits` is deprecated.

Switch to `systemd::dropin_file` for a source attributed `systemd::service_limits` or `systemd::manage_unit` for a `limits` attributed `systemd::service_limits`.


manifests/service_limits.pp
```diff
@@ -1,4 +1,4 @@
-# Adds a set of custom limits to the service
+# Deprecate - Adds a set of custom limits to the service
```
kenyon, line 1:
```suggestion
# Deprecated - Adds a set of custom limits to the service
```

manifests/service_limits.pp
```diff
@@ -33,7 +33,7 @@
   Enum['present', 'absent', 'file'] $ensure                  = 'present',
   Stdlib::Absolutepath              $path                    = '/etc/systemd/system',
   Boolean                           $selinux_ignore_defaults = false,
-  Optional[Systemd::ServiceLimits]  $limits                  = undef,
+  Optional[Systemd::Servicelimits]  $limits                  = undef,
```
kenyon, line 36: I think these are case insensitive, but is there a reason to change it? Seems like it'd be better to leave it matching the case of the type definition.

manifests/service_limits.pp
```diff
@@ -63,14 +63,44 @@
...
+
+    $_my_limits = $limits.map | $_directive, $_value | {
+      if $_directive in $_now_tupled {
+        { $_directive => $_value.map | $_pair | { Array($_pair).flatten } }  # Convert { 'a' => 'b' } to ['a', b']
+      } else {
+        { $_directive => $_value }
+      }
+    }.reduce | $_memo, $_value | { $_memo + $_value }
+
+    deprecation("systemd::servicelimits - ${title}",'systemd::servicelimits is deprecated, user systemd::manage_dropin')
```
kenyon, line 84:
```suggestion
    deprecation("systemd::servicelimits - ${title}",'systemd::servicelimits is deprecated, use systemd::manage_dropin')
```

manifests/service_limits.pp
```diff
@@ -63,14 +63,44 @@
...
+      ensure                  => $ensure,
+      unit                    => $name,
+      filename                => '90-limits.conf',
+      path                    => $path,
+      selinux_ignore_defaults => $selinux_ignore_defaults,
+      service_entry           => $_my_limits,
+      notify_service          => true,
+    }
+  } else {
+    deprecation("systemd::servicelimits ${title}",'systemd::servicelimits is deprecated, user systemd::dropin_file or systemd::manage_dropin')
```
kenyon, line 95:
```suggestion
    deprecation("systemd::servicelimits ${title}",'systemd::servicelimits is deprecated, use systemd::dropin_file or systemd::manage_dropin')
```

README.md
```diff
@@ -586,6 +586,11 @@ Finished on controller-0:
   }
 ```
 
+## Deprecate Notices
```
kenyon, line 589:
```suggestion
## Deprecation Notices
```

manifests/service_limits.pp
traylenator: Yep - I removed it originally and tried using Systemd::Unit::Service there.

traylenator:
I thought that should fail on 8- was confused why it didn't locall - sure it used to.

README.md
```diff
@@ -586,6 +586,11 @@ Finished on controller-0:
   }
 ```
 
+## Deprecation Notices
+
+The type `systemd::service_limits` is deprecated and `systemd::manage_dropin` or `systemd::dropin_file` can
```
TheMeier, line 591:
```suggestion
The type `systemd::service_limits` is deprecated and `systemd::manage_dropin` or `systemd::dropin_file` should
```

TheMeier (approved):

bastelfreak:
@traylenator is that actually a breaking change, or does it just prepare it? I would like to do the next major release somewhere in the next weeks and drop also support for EoL/soon-EoL operating systems.

traylenator:
> @traylenator is that actually a breaking change, or does it just prepare it? I would like to do the next major release somewhere in the next weeks and drop also support for EoL/soon-EoL operating systems.

It just prepares it - should be nothing breaking other than puppet8 rspec tests of course.

Had assumed that breaking is when you add the deprecation and then when its actually removed it is not breaking since declared deprecated?

bastelfreak:
I would say it doesn't break your existing setup, so this could be released as a minor release and when we remove the parameter it's a breaking change.

traylenator:
There's already a breaking change in HEAD however - https://github.com/voxpupuli/puppet-systemd/pull/438 ? Ship has sailed.

TheMeier:
I would be nice if we could get #406, #429, #437 and #443 done and make a major release, plus the proposed drop for EOL/soon-EOL from @bastelfreak? And hope @traylenator doesn't come up with another breaking feature with 1-2 weeks :D

TheMeier:
There is also #434 which is already merged

traylenator:
> There is also #434 which is already merged
> 

Yeah. I've had to change (bad) code due to this one.

PR #438, opened 2024-03-12
Use Stdlib::CreateResources type for hiera expansions

traylenator:

For a number of paramaters which generate resources typically from hiera files use `Stdlib::CreateResources` type.

This requires a newer puppetlabs-stdlib module.


kenyon (changes requested):
Also these ones:

https://github.com/voxpupuli/puppet-systemd/blob/63a41e024046862d739422d5e6831a96a795d4ca/manifests/init.pp#L197-L201

kenyon (approved):

ekohl (approved):

smortex (approved):

PR #443, opened 2024-03-29
Support reload of instances of systemd --user

traylenator:

This change increased the minimum required Puppet version to 6.24.0 or 7.9.0 [PUP-5704](https://puppet.atlassian.net/browse/PUP-5704) to support arrays to the command attribute of the exec type.

Support the calling the `systemd --user daemon-reload` for a particular user.

Example run:
```puppet
notify{'junk':
 notify => Systemd::Daemon_reload['user_foobar'],
}

systemd::daemon_reload{'user_steve':
  user => 'steve',
}
```

This results on a Fedora box:

```
Notice: /Stage[main]/Main/Notify[junk]/message: defined 'message' as 'junk'
Notice: /Stage[main]/Main/Systemd::Daemon_reload[user_steve]/Package[systemd-container]: Triggered 'refresh' from 1 event
Notice: /Stage[main]/Main/Systemd::Daemon_reload[user_steve]/Exec[systemd-user_steve-systemctl-user-steve-daemon-reload]: Triggered 'refresh' from 1 event
```
and a journal (debug on) for `user@1000.service` of

```
Mar 29 10:32:11 fedora systemd[2062]: Created slice background.slice - User Background Tasks Slice.
Mar 29 10:32:11 fedora systemd[2062]: Starting systemd-tmpfiles-clean.service - Cleanup of User's Temporary Files and Directories...
Mar 29 10:32:11 fedora systemd[2062]: Reloading requested from client PID 4379 ('systemctl')...
Mar 29 10:32:11 fedora systemd[2062]: Reloading...
Mar 29 10:32:11 fedora systemd[2062]: Reloading finished in 179 ms.
Mar 29 10:32:11 fedora systemd[2062]: Finished systemd-tmpfiles-clean.service - Cleanup of User's Temporary Files and Directories.
```

Only recent versions of `systemd-run` support the `--machine` option or actually manage to connect to DBUS.

### DIFF
--- a/manifests/daemon_reload.pp
+++ b/manifests/daemon_reload.pp
@@ -7,15 +7,49 @@
 #
 # @param enable
 #   Enable the reload exec
-#
 #   * Added in case users want to disable the reload globally using a resource collector
+#
+# @param user
+#   Specify user name of `systemd --user` to reload. This not supported **below** Redhat 9,
+#   Ubuntu 22.04 or Debian 12.
+#
+# @example Force reload the system systemd
+#   notify{ 'fake event to notify from':
+#     notify => Systemd::Daemon_reload['special']
+#   }
+#   systemd::daemon_reload{ 'special': }
+#
+# @example Force reload a systemd --user
+#   notify{ 'fake event to notify from':
+#     notify => Systemd::Daemon_reload['steve_user']
+#   }
+#   systemd::daemon_reload{ 'steve_user':
+#     user => 'steve',
+#   }
 #
 define systemd::daemon_reload (
   Boolean $enable = true,
+  Optional[String[1]] $user = undef,
 ) {
+  include systemd
+
   if $enable {
-    exec { "${module_name}-${name}-systemctl-daemon-reload":
-      command     => 'systemctl daemon-reload',
+    if $user {
+      if ($facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'],'9') < 0 ) or
+      ($facts['os']['name'] == 'Debian' and versioncmp($facts['os']['release']['major'],'12') < 0 ) or
+      ($facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['major'],'22.04') < 0 ) {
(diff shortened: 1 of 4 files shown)


manifests/daemon_reload.pp
```diff
@@ -7,15 +7,55 @@
...
+      ($facts['os']['name'] == 'Debian' and versioncmp($facts['os']['release']['major'],'12') < 0 ) or
+      ($facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['major'],'22.04') < 0 ) {
+        fail('systemd::daemon_reload_for a user is not supported below RedHat 9, Debian 12 or Ubuntu 22.04')
+      }
+
+      $_title   = "${module_name}-${name}-systemctl-user-${user}-daemon-reload"
+      # Must use full path to systemctl here.
+      $_command = ['machinectl', 'shell' , "${user}@.host", '/usr/bin/systemctl', '--user', 'daemon-reload']
+
+      if $systemd::machinectl_package {
```
bastelfreak, line 48: does it make sense to have an explicit parameter that configures the package installation? People might manage it already on other places in the catalog. That would also allow us to remove the `stdlib::ensure_packages()` call wich isn't 100% reliable anyways.

bastelfreak:
@traylenator I think the bump to Puppet >= 7.9 is missing? (And I'm still torn if that's a breaking change or not).

manifests/daemon_reload.pp
traylenator: I considered it for sure - I decided as it had no apparent configuration you could make for the package it was not needed. Your point is valid that ensure_packages can go wrong. So a 

* `systemd::manage_containers`
* `systemd::containers_package`

and a new class systemd::containers ?

? since that's the package name.

Prefer that?

bastelfreak: I think that's a good way forward.

traylenator:
```
systemd-run --pipe --wait --user --machine straylen@.host systemctl --user daemon-reload
```
is a better command since the return code makes it out.

TheMeier (approved):

PR #444, opened 2024-03-29
New parameters to manage systemd-nspawn

traylenator:

New parameters `systemd::manage_nspawn` defaulting to `false`, if `true` it will ensure that the `machinectl` and `systemd-nspawn` commands are available.

When `true` the systemd_internal_services fact will be populated.

```
facter -p systemd_internal_services
{
  systemd-nspawn@.service => "disabled",
}
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -31,6 +31,13 @@
 # @param resolved_package
 #   The name of a systemd sub package needed for systemd-resolved if one needs to be installed.
 #
+# @param manage_nspawn
+#   Manage the systemd-nspawn@service and machinectl subsystem.
+#
+# @param nspawn_package
+#   The name of a systemd sub package needed for the nspawn tools machinectl and
+#   systemd-nspawn if one needs to be installed.
+#
 # @param dns
 #   A space-separated list of IPv4 and IPv6 addresses to use as system DNS servers.
 #   DNS requests are sent to one of the listed DNS servers in parallel to suitable
@@ -243,6 +250,8 @@ class systemd (
   Stdlib::CreateResources                             $manage_dropins = {},
   Stdlib::CreateResources                             $udev_rules = {},
   Boolean                                             $manage_coredump = false,
+  Boolean                                             $manage_nspawn = false,
+  Optional[Enum['systemd-container']]                 $nspawn_package = undef,
   Systemd::CoredumpSettings                           $coredump_settings = {},
   Boolean                                             $coredump_backtrace = false,
   Boolean                                             $manage_oomd = false,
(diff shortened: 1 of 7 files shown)


manifests/init.pp
```diff
@@ -243,6 +250,8 @@
   Stdlib::CreateResources                             $manage_dropins = {},
   Stdlib::CreateResources                             $udev_rules = {},
   Boolean                                             $manage_coredump = false,
+  Boolean                                             $manage_nspawn = false,
+  Optional[Enum['systemd-container']]                 $nspawn_package = undef,
```
ekohl, line 254: Does it matter which string this is? If you go this route, why not a boolean?

spec/classes/init_spec.rb
```diff
@@ -279,6 +280,28 @@
...
+            when '7'
+              it { is_expected.not_to contain_package('systemd-container') } # rubocop:disable RSpec/RepeatedExample
+            else
+              it { is_expected.to contain_package('systemd-container').with_ensure('present') } # rubocop:disable RSpec/RepeatedExample
+            end
+          when 'Debian'
+            it { is_expected.to contain_package('systemd-container').with_ensure('present') } # rubocop:disable RSpec/RepeatedExample
+          else
+            it { is_expected.not_to contain_package('systemd-container') } # rubocop:disable RSpec/RepeatedExample
+          end
```
ekohl, line 302: Is this better?
```suggestion
          # rubocop:disable RSpec/RepeatedExample
          case facts[:os]['family']
          when 'RedHat'
            case facts[:os]['release']['major']
            when '7'
              it { is_expected.not_to contain_package('systemd-container') }
            else
              it { is_expected.to contain_package('systemd-container').with_ensure('present') }
            end
          when 'Debian'
            it { is_expected.to contain_package('systemd-container').with_ensure('present') }
          else
            it { is_expected.not_to contain_package('systemd-container') }
          end
          # rubocop:enable RSpec/RepeatedExample
```

ekohl (commented):
I wonder if it makes sense to implement a generic `RedHat-family.yaml` and then a `data/RedHat-family-7.yaml` with
```yaml
systemd::nspawn_package: %
```

manifests/init.pp
traylenator: This is consistent with the systemd-resolved package Enum.

Presumably the thought that random OS in the future would have a new string and it needs recording somewhere.

spec/classes/init_spec.rb
traylenator: I'd say the precise line comments was clearer as to where the "offence" is, more obvious what to do when EL7 dies.

ekohl (approved):

traylenator:
> I wonder if it makes sense to implement a generic `RedHat-family.yaml` and then a `data/RedHat-family-7.yaml` with
> ```yaml
> systemd::nspawn_package: %
> ```

Yeah. I'll do that, am a big fan of making the future default.

ekohl:
Sorry, I meant `~` but that's what you get for quickly trying to review something.

data/Fedora.yaml
```diff
@@ -1,4 +1,5 @@
 ---
+systemd::nspawn_package: 'systemd-container'
```
ekohl, line 2: Isn't this now caught by the RH family?

PR #445, opened 2024-04-03
 Add options to manage file system access options to Service

carthik:
####  Add options to manage file system access options to Service
Added in version 231, systemd provides options to manage file system access options to processes executed by systemd services. Adding these as valid options.

See https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#ReadWritePaths= for docs.

n/a


REFERENCE.md
```diff
@@ -2753,6 +2753,11 @@ Struct[{
     Optional['ProtectHome']               => Variant[Boolean, Enum['read-only', 'tmpfs']],
     Optional['BindPaths']                 => Variant[Stdlib::Unixpath,Pattern[/-\/.+/], Array[Variant[Stdlib::Unixpath,Pattern[/-\/.+/]],1]],
     Optional['BindReadOnlyPaths']         => Variant[Stdlib::Unixpath,Pattern[/-\/.+/], Array[Variant[Stdlib::Unixpath,Pattern[/-\/.+/]],1]],
+    Optional['ReadWritePaths']            => Variant[Stdlib::Unixpath,Pattern[/-\/.+/], Array[Variant[Stdlib::Unixpath,Pattern[/-\/.+/]],1]],
```
traylenator, line 2756:
```suggestion
    Optional['ReadWritePaths']            => Variant[Stdlib::Unixpath,Pattern[/^-\/.+/], Array[Variant[Stdlib::Unixpath,Pattern[/^-\/.+/]],1]],
```

and similar. The `-` is always at the start.

traylenator:
Could you add a few examples and counter examples to - https://github.com/voxpupuli/puppet-systemd/blob/master/spec/type_aliases/systemd_unit_service_spec.rb

REFERENCE.md
carthik: Thanks! 
I updated the regular expression to accommodate this (from the systemd docs): 
```
Paths in ReadWritePaths=, ReadOnlyPaths=, InaccessiblePaths=, ExecPaths= and NoExecPaths= may be prefixed with "-", in which case they will be ignored when they do not exist. If prefixed with "+" the paths are taken relative to the root directory of the unit, as configured with RootDirectory=/RootImage=, instead of relative to the root directory of the host (see above). When combining "-" and "+" on the same path make sure to specify "-" first, and "+" second.
```

Tested thus: https://rubular.com/r/oATb23JSEuE0MX

tavin:
due to vendor hardening, `logrotate` is now broken on our prod servers, and we can't use `systemd::manage_dropin` to fix it, until `ReadWritePaths` is added

thanks for your efforts

bastelfreak:
@tavin this is opensource. You can always jump in and pick up the PR or pay someone to implement it for you.

bastelfreak:
This was merged in https://github.com/voxpupuli/puppet-systemd/pull/545

PR #447, opened 2024-04-08
Release 7.0.0

traylenator:
Release 7.0.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,47 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
+## [v7.0.0](https://github.com/voxpupuli/puppet-systemd/tree/v7.0.0) (2024-04-08)
+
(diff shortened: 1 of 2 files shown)


traylenator:
Think #445 is only close one.

traylenator:
> #433 and #432 are duplicate in the changelog
Removed one of the enhancement tags  which is is I presume the correct thing to do.

TheMeier:
Hm I guess you also need the skip label on the issue #432

bastelfreak:
the systemd module is a dependency to many others. Does it make sense to already drop EL7 support as well? That will save us one breaking change that'e due in a few weeks.

bastelfreak:
And Debian 10 is really already dead, so please let us include https://github.com/voxpupuli/puppet-systemd/pull/448

ekohl (changes requested):
I think https://github.com/voxpupuli/puppet-systemd/pull/449 should be merged before we release this.

TheMeier:
#449 is merged, but changelog needs to be re-generated

traylenator:
Obsoleted by #456

PR #452, opened 2024-04-11
remove `systemd::escape` usage for `timer_wrapper`

### LINKED ISSUE #451: Usage of `systemd::escape` in `systemd::timer_wrapper` creates weird names

- Puppet: 7.29.1
- Ruby: 2.7.8p225
- Distribution: Ubuntu 20.04.6
- Module version: 6.6.0

```puppet
  systemd::timer_wrapper { 'prometheus-node-exporter-deleted_libraries-custom':
    ensure                 => present,
    command                => '/bin/sh -c "/usr/share/prometheus-node-exporter-collec
tors/deleted_libraries.py | sponge /var/lib/prometheus/node-exporter/lvm.prom"',
    on_boot_sec            => 0,
    on_unit_active_sec     => '15min',
    timer_unit_overrides   => {
      'Description' => 'Run deleted libraries metrics collection every 15 minutes',
    },
    service_unit_overrides => {
      'Description' => 'Collect deleted libraries metrics for prometheus-node-exporte
r',
    },
    service_overrides      => {
      'Environment' => 'TMPDIR=/var/lib/prometheus/node-exporter',
    },
  }
```

## What are you seeing

Due to the usage of `systemd::escape` in https://github.com/voxpupuli/puppet-systemd/blob/45e09532b9dc392fe767b555813945c94113422c/manifests/timer_wrapper.pp#L78 a `-` will be escaped and becomes `\x2d`.

```
● prometheus\x2dnode\x2dexporter\x2ddeleted_libraries\x2dcustom.timer - Run deleted libraries metrics collection every 15 minutes
     Loaded: loaded (/etc/systemd/system/prometheus\x2dnode\x2dexporter\x2ddeleted_libraries\x2dcustom.timer; enabled; vendor preset: enabled)
     Active: active (waiting) since Wed 2024-04-10 15:11:05 CEST; 17min ago
    Trigger: Wed 2024-04-10 15:41:06 CEST; 12min left
   Triggers: ● prometheus\x2dnode\x2dexporter\x2ddeleted_libraries\x2dcustom.service

Apr 10 15:11:05 host systemd[1]: Started Run deleted libraries metrics collection every 15 minutes.
```

Timer and service are named `prometheus-node-exporter-deleted_libraries-custom` not `prometheus\\x2dnode\\x2dexporter\\x2ddeleted_libraries\\x2dcustom`

Link to the discussion on the usage of `systemd::escape` https://github.com/voxpupuli/puppet-systemd/pull/419#discussion_r1510333656

TheMeier:
Don't treat unit names with `systemd::escape`. If you have `systemd::timer_wrapper` instances with names that contain special characters like slashes and umlauts this will result in a validation error.  In such cases you can produce the previous  behaviour by using `systemd::escape yourself` for example:

```puppet
systemd::timer_wrapper {systemd::escape( 'my_timer'):
  ensure        => 'present',
  command       => '/usr/bin/echo "Hello World"',
  on_calendar   => '*:0/5',
}
```

It is possible that this change will create duplicate resources in case the resource name was escaped but the unescaped name is also valid. In those cases make sure to remove the duplicate, e.g. like so:
```puppet
systemd::timer_wrapper { systemd::escape("someTitle"):
  ensure => absent,
}
```

Fixes #451

### DIFF
--- a/manifests/timer_wrapper.pp
+++ b/manifests/timer_wrapper.pp
@@ -75,9 +75,8 @@ define systemd::timer_wrapper (
   }
 
   $service_ensure = $ensure ? { 'absent' => false,  default  => true, }
-  $unit_name = systemd::escape($title)
 
-  systemd::manage_unit { "${unit_name}.service":
+  systemd::manage_unit { "${title}.service":
     ensure        => $ensure,
     unit_entry    => $service_unit_overrides,
     service_entry => {
@@ -86,7 +85,7 @@ define systemd::timer_wrapper (
       'Type'      => 'oneshot',
     }.filter |$key, $val| { $val =~ NotUndef } + $service_overrides,
   }
-  systemd::manage_unit { "${unit_name}.timer":
+  systemd::manage_unit { "${title}.timer":
     ensure        => $ensure,
     unit_entry    => $timer_unit_overrides,
     timer_entry   => $_timer_spec +  $timer_overrides,
@@ -95,19 +94,19 @@ define systemd::timer_wrapper (
     },
   }
 
-  service { "${unit_name}.timer":
+  service { "${title}.timer":
     ensure => $service_ensure,
     enable => $service_ensure,
   }
 
   if $ensure == 'present' {
-    Systemd::Manage_unit["${unit_name}.service"]
-    -> Systemd::Manage_unit["${unit_name}.timer"]
-    -> Service["${unit_name}.timer"]
+    Systemd::Manage_unit["${title}.service"]
+    -> Systemd::Manage_unit["${title}.timer"]
+    -> Service["${title}.timer"]
   } else {
     # Ensure the timer is stopped and disabled before the service
-    Service["${unit_name}.timer"]
-    -> Systemd::Manage_unit["${unit_name}.timer"]
-    -> Systemd::Manage_unit["${unit_name}.service"]
+    Service["${title}.timer"]
+    -> Systemd::Manage_unit["${title}.timer"]
+    -> Systemd::Manage_unit["${title}.service"]
   }
 }
--- a/spec/defines/timer_wrapper_spec.rb
+++ b/spec/defines/timer_wrapper_spec.rb
@@ -73,18 +73,7 @@ describe 'systemd::timer_wrapper' do
           end
 
           it {
-            is_expected.to compile
-            is_expected.to contain_file('/etc/systemd/system/t-i-t-l-e.timer')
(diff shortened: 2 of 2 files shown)


TheMeier:
Question is, is this considered breaking?

bastelfreak:
mhm this is probably a breaking change to some users. Should we remove it or should we document that some characters can cause strange unit names?

saz:
I'd still vote for getting rid of the `systemd::escape` usage, as it's completely unexpected behavior.

Another possible way: add a new parameter to force a specific name without this behavior?

TheMeier:
The 7.0.0 major release is still pending https://github.com/voxpupuli/puppet-systemd/pull/447 so we could get it in there ....
I am also not really happy with the escape

smortex:
I think this make sense. As a backwards-incompatible change we should indicate explicitly how users can check if they are affected and what they need to do in the first comment of the PR

webcompas:
> I'd still vote for getting rid of the `systemd::escape` usage, as it's completely unexpected behavior.

I'd also agree that there shouldn't be any automatic escaping. It's very confusing and it makes it hard for other tools which rely on predictable unit names.

saz:
If we're cleaning up unit files created with the escaped name and adding a new one, will this still count as a breaking change?
Somebody might depend on `systemd::manage_unit` directly, but I'd rather not expect that, as you will stumble upon the unexpected name.

Something like:
```
$unit_name_escaped = systemd::escape($title)
systemd::manage_unit { "${unit_name_escaped}.service":
  ensure => absent,
}
systemd::manage_unit { "${unit_name_escaped}.timer":
  ensure => absent,
}

systemd::manage_unit { "${title}.service":
...
}
systemd::manage_unit { "${title}.timer":
...
}
```

TheMeier:
As this feature is quite new, I don't expect a wide usage of it yet. I would like to avoid more complexity here and keep it as is. We have a major release with multiple breaking changes lined up anyway.

saz:
@TheMeier Maybe we should give an example on how to clean it up in the changelog?

Something like this should be working:
```
systemd::timer_wrapper { systemd::escape("someTitle"):
  ensure => absent,
}
```

TheMeier:
I have updated the MR description

traylenator (approved):
And much simpler - the point about expectation is super valid.

kenyon (approved):

TheMeier:
@traylenator can you include that in https://github.com/voxpupuli/puppet-systemd/pull/447?
I can also take care of that if you like.

Issue #377, opened 2023-10-04
timedatectl support


jcpunk:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 7.26.0
- Ruby: from AIO
- Distribution: CentOS Stream 9
- Module version: 6.0.0

## What behaviour did you expect instead

I'm wondering if it would make sense to add the `timedatectl set-timezone` and `timedatectl set-local-rtc` feature set to this module.

## Any additional information you'd like to impart

https://www.freedesktop.org/software/systemd/man/timedatectl.html

kenyon:
I would use it.

TheMeier:
The only sensible use cases would be `set-timezone` and `set-time` though. Setting `set-local-rtc ` to `1` is discouraged. And ntp should be managed with `systemd::timesyncd`

PR #462, opened 2024-05-22
Add support for timezone and hardware clock

jcpunk:
Add support for timedatectl to set timezone and hardware clock

Fixes: #377

### DIFF
--- a/manifests/timedatectl.pp
+++ b/manifests/timedatectl.pp
@@ -1,0 +1,46 @@
+# @api private
+#
+# @summary This class provides an abstract way to set elements with timedatectl
+#
+# @param set_local_rtc
+# Takes a boolean argument. If "false", the system is configured to maintain
+# the RTC in universal time.
+# If "true", it will maintain the RTC in local time instead.
+# Note that maintaining the RTC in the local timezone is not fully supported
+# and will create various problems with time zone changes and daylight saving
+# adjustments. If at all possible, keep the RTC in UTC mode.
+#
+# @param timezone
+# Set the system time zone to the specified value.
+# Available timezones can be listed with list-timezones.
+# If the RTC is configured to be in the local time, this will also update
+# the RTC time. This call will alter the /etc/localtime symlink.
+class systemd::timedatectl (
+  Optional[Boolean] $set_local_rtc = $systemd::set_local_rtc,
+  Optional[String[1]] $timezone = $systemd::timezone,
+) {
+  assert_private()
+
+  if $timezone {
+    exec { 'set system timezone':
+      command => "timedatectl set-timezone ${timezone}",
+      unless  => "test $(timedatectl show --property Timezone --value) = ${timezone}",
+      path    => $facts['path'],
+    }
+  }
+
+  # if $set_local_rtc is undef, this remains unmanaged
+  if $set_local_rtc {
+    exec { 'set local hardware clock to local time':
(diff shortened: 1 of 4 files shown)


jcpunk:
I almost went with

```puppet
  if $set_local_rtc {
    exec { 'set local hardware clock to local time':
      command => 'timedatectl set-local-rtc 1',
      onlyif  => 'test $(timedatectl show | grep LocalRTC | cut -d= -f2) = no',
      path    => $facts['path'],
    }
  } else {
    exec { 'set local hardware clock to UTC time':
      command => 'timedatectl set-local-rtc 0',
      unless  => 'test $(timedatectl show | grep LocalRTC | cut -d= -f2) = no',
      path    => $facts['path'],
    }
  }
```

But that make the hardware clock explicitly managed when it wasn't before.  I don't believe localtime is rational for POSIX system, but I'm not sure just pushing that on folks is the right call.

TheMeier:
I would have preferred `systemd::timedatectl` to be  not funneled through the main class, then the params could be non-optional. But the whole module is designed in that way, and others have different opinions on that ;)

So LGTM...

TheMeier (approved):

manifests/timedatectl.pp
```diff
@@ -0,0 +1,45 @@
...
+class systemd::timedatectl (
+  Optional[Boolean] $set_local_rtc = $systemd::set_local_rtc,
+  Optional[String[1]] $timezone = $systemd::timezone,
+) {
+  assert_private()
+
+  if $timezone {
+    exec { 'set system timezone':
+      command => "timedatectl set-timezone ${timezone}",
+      unless  => "test $(timedatectl show | grep Timezone | cut -d= -f2) = ${timezone}",
```
bastelfreak, line 27:
```suggestion
      unless  => "test $(timedatectl show --property Timezone --value) = ${timezone}",
```

jcpunk: Updated in all the snippets.

manifests/timedatectl.pp
```diff
@@ -0,0 +1,45 @@
...
+      path    => $facts['path'],
+    }
+  }
+
+  if $set_local_rtc == true {
```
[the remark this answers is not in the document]
jcpunk: Changed.

manifests/timedatectl.pp
```diff
@@ -0,0 +1,46 @@
+# @api private
+#
+# This class provides an abstract way to set elements with timedatectl
```
kenyon, line 3:
```suggestion
# @summary This class provides an abstract way to set elements with timedatectl
```

spec/classes/init_spec.rb
```diff
@@ -386,6 +386,43 @@
...
 
+        context 'when setting timezone' do
+          let(:params) do
+            {
+              timezone: 'America/Chicago',
+            }
+          end
+
+          it { is_expected.to compile.with_all_deps }
+          it { is_expected.to contain_exec('set system timezone') }
```
kenyon, line 397: Update to make use of the param:

```suggestion
          it { is_expected.to contain_exec('set system timezone').with_command('timedatectl set-timezone America/Chicago') }
```

jcpunk:
Changes merged.

kenyon (approved):
Reference needs a regen, otherwise 👍

jcpunk:
In theory this is sorted.

PR #468, opened 2024-06-06
Rework Red Hat resolved package list to be more future proof

jcpunk:
CentOS Stream10 continues to have systemd-resolved in its own package.

### DIFF
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -47,8 +47,7 @@ describe 'systemd' do
             it { is_expected.not_to contain_package('systemd-networkd') }
           end
 
-          case [facts[:os]['family'], facts[:os]['release']['major']]
-          when %w[RedHat 7], %w[RedHat 9], %w[Debian 12]
+          if (facts[:os]['family'] == 'RedHat' && facts[:os]['release']['major'] != '8') || (facts[:os]['name'] == 'Debian' && facts[:os]['release']['major'] == '12')
             it { is_expected.to contain_package('systemd-resolved') }
           else
             it { is_expected.not_to contain_package('systemd-resolved') }
--- a/data/RedHat-family-7.yaml
+++ b/data/RedHat-family-7.yaml
@@ -1,5 +1,4 @@
 ---
-systemd::resolved_package: 'systemd-resolved'
 systemd::nspawn_package: ~
 
 systemd::accounting:
--- a/data/RedHat-family-8.yaml
+++ b/data/RedHat-family-8.yaml
@@ -1,4 +1,6 @@
 ---
+systemd::resolved_package: ~
+
 systemd::accounting:
   DefaultCPUAccounting: 'yes'
   DefaultBlockIOAccounting: 'yes'
--- a/data/RedHat-family-9.yaml
+++ b/data/RedHat-family-9.yaml
@@ -1,5 +1,4 @@
 ---
-systemd::resolved_package: 'systemd-resolved'
 systemd::oomd_package: 'systemd-oomd'
 
 systemd::accounting:
--- a/data/RedHat-family.yaml
+++ b/data/RedHat-family.yaml
@@ -1,3 +1,4 @@
 ---
 systemd::networkd_package: systemd-networkd
 systemd::nspawn_package: 'systemd-container'
+systemd::resolved_package: 'systemd-resolved'


metadata.json
```diff
@@ -58,7 +58,8 @@
       "operatingsystemrelease": [
         "7",
         "8",
-        "9"
+        "9",
+        "10"
```
ekohl, line 62: Perhaps this is a good excuse to start supporting CentOS Stream 10 in our CI for acceptance tests, though I doubt there are Puppet packages available though :(

ekohl (changes requested):
I'd lean to leaving it out of `metadata.json` until we can verify it in CI.

Having said that, all the Red Hat family except EL 8 have `systemd-resolved`. Perhaps we should set it in `data/RedHat-family.yaml` and then set it to `~` in `data/RedHat-family-8.yaml` to unset it again. That way we're future proof.

At least, I see that on Fedora 39 & 40 systemd-resolved is a separate package as well that we don't have in `Fedora.yaml` right now.

metadata.json
jcpunk: I can confirm there is no puppet agent for CS10.  I pulled the one from Fedora to test it.

jcpunk:
Took a stab at moving the package name up a slot per your comment.

spec/classes/init_spec.rb
```diff
@@ -48,7 +48,7 @@
           end
 
           case [facts[:os]['family'], facts[:os]['release']['major']]
-          when %w[RedHat 7], %w[RedHat 9], %w[Debian 12]
+          when %w[RedHat 7], %w[RedHat 9], %w[Debian 12], %w[Fedora \d+]
```
evgeni, line 51: I think the `\d+` doesn't match. Did you try
```suggestion
          when %w[RedHat 7], %w[RedHat 9], %w[Debian 12], %w[Fedora /\d+/]
```

But even then, I think `%w[a b]` always produces strings, so you'd need to be explicit and say
```suggestion
          when %w[RedHat 7], %w[RedHat 9], %w[Debian 12], %w[Fedora 37],  %w[Fedora 38]
```

evgeni: Wait, lol, no!

`family` is `RedHat` for Fedora too. So how about

```suggestion
          when %w[RedHat 7], %w[RedHat 9..999], %w[Debian 12]
```

evgeni: Adding my brain dump from IRC:
```
19:19 < Zhenech> ooooh
19:20 < Zhenech> that major not mathing 9..999 is because it's a string, not an int
19:21 < Zhenech> mh, no, even that doesn't help
19:21 < Zhenech> fuckit
19:24 < Zhenech> okay, yeah, it constructs an Array, and that Array, as an object, must match, it can't contain matchers itself
```

spec/classes/init_spec.rb
```diff
@@ -47,8 +47,7 @@
             it { is_expected.not_to contain_package('systemd-networkd') }
           end
 
-          case [facts[:os]['family'], facts[:os]['release']['major']]
-          when %w[RedHat 7], %w[RedHat 9], %w[Debian 12]
+          if (facts[:os]['family'] == 'RedHat' && facts[:os]['release']['major'] != 7) || (facts[:os]['name'] == 'Debian' && facts[:os]['release']['major'] == '12')
```
evgeni, line 50:
```suggestion
          if (facts[:os]['family'] == 'RedHat' && facts[:os]['release']['major'] != '8') || (facts[:os]['name'] == 'Debian' && facts[:os]['release']['major'] == '12')
```

bastelfreak:
@jcpunk any chance you can add factsets to https://github.com/voxpupuli/facterdb as soon as CentOS offers vagrant images? That's required for us to properly test it.

jcpunk:
Stream 10 is still gathering shape, vagrant may be a ways off.

bastelfreak:
to quote from IRC:
> 15:38:29       Zhenech | https://mirror.stream.centos.org/10-stream/ is now a thing, so I assume containers should show up soon too and we can look into CI

ekohl:
https://quay.io/repository/centos/centos?tab=tags has `stream10-development` and `stream10-development-minimal`. But as I said earlier, we'll need Puppet packages. https://github.com/rpm-software-management/mock/pull/1383 is open, which is probably a requirement to easily build RPMs.

jcpunk:
https://kojihub.stream.centos.org/koji/packageinfo?packageID=3493 is also probably worth noting here.

ekohl:
Note our CI relies on Vagrant with Docker as a backend, so containers (which exist) are sufficient for that.

bastelfreak:
> Vagrant with Docker as a backend

Our CI uses docker directly, without vagrant (if you're speaking about the vox pupuli module pipeline). But we need Vagrant to generate factsets for facterdb (and that uses the facter ruby gems and doesn't require the puppet rpms).

ekohl:
> Our CI uses docker directly, without vagrant (if you're speaking about the vox pupuli module pipeline). But we need Vagrant to generate factsets for facterdb (and that uses the facter ruby gems and doesn't require the puppet rpms).

Oh yes, I indeed meant Beaker with Docker backend.

ekohl (approved):

evgeni (approved):

PR #473, opened 2024-07-04
RFC: add Kube unit support

dabelenda:
Podman integration brings support for multiple Unit types, including Kube unit that allows to start a pod with a YAML file declaration.

This PR in its current form is an RFC to discuss if these Unit Types are in the scope of this module. Or if they should be added to another module.

### DIFF
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -4,7 +4,7 @@
 #
 # @see systemd.unit(5)
 #
-# @param name [Pattern['^[^/]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope)$']]
+# @param name [Pattern['^[^/]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope|kube)$']]
 #   The target unit file to create
 #
 # @param ensure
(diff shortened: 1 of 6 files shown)


manifests/unit_file.pp
```diff
@@ -4,7 +4,7 @@
 #
 # @see systemd.unit(5)
 #
-# @param name [Pattern['^[^/]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope)$']]
+# @param name [Pattern['^[^/]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope|kube)$']]
```
bastelfreak, line 7: I think we shouldn't document it here, that's already in the type alias definition.

dabelenda:
Sorry I made a mistake, apparently there is a utility to run that will generate a service unit from the definition of the .kube  file.

systemd will not directly read kube units so it makes no sense to make this PR here...

traylenator:
I tried adding quadlet support https://github.com/southalc/podman/pull/83 which ,kube is an example of.

Having said give that MR seems to be going nowhere

PR #474, opened 2024-07-04
Add support for podman quadlets

dabelenda:
This PR adds support for podman Quadlets, a generator for systemd that creates units automatically for files in a specific folder.

Tested with a kube Yaml file and it starts the pod.

### DIFF
--- a/manifests/quadlet_file.pp
+++ b/manifests/quadlet_file.pp
@@ -1,0 +1,113 @@
+# @summary Creates a systemd Podman Quadlet file
+# Quadlet will generate a unit file, and this service can be managed by puppet.
+# @api public
+#
+# @see podman.systemd.unit(5)
+#
+# @param name
+#   The name of the quadlet file
+#
+# @param ensure
+#   The state of the quadlet file to ensure
+#
+# @param content
+#   The full content of the quadlet file
+#
+# @param path
+#   The path where the quadlet file will be created
+#   For systemd in user mode use any of
+#   - ~/.config/containers/systemd
+#   - /etc/containers/systemd/users/$(UID)
+#
+#   For global systemd use any of:
+#   - /etc/containers/systemd
+#   - /usr/share/containers/systemd
+#
+# @param source
+#   The ``File`` resource compatible ``source``
+#
+#   * Mutually exclusive with ``$content``
+#
+# @param owner
+#   The owner to set on the unit file
+#
+# @param group
+#   The group to set on the unit file
+#
+# @param mode
+#   The mode to set on the unit file
+#
+# @param enable
+#   If set, will manage the unit enablement status.
+#
+# @param active
+#   If set, will manage the state of the unit.
+#
+# @param restart
+#   Specify a restart command manually. If left unspecified, a standard Puppet service restart happens.
+#
+# @param service_parameters
+#   hash that will be passed with the splat operator to the service resource
+#
+# @param daemon_reload
+#   call `systemd::daemon-reload` to ensure that the modified unit file is loaded
+#
+# @param service_restart
+#   restart (notify) the service when unit file changed
+define systemd::quadlet_file (
+  Enum['present', 'absent']                $ensure    = 'present',
+  Stdlib::Absolutepath                     $path      = '/etc/containers/systemd',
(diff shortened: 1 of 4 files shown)


manifests/quadlet_file.pp
```diff
@@ -0,0 +1,113 @@
...
+#
+# @param service_restart
+#   restart (notify) the service when unit file changed
+define systemd::quadlet_file (
+  Enum['present', 'absent']                $ensure    = 'present',
+  Stdlib::Absolutepath                     $path      = '/etc/containers/systemd',
+  Optional[Variant[String, Sensitive[String], Deferred]] $content = undef,
+  Optional[String]                         $source    = undef,
+  String                                   $owner     = 'root',
+  String                                   $group     = 'root',
```
bastelfreak, line 63: can you please update those so we don't allow empty string? `String[1]`

dabelenda: Done.
Note: it was copy-pasted from `unit_file.pp`

manifests/quadlet_file.pp
```diff
@@ -0,0 +1,113 @@
...
+#   restart (notify) the service when unit file changed
+define systemd::quadlet_file (
+  Enum['present', 'absent']                $ensure    = 'present',
+  Stdlib::Absolutepath                     $path      = '/etc/containers/systemd',
+  Optional[Variant[String, Sensitive[String], Deferred]] $content = undef,
+  Optional[String]                         $source    = undef,
+  String[1]                                $owner     = 'root',
+  String[1]                                $group     = 'root',
+  String[1]                                $mode      = '0444',
+  Optional[Boolean]                        $enable    = undef,
```
bastelfreak, line 65: do those booleans need to be undef or can we default to true or false?

dabelenda: Not mandatory, it was copy/pasted directly from `unit_file.pp` to keep a consistent API in the module.

If you request a change, I can remove the Optional, and set a default value.

traylenator:
I'd been thinking about doing this.

If this was going to be done I would at least steal the types I added in https://github.com/southalc/podman/pull/83 so the contents of the quadlet files can be managed.

But I question if managing quadlets should be in here at all. These are not unit files processed by systemd. Having said that its easy place to have them and it will not intrude into systemd configuration.

I think a seperate quadlet module using as much of this module as possible would make more sense? Which I'm very happy to work on lifting much of what I did above. Result is somewhat we ignoring the existing podman module but that is fine for the sake of actual progress.

dabelenda:
@traylenator at least for unit_files this module does not manage the content of the files.

All type definitions and templating were created for the dropins. I copied the behavious of the unit_file.pp for everything which is why I did not add types.

I am also unsure if this is the correct spot for this, which is why I did not go all the way, like creating tests and managing the content of the quadlet files.

ekohl:
Like @traylenator I question it. It's not systemd, but podman. And even with podman you need a new version (4.4+ for basic support, which Debian 12 doesn't even have, newer for the more advanced features).

Given the lack of progress on https://github.com/southalc/podman/pull/83 I wouldn't mind taking over that module. Ideally transferring but forking if needed.

ekohl:
I currently don't use the podman module, but I can see how quadlets replace many of the old ways. A puppet-quadlets module sounds like a good plan to me.

TheMeier:
To me it is also not really clear to which domains really belong. Podman? Systemd? I tend more towards podman, thats also where its code is now.

dabelenda:
I agree this is a bit fuzzy in terms of scope, the generator for quadlet is in the podman package. At the same time, there are generators for a bunch of "core" features of podman, like mount points, etc.

An additionnal puppet module for quadlets that installs podman, manages the content of quadlet files, might be a good thing.
Will the new module depend on puppet/systemd for management of daemon-reload or things like that? Will the module contain its own logic for managing service like puppet/systemd ?

Conceptually the two modues will be pretty similar in features I think.

TheMeier:
If there is going to be a new module, I think it should re-use whatever it can from the systemd module

traylenator:
Add quadlet support in own module. - https://github.com/voxpupuli/puppet-quadlets/issues/1

dabelenda:
Closing this PR since we agree that going for another module is the correct path.

PR #482, opened 2024-08-13
add support for journal upload and remote server

trefzer:
add support for uploading journal to a remote server

### DIFF
--- a/manifests/journal_remote.pp
+++ b/manifests/journal_remote.pp
@@ -1,0 +1,37 @@
+# @api private
+# @summary This class manages and configures journal-remote.
+# @see https://www.freedesktop.org/software/systemd/man/journal-remote.conf.html
+#
+# @param package_name
+#  name of the package to install for the functionality
+#
+class systemd::journal_remote (
+  Optional[String[1]] $package_name = undef,
+) {
+  assert_private()
+
+  if $package_name {
+    stdlib::ensure_packages($package_name)
+  }
+
+  service { 'systemd-journal-remote':
+    ensure => running,
+  }
+  $systemd::journal_remote_settings.each |$option, $value| {
+    ini_setting { "journal-remote_${option}":
+      path    => '/etc/systemd/journal-remote.conf',
+      section => 'Remote',
+      setting => $option,
+      notify  => Service['systemd-journal-remote'],
+    }
+    if $value =~ Systemd::JournaldSettings::Ensure {
+      Ini_setting["journal-remote_${option}"] {
+        * => $value,
+      }
+    } else {
+      Ini_setting["journal-remote_${option}"] {
+        value   => $value,
+      }
+    }
+  }
+}
--- a/manifests/journal_remote.pp
+++ b/manifests/journal_remote.pp
@@ -16,6 +16,7 @@ class systemd::journal_remote (
 
   service { 'systemd-journal-remote':
     ensure => running,
+    enable => true,
   }
   $systemd::journal_remote_settings.each |$option, $value| {
     ini_setting { "journal-remote_${option}":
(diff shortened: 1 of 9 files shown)


manifests/journal_upload.pp
```diff
@@ -0,0 +1,38 @@
...
+  if $package_name {
+    stdlib::ensure_packages($package_name)
+  }
+
+  service { 'systemd-journal-upload':
+    ensure => running,
+  }
+  $systemd::journal_upload_settings.each |$option, $value| {
+    ini_setting {
+      $option:
```
bastelfreak, line 22: I think the following syntax is more common in our code base:
```suggestion
    ini_setting {  $option:
```

trefzer: took it from journald.pp, will change.

trefzer:
added journal-remote setting to this merge request to avoid merge conflicts.
This gives the possibility to use combined spec test.

kenyon:
You'll probably need to change the acceptance tests to use Vagrant instead of the default of Docker for systemd to work correctly.

types/journaluploadsettings.pp
```diff
@@ -0,0 +1,12 @@
+# Matches Systemd journal upload config Struct
```
bastelfreak, line 1: to please puppet-strings, please use:
```suggestion
# @summary Matches Systemd journal upload config Struct
```

types/journalremotesettings.pp
```diff
@@ -0,0 +1,16 @@
+# Matches Systemd journal remote config Struct
```
bastelfreak, line 1:
```suggestion
# @summary Matches Systemd journal remote config Struct
```

manifests/journal_upload.pp
```diff
@@ -0,0 +1,37 @@
...
+  Optional[String[1]] $package_name = undef,
+) {
+  assert_private()
+
+  if $package_name {
+    stdlib::ensure_packages($package_name)
+  }
+
+  service { 'systemd-journal-upload':
+    ensure => running,
```
bastelfreak, line 18: to ensure we don't have a corrective change after a reboot, should we add `enable => true`?

trefzer:
@kenyon probably, but no idea howto do so !

manifests/journal_remote.pp
```diff
@@ -0,0 +1,37 @@
...
+  Optional[String[1]] $package_name = undef,
+) {
+  assert_private()
+
+  if $package_name {
+    stdlib::ensure_packages($package_name)
+  }
+
+  service { 'systemd-journal-remote':
+    ensure => running,
```
bastelfreak, line 18: should we also set `enable => true,` to put it into autostart? usually we do that.

manifests/journal_remote.pp
```diff
@@ -0,0 +1,37 @@
...
+# @param package_name
+#  name of the package to install for the functionality
+#
+class systemd::journal_remote (
+  Optional[String[1]] $package_name = undef,
+) {
+  assert_private()
+
+  if $package_name {
+    stdlib::ensure_packages($package_name)
```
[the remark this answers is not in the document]
trefzer: right, but there is currently only one package for upload and remote journal in Debian and RedHat. 
This solution gives you the possibility to einer install upload and remote or both.

manifests/journal_remote.pp
trefzer: done

manifests/journal_remote.pp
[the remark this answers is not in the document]
bastelfreak: Oh it's the same package. This packaging is so stupid :(
yes in that case it makes sense to stick to `stdlib::ensure_packages()`

bastelfreak (approved):

PR #483, opened 2024-08-13
add journal-remote settings

trefzer:
WIP, DO NOT MERGE YET
add possibility to manage journal-remote (central log server)

### DIFF
--- a/data/Debian-family.yaml
+++ b/data/Debian-family.yaml
@@ -1,2 +1,3 @@
 ---
 systemd::nspawn_package: 'systemd-container'
+systemd::journal_remote::package_name: 'systemd-journal-remote'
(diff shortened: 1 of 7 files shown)


data/Debian-family.yaml
```diff
@@ -1,2 +1,3 @@
 ---
 systemd::nspawn_package: 'systemd-container'
+systemd::journal_remote::package_name: 'systemd-journal-remote'
```
traylenator, line 3: Would be good to add an acceptance test to keep confirming all the package names are correct.

trefzer:
I will include this in merge request #482 because of several merge conflicts and the possibility to use same spec tests.

data/Debian-family.yaml
trefzer: right, bu was not able to create a succesfull test. I assume, I cannot start systemd-journal-remote service in the environment. I'm giving up now ;(
(see #482  for the commit trying ;))

PR #485, opened 2024-08-27
Reload udev rules after change

### LINKED ISSUE #484: udev rules not reloaded after change

- Puppet: 7.31
- Ruby: 3.1
- Distribution: Debian 12
- Module version: 7.1.0

```
systemd::udev::rule:
  notify: "Service[systemd-udevd]"
  rules:
    - 'SUBSYSTEM=="block", KERNEL=="md*", ACTION=="change", TEST=="md/stripe_cache_size", ATTR{md/stripe_cache_size}="4096"'
```

## What are you seeing

`notify: "Service[systemd-udevd]"` doesn't reload udev rules

udev rules should be applied after change

deric:

Automatically reload udev rules after a change. 

To enable this feature use
```yaml
systemd::udev_reload: true
```

A rule with a `String` notification would be then appended with `Exec[systemd-udev_reload]`

```yaml
systemd::udev::rule:
  notify: "Service[systemd-udevd]"
```
e.g. 
```
notify => [ Service[systemd-udevd], Exec[systemd-udev_reload] ]
```
would be used.

Fixes #484

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -157,6 +157,9 @@
 #   Config Hash that is used to generate instances of our
 #   `udev::rule` define.
 #
+# @param udev_reload
+#   Whether udev rules should be automatically reloaded upon change.
+#
 # @param machine_info_settings
 #   Settings to place into /etc/machine-info (hostnamectl)
 #
@@ -261,6 +264,7 @@ class systemd (
   Optional[Integer]                                   $udev_event_timeout = undef,
   Optional[Enum['early', 'late', 'never']]            $udev_resolve_names = undef,
   Optional[Variant[Integer,String]]                   $udev_timeout_signal = undef,
+  Boolean                                             $udev_reload = false,
   Boolean                                             $manage_logind = false,
   Systemd::LogindSettings                             $logind_settings = {},
   Boolean                                             $manage_all_network_files = false,
--- a/manifests/udev/rule.pp
+++ b/manifests/udev/rule.pp
@@ -43,7 +43,7 @@ define systemd::udev::rule (
     owner                   => 'root',
     group                   => 'root',
     mode                    => '0444',
-    notify                  => $notify_services,
+    notify                  => $notify_services << 'Exec[systemd-udev_reload]',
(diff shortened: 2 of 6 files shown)


manifests/init.pp
```diff
@@ -261,6 +264,7 @@
   Optional[Integer]                                   $udev_event_timeout = undef,
   Optional[Enum['early', 'late', 'never']]            $udev_resolve_names = undef,
   Optional[Variant[Integer,String]]                   $udev_timeout_signal = undef,
+  Boolean                                             $udev_reload = true,
```
yakatz, line 267: For backwards compatibility, the default should probably be `false`

yakatz (changes requested):

manifests/init.pp
deric: Sure

manifests/udev/rule.pp
```diff
@@ -38,12 +38,19 @@
     fail("systemd::udev::rule - ${name}: param rules is empty, you need to pass rules")
   }
 
+  if $systemd::udev_reload and $notify_services =~ Array {
```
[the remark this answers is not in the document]
deric: Sure, that's much better idea.

manifests/udev/rule.pp
```diff
@@ -38,12 +38,22 @@
...
 
+  if $systemd::udev_reload {
+    if $notify_services =~ Array {
+      $_notify = $notify_services << 'Exec[systemd-udev_reload]'
+    } else {
+      $_notify = [$notify_services, 'Exec[systemd-udev_reload]']
+    }
+  } else {
+    $_notify = $notify_services
+  }
```
kenyon, line 49: Can remove some repetition:

```suggestion
  $_notify = if $systemd::udev_reload {
    if $notify_services =~ Array {
      $notify_services << 'Exec[systemd-udev_reload]'
    } else {
      [$notify_services, 'Exec[systemd-udev_reload]']
    }
  } else {
    $notify_services
  }
```

kenyon (approved):

PR #486, opened 2024-08-27
Speedup tests: execute OS independent code once

deric:
This PR might improve speed of rspec tests, ~2k test cases removed.

OS independent tests will be executed only once instead of running for each supported platform.


spec/defines/network_spec.rb
ekohl, line 1: At least this file isn't entirely fact independent since it uses `systemd_internal_services`.

spec/defines/dropin_file_spec.rb
```diff
@@ -3,187 +3,182 @@
...
-
-        it {
-          expect(subject).to create_file("/etc/systemd/system/#{params[:unit]}.d").with(
-            ensure: 'directory',
-            recurse: 'true',
-            purge: 'true',
-            selinux_ignore_defaults: false
-          )
-        }
+  _, facts = on_supported_os.first
```
ekohl, line 6: Perhaps you can introduce a helper method `default_os_facts`. I also wouldn't mind knowing which OS is being tested on.

ekohl (commented):
I have mixed feelings about this. On the one hand, we're certainly testing way too many combinations that really don't need to be tested. On the other hand, it can also be very easy to miss it when we do become more fact dependent.

Another option is to introduce an alternative to `on_supported_os`. Something like `on_reduced_supported_os` that could be smarter and filter out some combinations: if RHEL is supported and also all compatible versions then it's sufficient to only test RHEL. That's something we can also reuse in many other modules to cut down on CI testing.

spec/defines/network_spec.rb
deric: You're right, but `$fact['systemd_internal_services']` isn't defined in any OS family hiera `data/`. During tests it's currently using the same value from `default_module_facts.yml` for all test cases.

deric:
> On the other hand, it can also be very easy to miss it when we do become more fact dependent.

When you add new feature, you should write tests to cover that feature. By executing the same tests many times with different OS facts, you won't improve quality of tests (when the code doesn't depend on those facts). 

> Another option is to introduce an alternative to `on_supported_os`. Something like `on_reduced_supported_os` that could be smarter and filter out some combinations.

Having a method like `on_supported_family` where for each OS family it would pick e.g. the latest supported OS would be nice. It

PR #496, opened 2024-10-31
Drop EOL RHEL, CentOS, Fedora, VirtuozzoLinux versions

jay7x:
* VirtuozzoLinux 7
* Fedora 38, 39
* Centos 7, 8
* RHEL 7

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -43,22 +43,13 @@
     {
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
-        "7",
         "8",
         "9"
       ]
     },
     {
-      "operatingsystem": "VirtuozzoLinux",
-      "operatingsystemrelease": [
-        "7"
-      ]
-    },
-    {
       "operatingsystem": "CentOS",
       "operatingsystemrelease": [
-        "7",
-        "8",
         "9"
       ]
     },
@@ -92,8 +83,6 @@
     {
       "operatingsystem": "Fedora",
       "operatingsystemrelease": [
-        "37",
-        "38",
         "40"
       ]
     }
--- a/data/RedHat-family-7.yaml
+++ b/data/RedHat-family-7.yaml
@@ -1,8 +1,0 @@
----
-systemd::nspawn_package: ~
-
-systemd::accounting:
-  DefaultCPUAccounting: 'yes'
-  DefaultBlockIOAccounting: 'yes'
-  DefaultMemoryAccounting: 'yes'
-  DefaultTasksAccounting: 'yes'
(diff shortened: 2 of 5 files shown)


kenyon (approved):

TheMeier:
Please update the PR title to list all affected OSes.

TheMeier (approved):

jay7x:
Do we want to make a release before dropping EOL OSes?

metadata.json
```diff
@@ -42,22 +42,13 @@
     {
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
-        "7",
```
ekohl, line 45: Technically RHEL 7 isn't really EOL. You can still buy ELS for another 4 years: https://www.redhat.com/en/blog/announcing-4-years-extended-life-cycle-support-els-red-hat-enterprise-linux-7

Whether we want to support that is a valid question.

ekohl (changes requested):
If we're dropping this, I'd also like to drop the data files inhttps://github.com/voxpupuli/puppet-systemd/tree/master/data

metadata.json
jay7x: We may offer paid support for RHEL7 in Voxpupuli modules then as well :-D

metadata.json
```diff
@@ -87,13 +78,6 @@
     },
     {
       "operatingsystem": "Gentoo"
-    },
-    {
-      "operatingsystem": "Fedora",
-      "operatingsystemrelease": [
-        "37",
```
silug, line 94: Why drop Fedora support instead of updating the supported versions to `["40", "41"]`?

silug (changes requested):
Why drop Fedora support instead of updating the supported versions to `["40", "41"]`?

jay7x:
> Why drop Fedora support instead of updating the supported versions to `["40", "41"]`?

Hi @silug! I saw somewhere before that dropping EOL OSes and adding new OSes should be different PRs.. though.. maybe it was about different commits instead 🤔

data/Fedora.yaml
```diff
@@ -1,8 +0,0 @@
----
-systemd::accounting:
```
bastelfreak, line 2: my problem with this is that it actually breaks Fedora. Aren't the settings the same on all operating systems nowadays? can't we move that to init.pp instead?

jay7x: I'm heading to redo the PR with adding new OSes before dropping old. So this file will stay.

jay7x:
Converting to Draft. Will work on adding new OSes first (to keep Fedora files untouched e.g.).

jay7x:
Should be ready to merge now

bastelfreak:
Before we merge this, I would like to do a minor release to bring out all merged changes: https://github.com/voxpupuli/puppet-systemd/pull/501

bastelfreak:
ah derp, we already have https://github.com/voxpupuli/puppet-systemd/pull/480 as a breaking change

bastelfreak (approved):

PR #499, opened 2024-11-14
feat: use systemctl json output to build up fact for internal services

rwaffen:


traylenator:
Certainly on EL8 the json option does nothing good:

```
# systemctl list-unit-files "systemd-*" -t service --state=enabled,disabled,enabled-runtime,indirect -o json
UNIT FILE                STATE   
systemd-nspawn@.service  disabled
systemd-pstore.service   enabled 
systemd-resolved.service disabled

3 unit files listed.
```

rwaffen:
hmpf, damnit :(

rwaffen:
okay, this feature comes in with version 239 (May 2017)

lib/facter/systemd.rb
```diff
@@ -50,14 +50,10 @@
...
   confine systemd: true
   setcode do
-    command_output = Facter::Util::Resolution.exec(
-      'systemctl list-unit-files --no-legend --no-pager "systemd-*" -t service --state=enabled,disabled,enabled-runtime,indirect'
-    )
-    lines = command_output.lines.lazy.map { |line| line.split(%r{\s+}) }
-    lines.each_with_object({}) do |(service, status, *), result|
-      result[service] = status
-    end
+    command_output = Facter::Util::Resolution.exec('systemctl list-unit-files "systemd-*" -t service --state=enabled,disabled,enabled-runtime,indirect -o json')
```
bastelfreak, line 56: I didn't use it in the past because not all systemctl versions support the option. Do you know when the option was introduced?

rwaffen: google told me: version 239 (May 2017). maybe i drop this PR. if we still have to support older systemds the more "hacky" version was a better approach for this 🤔

rwaffen: may i wait until the older osses are dropped 😆 
https://github.com/voxpupuli/puppet-systemd/pull/496

traylenator: EL8 will not be dropped with that patch and EOL for EL8 is 2029.

rwaffen: ah damnit okay... then i think i drop it for the moment. the current implementation is working for older and newer versions of systemd. didnt know the feature is this "new" 😆

traylenator:
> Certainly on EL8 the json option does nothing good:
> 
> ```
> # systemctl list-unit-files "systemd-*" -t service --state=enabled,disabled,enabled-runtime,indirect -o json
> UNIT FILE                STATE   
> systemd-nspawn@.service  disabled
> systemd-pstore.service   enabled 
> systemd-resolved.service disabled
> 
> 3 unit files listed.
> ```
> 

It's documented as working in EL8. I'll report it as a bug to Redhat. Maybe it will get fixed.

jcpunk:
With EL8 in bugfix only mode, I'd be curious if this works now.

Alas, I don't have any EL8 boxes left to test against.

PR #502, opened 2024-11-26
Support management of Swap unit files

traylenator:

Let `.swap` unit files be managed with `systemd::manage_dropin` and `systemd::manage_unit`.

For example:

```puppet
systemd::manage_unit{'swapfile.swap':
  ensure        => 'present',
  active        => true,
  enable        => true,
  unit_entry    => {
    'Description' => 'Enable a swapfile at /swapfile',
  }
  swap_entry    => {
    'What' => '/swapfile',
  }
  install_entry => {
    'WantedBy' => 'multi-user.target',
  },
}
```

* https://www.freedesktop.org/software/systemd/man/latest/systemd.swap.html


kenyon:
Huh, looks like this can pretty much replace our entire swap_file module. https://github.com/voxpupuli/puppet-swap_file

traylenator:
> Huh, looks like this can pretty much replace our entire swap_file module. https://github.com/voxpupuli/puppet-swap_file

Indeed. It has the advantage you can make it async to the puppet run which matters when creating a 100GBs of zeros.

Issue #505, opened 2025-01-07
Duplicate resource


ralfbosz:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 7.2.0
- Ruby:
- Distribution: RedHat 8
- Module version: v8.0.0

## How to reproduce (e.g Puppet code you use)

systemd::unit_file { '<application>service':
  ensure  => present,
  ..
}

class { 'systemd':
  manage_journald   => true,
  journald_settings => true,
}      

## What are you seeing

Duplicate resource for class systemd

## What behaviour did you expect instead

No duplicate resource, in the transition from the camptocamp to the voxpupuli release, an include was set in systemd::unit_file to include systemd. Since we directly include the class we get an 'duplicate resource'. Ideal we would set the hiera-entry for systemd directly, but this is not possible in our setup (different teams).

By setting an 'if !defined' around all thse includes this should be solved.

PR #506, opened 2025-01-07
Enclose the 'include systemd' in the module

ralfbosz:
When using a different module to call the systemd
class and also using systemd::unit_file you get a
duplicate resource. By this 'if !defined' this is
solved. Issue #505


bastelfreak:
Hi,
I don't think that this is a good idea. You're running into errors because the ordering of your resources isn't correct. You need to call the systemd class before systemd::unit_file resources.

ralfbosz:
Okay I will close it...

Issue #510, opened 2025-01-15
unit_file should accept relative paths


anarcat:
This seems, to me, perfectly acceptable:

```puppet
    systemd::unit_file { "pgbackrest-backup-${kind}@${shortname}.service":
      enable => false,
      active => false,
      target => "pgbackrest-backup-${kind}@.service",
    }
```

yet `unit_file` wants a stdlib AbsolutePath so this crashes violently with:

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Systemd::Unit_file[pgbackrest-backup-full@rude.service]: parameter 'target' expects a Stdlib::Absolutepath = Variant[Stdlib::Windowspath = Pattern[/\A(([a-zA-Z]:[\\\/])|([\\\/][\\\/][^\\\/]+[\\\/][^\\\/]+)|([\\\/][\\\/]\?[\\\/][^\\\/]+)).*\z/], Stdlib::Unixpath = Pattern[/\A\/([^\n\/\0]+\/*)*\z/]] value, got String (file: /etc/puppet/code/environments/prom_module_upgrade/3rdparty/modules/pgbackrest/manifests/repository/stanza.pp, line: 77) on node backup-storage-01.torproject.org
```

seems to me we should be a little more tolerant here.

thoughts? concretely i think this means just switching to a plain string...

smortex:
That looks very reasonable to me! :+1:

PR #511, opened 2025-01-15
unit_file: allow arbitrary symlink targets

anarcat:
Absolutepath is exactly what it says: it forces the link target to be an absolute path, that is to start with a slash (`/`). But I actually *want* to have relative symlinks.

This is particularly useful when doing a @.service template where the instance refers to the template, often in the same directory.

This, for example, should be allowed:

    systemd::unit_file { "pgbackrest-backup-${kind}@${shortname}.service":
      enable => false,
      active => false,
      target => "pgbackrest-backup-${kind}@.service",
    }

Yet it currently crashes with:

    Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Systemd::Unit_file[pgbackrest-backup-full@rude.service]: parameter 'target' expects a Stdlib::Absolutepath = Variant[Stdlib::Windowspath = Pattern[/\A(([a-zA-Z]:[\\\/])|([\\\/][\\\/][^\\\/]+[\\\/][^\\\/]+)|([\\\/][\\\/]\?[\\\/][^\\\/]+)).*\z/], Stdlib::Unixpath = Pattern[/\A\/([^\n\/\0]+\/*)*\z/]] value, got String (file: /etc/puppet/code/environments/prom_module_upgrade/3rdparty/modules/pgbackrest/manifests/repository/stanza.pp, line: 77) on node backup-storage-01.torproject.org

I don't know of another way to fix this than to make this a proper string.

Closes: #510

### DIFF
--- a/manifests/unit_file.pp
+++ b/manifests/unit_file.pp
@@ -73,7 +73,7 @@ define systemd::unit_file (
   Stdlib::Absolutepath                     $path      = '/etc/systemd/system',
   Optional[Variant[String, Sensitive[String], Deferred]] $content = undef,
   Optional[String]                         $source    = undef,
-  Optional[Stdlib::Absolutepath]           $target    = undef,
+  Optional[String]                         $target    = undef,
   String                                   $owner     = 'root',
   String                                   $group     = 'root',
   String                                   $mode      = '0444',


manifests/unit_file.pp
```diff
@@ -73,7 +73,7 @@
   Stdlib::Absolutepath                     $path      = '/etc/systemd/system',
   Optional[Variant[String, Sensitive[String], Deferred]] $content = undef,
   Optional[String]                         $source    = undef,
-  Optional[Stdlib::Absolutepath]           $target    = undef,
+  Optional[String]                         $target    = undef,
```
kenyon, line 76: Empty string isn't valid here, right?

```suggestion
  Optional[String[1]]                      $target    = undef,
```

traylenator:
> systemd::unit_file { "pgbackrest-backup-${kind}@${shortname}.service":
>   enable => false,
>   active => false,
>   target => "pgbackrest-backup-${kind}@.service",
> }


Presumably 

```
target => '/etc/systemd/system/pgbackrest-backup-${kind}@.service'
```
would be okay ? Any reason it needs to be relative ?

Second question,  The instance of a unit `pgbackrest-backup-${kind}@${shortname}.service` can be started , stopped , enabled, disabled without the symlink being created .  Why even create the symlink in the first place?

traylenator:
One other consideration - if targets names are permitted non-absolute maybe they should behave like the unit_name itself does.
These default to  the directory `$path      = '/etc/systemd/system',`  so maybe the target should be assumed to be in `$path` as well?

manifests/unit_file.pp
traylenator: Could it be reduced to  `Enum[Stdlib::Absolutepath,Systemd::Unit,Systemd::Unit]` or really need things like `../sshd.service` for instance.

anarcat:
> Presumably
> 
> ```
> target => '/etc/systemd/system/pgbackrest-backup-${kind}@.service'
> ```
> 
> would be okay ? Any reason it needs to be relative ?

this is what i said in #510: the absolute path works, of course, but it's a lot of typing and obscures the fact that the target is in the same directory, for someone trying to debug things. it might also make some security policies harder to implement (although I admit we don't have those).

> Second question, The instance of a unit `pgbackrest-backup-${kind}@${shortname}.service` can be started , stopped , enabled, disabled without the symlink being created . Why even create the symlink in the first place?

that, i admit ignorance. i guess i made this because i had a hammer and this looked like a nail?

how would you enable the service without the symlink in Puppet?

either way, it's not clear to me at all how to do this with this module, and if i'm doing this wrong, we should document how to do this right.

manifests/unit_file.pp
anarcat: > Could it be reduced to Enum[Stdlib::Absolutepath,Systemd::Unit,Systemd::Unit] or really need things like ../sshd.service for instance.

that would work for me, i don't think there's a case for ../

traylenator:
> > Second question, The instance of a unit `pgbackrest-backup-${kind}@${shortname}.service` can be started , stopped , enabled, disabled without the symlink being created . Why even create the symlink in the first place?

You litrally just do it:

```puppet
service{`pgbackrest-backup-${kind}@${shortname}.service':
   ensure => true,
   enable => true,
 }
 ```

 Just ignore the unit file is not there.

anarcat:
okay, then i'll reroll this as a docs update.

PR #512, opened 2025-01-16
docs: describe how to deploy systemd templates

anarcat:
I've bent over backwards to make this work, while the solution is much simpler than I thought it was.

Let's share the knowledge with the world so that others don't make the same mistake I did.

Closes: #510
Replaces: #511

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ systemd::unit_file { 'foo.service':
 ```
 
 If you're using "template" units (those with a `@` in it), you can
-simply enable it with a `service` resource. This, for example, will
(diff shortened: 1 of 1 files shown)


kenyon (approved):

README.md
```diff
@@ -58,6 +58,22 @@ systemd::unit_file { 'foo.service':
 }
 ```
 
+If you're using "template" units (those with a `@` in it), you can
+simply enable it with a `service` resource. This, for example, will
```
traylenator, line 62:
```suggestion
simply enable an instance of the template with a `service` resource. This, for example, will
```

traylenator (approved):

PR #524, opened 2025-03-23
`resolved` acceptance test: add Ubuntu 24.04

kenyon:
Ubuntu 24.04 and newer have a separate `systemd-resolved` package.


kenyon:
Test passes because the `systemd` package recommends `systemd-resolved`, even though this module's hiera data is not ensuring that `systemd-resolved` is installed.

PR #548, opened 2025-08-06
Allow quadlets to be referenced as units

JGodin-C2C:

This PR allows quadlets to be referenced as units. 
As the puppet-quadlet uses systemd::unit to validate the entry , it think that is an interesting feature.
Reference here.
 
https://github.com/voxpupuli/puppet-quadlets/blob/master/REFERENCE.md#unit_entry

HOWEVER, i understand that systemd  module should not bear this burden, I would be more than happy to implement alternative solutions.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -5430,7 +5430,7 @@ custom datatype that validates different filenames for systemd units and unit te
 * **See also**
   * https://www.freedesktop.org/software/systemd/man/systemd.unit.html
 
-Alias of `Pattern[/^[a-zA-Z0-9:\-_.\\@%]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope)$/]`
+Alias of `Pattern[/^[a-zA-Z0-9:\-_.\\@%]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope|volume|pod|kube|container)$/]`
 
 ### <a name="Systemd--Unit--Amount"></a>`Systemd::Unit::Amount`
 
--- a/types/unit.pp
+++ b/types/unit.pp
@@ -1,3 +1,3 @@
 # @summary custom datatype that validates different filenames for systemd units and unit templates
 # @see https://www.freedesktop.org/software/systemd/man/systemd.unit.html
-type Systemd::Unit = Pattern[/^[a-zA-Z0-9:\-_.\\@%]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope)$/]
+type Systemd::Unit = Pattern[/^[a-zA-Z0-9:\-_.\\@%]+\.(service|socket|device|mount|automount|swap|target|path|timer|slice|scope|volume|pod|kube|container)$/]


traylenator:
> puppet-quadlet uses systemd::unit 

It does not really ? 

Things like activate, deactivate make no sense to a .container file , its not a unit.

This is a close from me.

traylenator:
Can you give me a puppet example of what you want to work ?

JGodin-C2C:
Sure ! here you go : 

```
  quadlets::quadlet { 'openvoxservercode.volume':
    ensure       => present,
    unit_entry   => {
      'Description' => 'openvoxserver Code Volume',
    },
    volume_entry => {
      'VolumeName' => 'openvoxservercode',
      'Type'       => 'nfs',
      'Device'     => 'mynfsserver.com:/mynfspath',
      'Options'    => 'o=rw,nfsvers=4',
    },
  }
  
  
  quadlets::quadlet{'centos.container':
  ensure          => present,
  unit_entry     => {
   'Description' => 'Trivial Container that will be very lazy',
   'Requires'    => 'openvoxservercode.volume', 	< This one
   'Wants'       => 'openvoxservercode.volume'          < or this one 
  },
  service_entry       => {
    'TimeoutStartSec' => '900',
  },
  container_entry => {
    'Image' => 'quay.io/centos/centos:latest',
    'Exec'  => 'sh -c "sleep inf"'
    'Mount' => ['openvoxservercode:/openvoxservercode'] < Not sure about this syntax, but you get the idea
  },
  install_entry   => {
    'WantedBy' => 'default.target'
  },
  active          => true,
}
```

So, I would like the volume to be mounted BEFORE the container is launched. 
I came up with this solution. 
The problem is , if we do not add the Requires or Wants here,  a new named volume is created and that could lead to errors.

traylenator:
Thanks for the example, I played and I now understood.

```puppet
quadlets::quadlet { 'openvoxservercode.volume':
    ensure       => present,
    unit_entry   => {
      'Description' => 'openvoxserver Code Volume',
    },
    volume_entry => {
      'VolumeName' => 'openvoxservercode',
      'Driver'     => 'local',
    },
  }
  
  
  quadlets::quadlet{'centos.container':
  ensure          => present,
  unit_entry     => {
   'Description' => 'Trivial Container that will be very lazy',
   'Requires'    => 'openvoxservercode.volume',
   'After'       => 'openvoxservercode.volume',
  },
  service_entry       => {
    'TimeoutStartSec' => '900',
  },
  container_entry => {
    'Image' => 'quay.io/centos/centos:latest',
    'Exec'  => 'sh -c "sleep inf"',
    'Volume' => ['openvoxservercode:/foobar:Z'],
  },
  install_entry   => {
    'WantedBy' => 'default.target'
  },
  active          => true,
}

```

as you say this needs to be possible.

However this patch is the wrong fix for that.

We should change the type here  

https://github.com/voxpupuli/puppet-quadlets/blob/master/manifests/quadlet.pp#L59C3-L59C52

and create a new type `Quadlet::Unit::Unit` its annoying we will have to duplicate much of  `Systemd::Unit::Unit` but they are as demonstrated here not the same.



In the immediate term to unstick your self 

```
'Requires'    => 'volume-openvoxservercode.service',
'After'       => 'volume-openvoxservercode.service',
```
Will presumably work.


@JGodin-C2C Can you create a new issue or patch in the quadlets module?

PR #550, opened 2025-08-11
Support for Debian 13 (Trixie)

### LINKED ISSUE #549: Support Debian 13 (Trixie)
### Expected behavior:
With this configuration in hiera:
```
systemd::manage_timesyncd: true
systemd::timesyncd_ensure: "running"
```
The package systemd-timesyncd shall be installed and the service shall be started. This works in Debian 12 (Bookworm).

### Actual behavior:
No package was installed. The service is missing.

Trixie was released on August 9th. It would be great if you could support it. I don't use very much of this module that's why i haven't a complete list of what is needed to support Trixie.

luitzifa:
Add support for Debian 13 (Trixie) and get rid of duplicated code by utilizing os-family.yaml functionality.

Fixes #549


luitzifa:
@bastelfreak there you go

TheMeier:
If you remove data/Debian-11.yaml you also need to remove d11 from metadata and mark this as breaking

luitzifa:
@TheMeier I don´t understand why it should be breaking. All default that were provided by the Debian-11.yaml were moved to the Debian-family.yaml and should be available to Debian 11 agents, shouldn't they?

smortex (approved):

PR #562, opened 2025-09-05
Allow to disable the DNS stub listener

mpdude:
I guess #429 broke the ability to set

```
    class { '::systemd':
        dns_stub_listener => false,
    }
```

... and get `DNSStubListener = no` in `/etc/systemd/resolved.conf`, which was working before.

I am not at all familiar with the test suite here or Ruby tests in general, so I could use some help to make this pass.

### DIFF
--- a/manifests/resolved.pp
+++ b/manifests/resolved.pp
@@ -239,7 +239,7 @@ class systemd::resolved (
     default => $dns_stub_listener,
   }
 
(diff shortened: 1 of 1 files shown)


ekohl:
> I am not at all familiar with the test suite here or Ruby tests in general, so I could use some help to make this pass.

It's passing. Is your question "how do I implement a test for this"?

PR #563, opened 2025-09-16
Add EL10 support

bastelfreak:


bastelfreak:
This will have to work until https://packages.fedoraproject.org/pkgs/systemd-extras/systemd-networkd/ is moved from EPEL Testing to regular EPEL.

PR #573, opened 2025-10-15
Add `AssertFileNotEmpty` as optional parameter for unit

jplindquist:
Add `AssertFileNotEmpty` as an optional parameter for a unit entry, similar to ConditionFileNotEmpty, just logged more loudly :)

https://man7.org/linux/man-pages/man5/systemd.unit.5.html
           Added in version 218


jplindquist:
Thanks for looking so quick @kenyon, I don't know if the other test failure is caused by this, or something else?

kenyon:
That was just a network error. I've rerun the job.

PR #577, opened 2025-10-23
Use `runuser`/`run0` to connect to `systemd --user` instance

### LINKED ISSUE #459: systemd::user_service non functional with puppet running in background daemon mode

- Puppet: 7.24
- Ruby: 2
- Distribution: RHEL9
- Module version: 7.0.0

```
systemd::user_service { "gitlab-podman-auto-update.timer":
  ensure  => true,
  enable  => true,
  unit    => podman-auto-update.timer,                            
  user    => 'gitlab-runner',
}
```
Run puppet agent in background, i.e `pkill -SIGUSR1 puppet` and not `puppet agent -t -v`

## What are you seeing

The service does not start, the type fails to check the status of the service assumes it is not running and then
fails to start it.  ( In fact the service is running however that is some what immaterial if puppet can't detect that. 

The service should be  started and enabled enabled correctly.

## Output log

```
Apr 29 19:38:46 cirunner01.example.ch puppet-agent[1816569]: (/Stage[main]/Hg_punch::Cirunners/Systemd::User_service[gitlab-runner-podman-auto-update.timer]/Exec[Start user service podman-auto-update.timer for user gitlab-runner]/returns) Failed to start transient service unit: Transport endpoint is not connected
Apr 29 19:38:46 cirunner01.example.ch puppet-agent[1816569]: '["systemd-run", "--pipe", "--wait", "--user", "--machine", "gitlab-runner@.host", "systemctl", "--user", "start", "podman-auto-update.timer"]' returned 1 instead of one of [0]
Apr 29 19:38:46 cirunner01.example.ch puppet-agent[1816569]: (/Stage[main]/Hg_punch::Cirunners/Systemd::User_service[gitlab-runner-podman-auto-update.timer]/Exec[Start user service podman-auto-update.timer for user gitlab-runner]/returns) change from 'notrun' to ['0'] failed: '["systemd-run", "--pipe", "--wait", "--user", "--machine", "gitlab-runner@.host", "systemctl", "--user", "start", "podman-auto-update.timer"]' returned 1 instead of one of [0] (corrective)
```

Note the error:
```
 Failed to start transient service unit: Transport endpoint is not connected
```

Expanding the underlying exec it is:

```bash
systemd-run --pipe --wait --user --machine gitlab-runner@.host  systemctl --user   is-active podman-auto-update.timer                                                  
```
This command is fine from the cmd line:

```
# systemd-run --pipe --wait --user --machine gitlab-runner@.host  systemctl --user   is-active podman-auto-update.timer                                                  
Running as unit: run-u10229.service
active
Finished with result: success
Main processes terminated with: code=exited/status=0
Service runtime: 7ms
CPU time consumed: 6ms
```
Just will not run in background puppet exec.

traylenator:

Create a new function `systemd::systemctl_user` to construct array of command and args used to contact `systemd --user` instances so `systemctl --user status my.service` can be executed.

On systemd >= 256 `run0` is used with `runuser` used on older OSes (RHEL9, Debian 12, .. )

A new boolean is also added `install_runuser` to install `util-linux` if needed which seems to be the correct package for all distributions. In reality this package will be installed on most systems anyway.

We switch to using `runuser` or `run0` to become the user since:

* Using `systemctl` or `systemd-run` with `--machinectl $user@.host` always fails when puppet is ran inside its own systemd unit. I believe systemd services can not access `systemd --user` instances.
* We use `runuser -u $user` rather than a `user => $user` on the exec since the $(id -u $user) must be resolved late within the exec itself.

Now the command is generated inside a function this will make future improvements to this method much simpler.

* Fixes #459

### DIFF
--- a/functions/systemctl_user.pp
+++ b/functions/systemctl_user.pp
@@ -1,0 +1,29 @@
+# @summary Construct command array for running `systemctl --user` for particular arguments
+# @param user User instance of `systemd --user` to connect to.
+# @param arguments Arguments to run after `systemctl --user`
+#
+# @return Array Array of command and arguments
+# @example Start a user service with an exec
+#   exec { 'start_service':
+#    command => systemd::systemctl_user('santa', 'start myservice.service'),
+#   }
+function systemd::systemctl_user(String[1] $user, Array[String[1],1] $arguments) >> Array {
+  # Why is runuser being used here for older systemds?
+  # More obvious command arrays to return would be:
+  # * ['run0','--user',$user,'systemctl','--user'] + $arguments
+  # * ['systemctl', '--user', '--machine', "${user}@.host"] + $arguments
+  # * ['systemd-run','--wait','--pipe', 'systemctl', '--user', '--machine', "${user}@.host"] + $arguments
+  # However none of these work when puppet is run as a background service. They only work with
+  # puppet apply in the foreground. Reason is unclear, possibly polkit blocking access
+  # https://github.com/voxpupuli/puppet-systemd/issues/459
+
+  $_cmd_array  = Integer($facts['systemd_version']) < 256 ? {
+    true    => [
+      'runuser', '-u', $user, '--' ,'/usr/bin/bash', '-c',
+      "env XDG_RUNTIME_DIR=/run/user/\$(id -u) /usr/bin/systemctl --user ${arguments.join(' ')}",
+    ],
+    default => ['run0','--user',$user,'/usr/bin/systemctl','--user'] + $arguments,
+  }
+
+  return $_cmd_array
+}
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -256,6 +256,10 @@
 #   Config Hash that is used to configure settings in user.conf
 #   NOTE: It's currently impossible to have multiple entries of the same key in
 #   the settings.
+#
+# @param install_runuser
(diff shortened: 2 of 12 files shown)


functions/systemctl_user.pp
```diff
@@ -0,0 +1,22 @@
+# @summary Construct command array for running `systemctl --user` for a particular arguments
```
kenyon:
```suggestion
# @summary Construct command array for running `systemctl --user` for particular arguments
```

manifests/init.pp
```diff
@@ -256,6 +256,10 @@
 #   Config Hash that is used to configure settings in user.conf
 #   NOTE: It's currently impossible to have multiple entries of the same key in
 #   the settings.
+#
+# @param install_runuser
+#   If true parameter the util-linux package is installed, for runuser command.
```
kenyon:
```suggestion
#   If true, the util-linux package is installed, for runuser command.
```

manifests/user_service.pp
```diff
@@ -1,6 +1,9 @@
 # @summary Manage a user service running under systemd --user
 #
 # @example Enable a service for all users
+#  class['systemd:
+#    runuser => true
```
kenyon:
```suggestion
#  class { 'systemd':
#    install_runuser => true,
```

PR #585, opened 2025-11-09
Support creation of `sysusers.d` entries

### LINKED ISSUE #584: Support for managing `sysusers.d` files
It would be handy if I could also define some of my system users via sysusers.

https://www.freedesktop.org/software/systemd/man/latest/sysusers.d.html#

The `systemd::dropin_file` type is close focused on services and their needs.

traylenator:

Example:

```puppet
systemd::sysuser { 'plato.conf':
  content => 'u plato - "Be kind"',
}
```

will do the following

* Create the directory `/etc/sysusers.d`
* Validate the new file `systemd-sysuser --dry-run` so long as the systemd version is more than 250.
* Create thefile `/etc/sysusers.d/plato.conf`
* Run `systemd-sysusers`.

Reference:

* https://www.freedesktop.org/software/systemd/man/latest/sysusers.d.html
* https://www.freedesktop.org/software/systemd/man/latest/systemd-sysusers.html

* Fixes #584

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ systemd::modules_load { 'impi.conf':
 Module can create sysusers.d files and will run `systemd-sysusers` afterwards.
 
 ```puppet
-systemd::sysuser{ 'plato.conf':
+systemd::sysuser { 'plato.conf':
   content => 'u plato - "Be Kind"',
 }
 ```
--- a/manifests/sysuser.pp
+++ b/manifests/sysuser.pp
@@ -24,25 +24,25 @@
 #   Validate the file `systemd-sysusers --dry-run`, The parameter is ignored for systemd version less than 250 where validation is not available.
 #
 # @example Add a user with systemd-sysusers
-#   systemd::sysuser{ 'plato':
+#   systemd::sysuser { 'plato':
 #     content => 'u plato - "be kind"',
 #   }
 #
 # @example Manage /etc/sysusers.d directory with purge disabled
-#   class{' systemd::sysusers:
+#   class { 'systemd::sysusers:
 #     purgedir => false,
 #   }
-#   systemd::sysuser{ 'plato':
+#   systemd::sysuser { 'plato':
 #     content => 'u plato - "be kind"',
 #   }
 #
 define systemd::sysuser (
-  Enum['present', 'absent']         $ensure   = 'present',
-  Systemd::Dropin                   $filename = $name,
-  Stdlib::Absolutepath              $path     = '/etc/sysusers.d',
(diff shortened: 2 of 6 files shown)


manifests/sysusers.pp
```diff
@@ -0,0 +1,27 @@
...
+    purge   => $purgedir,
+    recurse => $purgedir,
+    force   => $purgedir,
+  }
+
+  exec { 'systemd-sysusers':
+    command     => 'systemd-sysusers',
+    refreshonly => true,
+    path        => $facts['path'],
+  }
```
kenyon: Should the `systemd-sysusers` exec run if the file resource removes some files from `/etc/sysusers.d`?

manifests/sysuser.pp
```diff
@@ -0,0 +1,69 @@
...
+#   * Mutually exclusive with `$source`
+#
+# @param source
+#   A `File` resource compatible ``source``
+#
+# @param validate
+#   Validate the file `systemd-sysusers --dry-run`, The parameter is ignored for systemd version less than 250 where validation is not available.
+#
+# @example Add a user with systemd-sysusers
+#   systemd::sysuser{ 'plato':
```
kenyon:
```suggestion
#   systemd::sysuser { 'plato':
```

README.md
```diff
@@ -195,6 +195,16 @@ systemd::modules_load { 'impi.conf':
 }
 ```
 
+### sysusers
+
+Module can create sysusers.d files and will run `systemd-sysusers` afterwards.
+
+```puppet
+systemd::sysuser{ 'plato.conf':
```
kenyon:
```suggestion
systemd::sysuser { 'plato.conf':
```

manifests/sysusers.pp
traylenator: It's harmless but indeed pointless. I will remove the notify for that case.

kenyon: Oh, I did not see a notify for that case, I was wondering if there needs to be one 🙂

manifests/sysuser.pp
```diff
@@ -0,0 +1,69 @@
...
+) {
+  include systemd::sysusers
+
+  $_sysusers_file_ensure = $ensure ? {
+    'present' => 'file',
+    default   => $ensure,
+  }
+
+  $_validate_cmd = ($validate and Integer($facts['systemd_version']) >= 250 ) ? {
+    true    => '/usr/bin/systemd-sysusers --dry-run %',
```
traylenator: For the validation maybe we should set `--replace` flag as well to specify the eventual path of the file.

Issue #586, opened 2025-11-14
incorrect network options (multiple)


trefzer:
The network types have the following issues: 

* MIIMonitorSec for bonds should be of type Numeric (currently Integer)
* correct vlan section name is VLAN (not Vlan)

PR #587, opened 2025-11-14
Fix incorrect network options (multiple)

trefzer:
The network types have the following issues:

* MIIMonitorSec for bonds should be of type Numeric (currently Integer)
* correct vlan section name is VLAN (not Vlan)

Fixes: #586

### DIFF
--- a/spec/type_aliases/interface/netdev_spec.rb
+++ b/spec/type_aliases/interface/netdev_spec.rb
@@ -7,6 +7,7 @@ describe 'Systemd::Interface::Netdev' do
     context 'with valid types' do
       [
         {},
+        { 'VLAN' => { 'Id' => 42 } },
       ].each do |value|
         describe value.inspect do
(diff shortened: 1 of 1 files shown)


REFERENCE.md
```diff
@@ -3524,7 +3524,7 @@ Struct[{
   ]],
   'TransmitHashPolicy'           => Optional[Enum['layer2', 'layer3+4', 'layer2+3', 'encap2+3', 'encap3+4']],
   'LACPTransmitRate'             => Optional[Enum['slow','fast']],
-  'MIIMonitorSec'                => Optional[Integer],
+  'MIIMonitorSec'                => Optional[Numeric],
```
bastelfreak: can you please add a unit test here that explicitly tests for a numeric, so people don't update it later to an integer again?

kenyon (approved):

PR #594, opened 2025-11-26
Also permit `keep` for journald `Audit`

jcpunk:
The `Audit` key also supports `keep`.


kenyon (approved):
I didn't see `keep` in the documentation, it is apparently not released yet:

* systemd 258 docs: https://www.freedesktop.org/software/systemd/man/latest/journald.conf.html#Audit=
* commit that adds `keep`: https://github.com/systemd/systemd/commit/b5fdfedf729712b9824a5cb457a07d5699d2946c

PR #596, opened 2025-11-27
fix: networkd is not an internal service

vchepkov:
systemd-networkd is an external package managed by the module and existing code doesn't manage service until second puppet run

### DIFF
--- a/manifests/network.pp
+++ b/manifests/network.pp
@@ -26,7 +26,7 @@ define systemd::network (
 ) {
   include systemd
 
-  if $restart_service and $systemd::manage_networkd and $facts['systemd_internal_services'] and $facts['systemd_internal_services']['systemd-networkd.service'] {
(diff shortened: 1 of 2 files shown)


manifests/network.pp
```diff
@@ -26,7 +26,7 @@
 ) {
   include systemd
 
-  if $restart_service and $systemd::manage_networkd and $facts['systemd_internal_services'] and $facts['systemd_internal_services']['systemd-networkd.service'] {
+  if $restart_service and $systemd::manage_networkd {
     $notify = Service['systemd-networkd']
   } else {
```
trefzer, line 31: I suggest you can also change that if/else to
```
if $restart_service and $systemd::manage_networkd {
  File["${path}/${name}"] ~> Service['systemd-networkd']
}
```
So no else needed, and remove the notify line (line 52).

vchepkov:
I will add this change in a separate MR

PR #597, opened 2025-12-05
Rewrite `loginctl_user` provider

### LINKED ISSUE #301: loginctl disable-linger runs even when disabled

- Puppet:7.17.0
- Ruby:ruby-3.0.4-160.el9_0.x86_64
- Distribution: AlmaLinux 9
- Module version: 3.10.0

```yaml
stdlib::manage::create_resources:
  loginctl_user:
    sysoptmgmt:
      linger: disabled

```

## What are you seeing

Every puppet run sets this to `disabled` even though it is already disabled.  Seems like the check to see if it is already disabled isn't actually working....

Already disabled linger users shouldn't trigger a corrective action

## Output log

```
Notice: /Stage[main]/Stdlib::Manage/Loginctl_user[sysoptmgmt]/linger: defined 'linger' as 'disabled' (corrective)
```

### LINKED ISSUE #302: Race condition within loginctl_user type

- Puppet: 7.21.0
- Ruby: vendored
- Distribution: CentOS Stream 9
- Module version: 469b0f271797e8dac57ba6c29822c92e96199894

I was running puppet on a node and suddenly I got:

```
Error: Could not prefetch loginctl_user provider 'ruby': Execution of '/bin/loginctl show-user -p Name -p Linger root mail nginx-edge' returned 1: Failed to get user: User ID 8 is not logged in or lingering
Name=root
Linger=no
```

nginx-edge is a lingering user. mail normally not, BUT user mail has certain systemd timers that are running.

Since managing the lingering state of nginx-edge failed, management of that user failed.

When looking at the code (https://github.com/voxpupuli/puppet-systemd/blob/1d785f129b5b5395860b1d39723e0aa7b31f7c74/lib/puppet/provider/loginctl_user/ruby.rb) we see that it first the currently users and then tries to get more info about them in one shot.

When I ran the first command manually, the user mail did not show up. Also subsequent puppet runs never had that issue anymore.

So what I assume is that at the moment puppet run the first command the user mail had a systemd timer running and thus was active and showed up in the `show-user` command. BUT when running the second command, the timer was already done and thus the user not anymore active and thus it didn't show up.

What is a bit worse is that the provider errors out and thus everything errors out.

Not really sure how to fix that issue properly, but it can definitely be that you hit that race condition.

trefzer:
The current provider uses two steps in the self.instances routine:
1. Fetch all users currently logged in with the command `loginctl list-users`. Result is a list of all users with an open session (so this includes user which are lingering, users that are logged in etc.).
2. For each found user the command `loginctl show-user` is executed to get the current state of the 'Linger' setting.

The second command also fails for users not logged in. Since the first command only lists logged in users, the failure is a quite seldom event (also described in #302 ).

A user with disabled linger is 'usually' (at least I'm not logged in on all machines running puppet ;)) not logged in, so it does not appear on the first command. Therefore it's linger status is not checked. This results in running loginctl to disable the setting on each puppet run (described in #301 ).

The approach I took in this MR is to get the Linger status in a single step and handle the failure in case the user is not logged in. If the user is logged in, we an use the Linger state to decide if lingering is activated or not. In the case the user is not logged in, it's for sure not lingering (since these are listed). So setting the is value to disabled should be correct.

Fixes: #302
Fixes: #301

### DIFF
--- a/lib/puppet/provider/loginctl_user/ruby.rb
+++ b/lib/puppet/provider/loginctl_user/ruby.rb
@@ -5,29 +5,17 @@
 # @see https://wiki.archlinux.org/title/Systemd/User
 Puppet::Type.type(:loginctl_user).provide(:ruby) do
   desc 'custom provider to manage systemd user sessions/linger'
+
   commands loginctl: 'loginctl'
 
-  def self.instances
-    users = loginctl('list-users', '--no-legend').split("\n").map { |l| l.split[1] }
-    loginctl('show-user', '-p', 'Name', '-p', 'Linger', *users).split("\n\n").map do |u|
-      user = u.split("\n").to_h { |f| f.split('=') }
-      linger = if user['Linger'] == 'yes'
-                 'enabled'
-               else
-                 'disabled'
-               end
-      new(name: user['Name'],
-          linger: linger)
-    end
+  def linger
+    # loginctl is only successful if the user has an active session (so either logged in or lingering
+    # so if loginctl fails, linger is definitly disabled, for users with an active session
+    # (eg. logged in or running a timer or ...), the Linger property displays if lingering is activated.
+    :enabled if loginctl('show-user', resource[:name], '--property=Linger', '--value').chomp == 'yes'
+  rescue Puppet::ExecutionFailure
+    :disabled
   end
-
-  def self.prefetch(resources)
-    instances.each do |prov|
-      resources[prov.name].provider = prov if resources[prov.name]
-    end
-  end
-
-  mk_resource_methods
 
   def linger=(value)
     case value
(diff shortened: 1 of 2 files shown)


kenyon (approved):

lib/puppet/provider/loginctl_user/ruby.rb
```diff
@@ -5,30 +5,18 @@
...
-    instances.each do |prov|
-      resources[prov.name].provider = prov if resources[prov.name]
-    end
+  def linger
+    # loginctl is only successful if the user has an active session (so either logged in or lingering
+    # so if loginctl fails, linger is definitly disabled, for users with an active session
+    # (eg. logged in or running a timer or ...), the Linger property displays if lingering is activated.
+    :enabled if loginctl('show-user', resource[:name], '--property=Linger', '--value').chomp == 'yes'
+  rescue Puppet::ExecutionFailure
+    :disabled
```
smortex, line 17: This return `nil` if `loginctl('show-user', resource[:name], '--property=Linger', '--value').chomp == 'yes'` is not true.  Not sure this is expected.

trefzer: you are right. And this happens (and is relevant) if the queried user is logged in and loginctl is 'no'.
This would not hurt but in that moment the code is not idempotent anymore. Fixed.

kenyon:
#625 fixes the test failure.

kenyon (approved):

PR #610, opened 2026-02-24
`Systemd::Unit::Unit`: Allow `infinity` for `StartLimitBurst`

### LINKED ISSUE #609: Support `infinity` for `StartLimitBurst`
The documentation:
https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html
says:
'infinity' is an valid value for StartLimitBurst.
But the module reject it with:
`parameter 'unit_entry' entry 'StartLimitBurst' expects an Integer value, got String`

tuxmaster5000:
Allow infinity for StartLimitBurst
Fixes #609


types/unit/unit.pp
```diff
@@ -50,6 +50,6 @@
     Optional['ConditionFileNotEmpty']       => Variant[Enum[''],Stdlib::Unixpath,Pattern[/^!.*$/],Array[Variant[Enum[''],Stdlib::Unixpath,Pattern[/^!.*$/]],1]],
     Optional['ConditionFileIsExecutable']   => Variant[Enum[''],Stdlib::Unixpath,Pattern[/^!.*$/],Array[Variant[Enum[''],Stdlib::Unixpath,Pattern[/^!.*$/]],1]],
     Optional['StartLimitIntervalSec']       => String[1],
-    Optional['StartLimitBurst']             => Integer[1],
+    Optional['StartLimitBurst']             => Variant[Enum['infinity'],Integer[1]],
```
traylenator, line 53: Could you add this case to  https://github.com/voxpupuli/puppet-systemd/blob/master/spec/type_aliases/systemd_unit_unit_spec.rb as well so we don't loose it some later date.

jcpunk:
I'm showing this is currently fixed on HEAD

PR #613, opened 2026-04-09
networkd: stop all socket units to prevent socket activation restart

### LINKED ISSUE #612: networkd_ensure => stopped is not idempotent on Archlinux (systemd 260+)
## Summary

When `manage_networkd => true` and `networkd_ensure => 'stopped'`, Puppet fails the idempotency check on Archlinux rolling (systemd 260+). On every subsequent run, Puppet reports that `systemd-networkd` was changed from `running` to `stopped`, even though it was stopped by the previous run.

This has been causing the `Distro Puppet - Archlinux rolling` acceptance job to fail on every CI run since at least February 2026.

## Root Cause

Systemd 260 introduced additional socket units for `systemd-networkd`:

- `systemd-networkd-varlink.socket`
- `systemd-networkd-varlink-metrics.socket`
- `systemd-networkd-resolve-hook.socket`

These join the long-standing `systemd-networkd.socket`. All four are listed in the service's `Sockets=` directive and act as socket activation triggers. When `systemd-networkd.service` is stopped but these socket units remain active, systemd immediately re-spawns the service:

```
systemd[1]: Stopped Network Management.
systemd[1]: Starting Network Management...   ← immediate restart
systemd[1]: Started Network Management.
```

Older systemd versions (Debian 12: 252, Ubuntu 24.04: 255) only ship `systemd-networkd.socket`, which does not exhibit this behaviour.

## Steps to Reproduce

Run the following in a privileged Archlinux container with systemd as PID 1:

```bash
docker run -d --privileged --tmpfs /tmp --tmpfs /run --tmpfs /run/lock \
  -v /sys/fs/cgroup:/sys/fs/cgroup:rw --cgroupns=host \
  --name arch-test archlinux/archlinux /sbin/init

# wait for systemd to start
sleep 5

docker exec arch-test systemctl is-active systemd-networkd
# → active

docker exec arch-test systemctl stop systemd-networkd
docker exec arch-test systemctl is-active systemd-networkd
# → active  (immediately restarted)

docker exec arch-test systemctl list-units --no-pager 'systemd-networkd*.socket'
# → 4 socket units all active
```

Install the module and apply the manifest twice, observing the second run makes changes:

```bash
docker exec arch-test pacman --sync --noconfirm puppet
docker exec arch-test puppet module install puppetlabs-stdlib
# copy module into container
docker cp . arch-test:/etc/puppetlabs/code/modules/systemd
```

```puppet
class { 'systemd':
  manage_networkd  => true,
  networkd_ensure  => 'stopped',
}
```

```
# First run
Notice: /Stage[main]/Systemd::Networkd/Service[systemd-networkd]/ensure: ensure changed 'running' to 'stopped'

# Second run — should produce no changes, but:
Notice: /Stage[main]/Systemd::Networkd/Service[systemd-networkd]/ensure: ensure changed 'running' to 'stopped'
```

miharp:

On systemd 260+ (Archlinux rolling), additional socket units
(`systemd-networkd-varlink.socket`, `systemd-networkd-varlink-metrics.socket`,
`systemd-networkd-resolve-hook.socket`) cause systemd to immediately re-activate
`systemd-networkd.service` via socket activation when it is stopped, making
`networkd_ensure => stopped` non-idempotent.

Add a `systemd_networkd_socket_units` fact that discovers all
`systemd-networkd*.socket` unit files present on the node. The manifest
iterates over them, stopping all socket units before stopping the service.
This handles the current set of sockets as well as any added in future
systemd versions.

On platforms where systemd-networkd is not installed (RHEL/CentOS) the fact
returns an empty array and no socket resources are declared.

Fixes #612

### DIFF
--- a/lib/facter/systemd.rb
+++ b/lib/facter/systemd.rb
@@ -49,6 +49,28 @@ Facter.add(:systemd_version) do
   end
 end
 
+# Fact: systemd_networkd_socket_units
+#
+# Purpose:
+#   List all systemd-networkd socket units present on the system
+#
+# Resolution:
+#   Check unit files matching systemd-networkd*.socket. The set of socket units
+#   has grown over time (systemd 260+ ships additional varlink socket units);
+#   this fact returns whatever is present so the manifest can manage them all.
+#
+# Caveats:
+#
+Facter.add(:systemd_networkd_socket_units) do
+  confine systemd: true
+  setcode do
+    command_output = Facter::Util::Resolution.exec(
+      'systemctl list-unit-files --no-legend --no-pager "systemd-networkd*.socket" 2>/dev/null',
+    )
+    command_output.lines.map { |line| line.split.first }.compact
+  end
+end
+
 Facter.add(:systemd_internal_services) do
   confine systemd: true
   setcode do
--- a/manifests/networkd.pp
+++ b/manifests/networkd.pp
@@ -2,7 +2,10 @@
 #
 # @summary This class provides an abstract way to trigger systemd-networkd
 #
-# @param ensure The state that the ``networkd`` service should be in
+# @param ensure The state that the ``networkd`` service should be in.
+#   When set to ``stopped``, any ``systemd-networkd*.socket`` units present on
+#   the node are also stopped before the service to prevent socket activation
+#   from immediately restarting it.
 # @param path path where all networkd files are placed in
 # @param manage_all_network_files if enabled, all networkd files that aren't managed by puppet will be purged
 # @param link_profiles
@@ -60,6 +63,23 @@ class systemd::networkd (
     'stopped' => false,
     'running' => true,
     default   => $ensure,
+  }
+
+  # Newer systemd versions (260+) ship additional socket units for networkd
+  # (systemd-networkd-varlink.socket, systemd-networkd-varlink-metrics.socket,
+  # systemd-networkd-resolve-hook.socket). When any remain active, systemd
+  # re-activates systemd-networkd.service immediately via socket activation.
+  # Stop all present socket units before the service to prevent this.
+  # Only act when stopping — starting networkd does not require managing sockets
+  # explicitly, and doing so would override a user's deliberate socket configuration.
+  # Guard against undef in case the fact is not collected (e.g. non-systemd node).
+  if $ensure == 'stopped' and $facts['systemd_networkd_socket_units'] =~ Array {
+    $facts['systemd_networkd_socket_units'].each |$socket| {
+      service { $socket:
+        ensure => stopped,
+        before => Service['systemd-networkd'],
+      }
+    }
   }
 
   service { 'systemd-networkd':
--- a/spec/classes/networkd_spec.rb
+++ b/spec/classes/networkd_spec.rb
@@ -53,6 +53,34 @@ describe 'systemd::networkd' do
       it {
         is_expected.to contain_file('/etc/systemd/network/50-vrf.link')
       }
+
+      context 'with networkd_ensure => stopped' do
+        let(:facts) { os_facts.merge({ systemd_networkd_socket_units: ['systemd-networkd.socket'] }) }
+        let(:pre_condition) do
+          [
+            'class { "systemd": manage_networkd => true, networkd_ensure => "stopped" }',
+            'function assert_private () { }',
+          ]
+        end
+
+        it {
+          is_expected.to contain_service('systemd-networkd.socket')
+            .with_ensure('stopped')
+            .that_comes_before('Service[systemd-networkd]')
+        }
+      end
+
+      context 'with networkd_ensure => running' do
+        let(:facts) { os_facts.merge({ systemd_networkd_socket_units: ['systemd-networkd.socket'] }) }
+        let(:pre_condition) do
+          [
+            'class { "systemd": manage_networkd => true, networkd_ensure => "running" }',
+            'function assert_private () { }',
+          ]
+        end
+
+        it { is_expected.not_to contain_service('systemd-networkd.socket') }
+      end
     end
   end
 end
--- a/spec/default_module_facts.yml
+++ b/spec/default_module_facts.yml
@@ -1,4 +1,6 @@
 ---
+systemd_networkd_socket_units:
+  - systemd-networkd.socket
 systemd_internal_services:
   systemd-fsck-root.service: enabled-runtime
   systemd-journal-gatewayd.service: indirect


trefzer:
Using a fact for getting the remaining open sockets, will not work since facts computation is done asynchronous from the puppet run. A function is needed for this.

manifests/networkd.pp
```diff
@@ -62,6 +62,19 @@
...
+  # re-activates systemd-networkd.service immediately via socket activation.
+  # Stop all present socket units before the service to prevent this.
+  $facts['systemd_networkd_socket_units'].each |$socket| {
+    service { $socket:
+      ensure => $ensure,
+      enable => $_enable_networkd,
+      before => Service['systemd-networkd'],
+    }
+  }
+
```
trefzer, line 84: this will also enable all available network sockets ! this prevents users to disable only a specific network socket.

trefzer (changes requested):

miharp:
Thanks for the review @trefzer!

To clarify the intent: the fact queries \`systemctl list-unit-files\` (unit file existence on disk) rather than \`list-units\` (runtime active state). Unit files are stable — they should not appear or disappear during a puppet run under normal circumstances.

I guess this could be a problem when puppet installs \`systemd-networkd\` during the same run on a platform where it's a separate package (e.g. Debian). In that case the socket unit files wouldn't exist at fact-collection time, so no socket resources would be declared. However, on those platforms (Debian/Ubuntu with systemd < 256) there is only \`systemd-networkd.socket\` and it doesn't trigger the restart behaviour we're fixing — I think the problem is specific to systemd 260+ on Archlinux, where networkd is always pre-installed as part of the base system.

That said, if you think a function is still the right approach here to be more robust, happy to discuss. Are you thinking of using a custom Puppet function that calls \`systemctl\` at catalog application time rather than collection time?"

lib/facter/systemd.rb
```diff
@@ -49,6 +49,28 @@
...
+#   List all systemd-networkd socket units present on the system
+#
+# Resolution:
+#   Check unit files matching systemd-networkd*.socket. The set of socket units
+#   has grown over time (systemd 260+ ships additional varlink socket units);
+#   this fact returns whatever is present so the manifest can manage them all.
+#
+# Caveats:
+#
+Facter.add(:systemd_networkd_socket_units) do
```
bastelfreak, line 64: Would it help us in the future if it's a generic fact that gives us all systemd sockets, not just the networkd  ones?

miharp: Good question — I considered it, but kept it scoped to networkd to avoid shipping a large structured fact on every node for something only needed in the networkd context. If another daemon hits the same socket activation issue we could add a similar fact for it then. Happy to reconsider if you see a concrete use case.

manifests/networkd.pp
trefzer: I do not think that this is resolved.

trefzer:
Yes I definitly think that a function is a better approach than a fact.
But function is probably the wrong term, since functions are running on the puppetserver ;(. So I only see implementing it via type and provider (which I agree is not so nice).

I disagree, that it's only an Archlinux problem since Archlinux is now the first using systemd 260, but others will follow ! So implementing a generic solution should be the  goal.

(Sorry I rewrote know my whole reply, since function is on server !)

miharp:
## Local Testing

### Platform systemd versions

| Distro | systemd version | Socket units |
|--------|----------------|--------------|
| Archlinux (rolling) | 260+ | `systemd-networkd.socket`, `systemd-networkd-varlink.socket`, `systemd-networkd-varlink-metrics.socket`, `systemd-networkd-resolve-hook.socket` |
| openSUSE Tumbleweed | ~257 | — |
| Fedora 42 | 257 | — |
| CentOS Stream 10 | 257 | — |
| Ubuntu 24.04 LTS | 255 | `systemd-networkd.socket` |
| Ubuntu 24.10 | ~256 | — |
| Debian testing | ~257 | — |
| RHEL/CentOS Stream 9 | 252 | — |

Archlinux is currently the only mainstream distro shipping systemd 260+.

### Archlinux / systemd 260+ (the bug platform)

Confirmed in an Archlinux container running systemd 260:

```
$ facter systemd_networkd_socket_units
[
  "systemd-networkd-resolve-hook.socket",
  "systemd-networkd-varlink-metrics.socket",
  "systemd-networkd-varlink.socket",
  "systemd-networkd.socket"
]
```

Without the fix, stopping `systemd-networkd` and then running `networkctl status` immediately restarts the service via socket activation. With the fix (all socket units stopped first), `networkctl status` returns `Connection refused` and the service stays stopped.

### Ubuntu 24.04 agent / systemd 255 (Vagrant)

Socket units on this platform:
```
systemd-networkd.socket   disabled   enabled
```

**First puppet run** (`networkd_ensure => stopped`, service was running):
```
Notice: Catalog compiled by puppet.example.com
Notice: /Stage[main]/Systemd::Networkd/Service[systemd-networkd.socket]/ensure: ensure changed 'running' to 'stopped'
Notice: /Stage[main]/Systemd::Networkd/Service[systemd-networkd]/ensure: ensure changed 'running' to 'stopped'
Notice: Applied catalog in 0.44 seconds
```

**Second puppet run** (idempotency check):
```
Notice: Catalog compiled by puppet.example.com
Notice: Applied catalog in 0.38 seconds
```

No changes on the second run. ✓

### CentOS Stream 10 agent / systemd 252 (Vagrant)

No `systemd-networkd*.socket` unit files present — networkd is not a separate package on RHEL. The fact returns an empty array and no socket resources are declared. Catalog applies cleanly with no socket-related changes. ✓

trefzer:
Debian testing (forky) is on systemd 260.1 (https://packages.debian.org/source/forky/systemd).

The problem is not to set the sockets to 'stopped'. But if ensure is set to 'running' then all available sockets will be enabled. 
So there will be no possibility to finegrain enable and disable individual sockets (which is probably the purpose why several sockets exists !)

miharp:
@trefzer Ah, I get it now, missed the fact that `ensure => running`  implicitly makes it active :(

manifests/networkd.pp
trefzer: can you add a remark here, that a value of 'stopped', also disables all related sockets so it's clear whats happening.

trefzer (changes requested):

bastelfreak:
I want to point out that functions Always run on the server (except for referred functions) and facts on the agent.

miharp:
> I want to point out that functions Always run on the server (except for referred functions) and facts on the agent.

Yeah, I thought he might have been thinking about deferred functions which I always try and avoid :)

trefzer (approved):
looks good, thanks

miharp:
@bastelfreak @trefzer ready to merge

traylenator:
Do we know why each of the sockets does not have a 

```
PartOf=systemd-networkd.service
```
so the socket would stop on service stop would be automatic.

The behaviour of leaving the sockets behind is either deliberate or not.

Example that uses `PartOf`.

```
root@moving:~# systemctl is-active cups.socket cups.service
active
active
root@moving:~# systemctl stop cups.service
root@moving:~# systemctl is-active cups.socket cups.service
inactive
inactive
```

traylenator:
With a bit of digging I answered my own question. 

Using `PartOf` of is bad since also restarts sockets on restart of the network and that's bad.

There is `StopWhenUnneeded=yes` but a good chance it would never kick in anyway.

traylenator:
Do wonder if there is ever an actual use case to stop systemd-networkd via puppet ? I mean just document that this particular one is not tested deliberately.

traylenator:
Or this is a case of user should be aware.

If they really want to stop all the network  service it's up to them to also stop the sockets.

Add those explicit stops to the acceptance tests.

I'm not convinced trying to be smarter than what systemd is going to do is a good idea.

traylenator:
https://github.com/voxpupuli/puppet-systemd/pull/620 is the alternative.

traylenator:
Thanks for doing all the ground work here to work out which units it was - made doing  the test changes trivial.

PR #616, opened 2026-04-21
Sync up service options list with upstream

jcpunk:
I did a quick generation of the options defined in the manpages that weren't present in the module after finding a few more that I missed...


cvquesty (approved):
That was a lot of readin... I hope I got to all of it.  LGTM

PR #630, opened 2026-06-03
Add networkctl type & provider

bastelfreak:


lib/puppet/provider/networkctl/ruby.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+require 'json'
+
+# @summary custom provider to manage systemd user sessions/linger
```
kenyon, line 5: Copypaste typo about "user sessions/linger" I think 😜

PR #631, opened 2026-06-03
Add support for automount units in `manage_unit` and `manage_dropin`

kenyon:
This adds support for automount unit files in `manage_unit` and `manage_dropin`.

This also has a few commits for cleaning up typos and whitespace style in the docs.

### DIFF
--- a/spec/defines/manage_unit_spec.rb
+++ b/spec/defines/manage_unit_spec.rb
@@ -99,6 +99,45 @@ describe 'systemd::manage_unit' do
           end
         end
 
+        context 'on an automount' do
+          let(:title) { 'some-dir.automount' }
+
+          let(:params) do
+            {
+              unit_entry: {
+                Description: 'Automount /some/dir',
+              },
+              automount_entry: {
+                'Where' => '/some/dir',
+                'TimeoutIdleSec' => '5min 20s',
+              },
+              install_entry: {
+                WantedBy: 'multi-user.target',
+              },
+            }
+          end
+
+          it do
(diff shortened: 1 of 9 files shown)


spec/defines/manage_unit_spec.rb
```diff
@@ -99,6 +99,35 @@
...
+          end
+
+          it do
+            is_expected.to compile.with_all_deps
+            is_expected.to contain_systemd__unit_file('some-dir.automount')
+              .with_content(%r{^\[Automount\]$})
+              .with_content(%r{^Where=/some/dir$})
+              .with_content(%r{^TimeoutIdleSec=5min 20s$})
+              .with_content(%r{^\[Install\]$})
+              .with_content(%r{^WantedBy=multi-user\.target$})
```
smortex: Instead of looking for each-line "off-context" maybe we can have a single check of the whole file that would catch line order changes and the whole structure of the file?  Something like (not tested):

```suggestion
            is_expected.to contain_systemd__unit_file('some-dir.automount')
              .with_content(<<~CONTENT)
              [Automount]
              Where=/some/dir
              TimeoutIdleSec=5min 20s
              [Install]
              WantedBy=multi-user.target
              CONTENT
```

smortex (approved):
LGTM, only added a note about something that marginally bother me in a lot of modules 😄.

spec/defines/manage_unit_spec.rb
kenyon: Good idea, I changed it to that.

PR #632, opened 2026-06-04
lib/facter/systemd: Also include masked services

b4ldr:
update the systemd flag to also include services in the masked state


jcpunk:
I think copilot's request for tests here is a good idea.

b4ldr:
> I think copilot's request for tests here is a good idea.

i have updated to add tests

jcpunk (approved):

PR #634, opened 2026-06-16
feat: use config drop-in for journald settings

TheMeier:
use a file in /etc/systemd/journald.conf.d/ instead of modifying the main journald config file. Debian ships a /usr/lib/systemd/journald.conf.d/syslog.conf setting ForwardToSyslog=yes which can not be overridden by the main config file, so we need to use a drop-in.

### DIFF
--- a/manifests/journald.pp
+++ b/manifests/journald.pp
@@ -7,10 +7,18 @@ class systemd::journald {
   service { 'systemd-journald':
     ensure => running,
   }
+  if !empty($systemd::journald_settings) {
+    file { '/etc/systemd/journald.conf.d':
+      ensure => directory,
+      owner  => 'root',
+      group  => 'root',
+      mode   => '0755',
+    }
+  }
   $systemd::journald_settings.each |$option, $value| {
     ini_setting {
       $option:
-        path    => '/etc/systemd/journald.conf',
+        path    => '/etc/systemd/journald.conf.d/puppet.conf',
         section => 'Journal',
         setting => $option,
         notify  => Service['systemd-journald'],
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -771,23 +771,29 @@ describe 'systemd' do
             )
 
             is_expected.to have_ini_setting_resource_count(3)
+            is_expected.to contain_file('/etc/systemd/journald.conf.d').with(
+              ensure: 'directory',
+              owner: 'root',
+              group: 'root',
+              mode: '0755',
+            )
 
             expect(subject).to contain_ini_setting('Storage').with(
-              path: '/etc/systemd/journald.conf',
+              path: '/etc/systemd/journald.conf.d/puppet.conf',
               section: 'Journal',
               notify: 'Service[systemd-journald]',
               value: 'auto',
             )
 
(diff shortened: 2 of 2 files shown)


kenyon (approved):

manifests/journald.pp
```diff
@@ -7,10 +7,18 @@
...
   }
+  if !empty($systemd::journald_settings) {
+    file { '/etc/systemd/journald.conf.d':
+      ensure => directory,
+      owner  => 'root',
+      group  => 'root',
+      mode   => '0755',
+    }
+  }
   $systemd::journald_settings.each |$option, $value| {
```
jay7x, line 18: As we have our own dedicated config file, shall we use a EPP template instead of managing bunch of ini_setting maybe?

TheMeier: fine by me.. Extend https://github.com/voxpupuli/puppet-systemd/blob/master/templates/unit_file.epp or a new epp-file?

jay7x: I'd say a new one, just for this case. But I don't have a full picture in my mind. So feel free to choose the way :)

TheMeier: Well turns out there is a behavioural change I did not consider. The existing implementation allows an `ensure` paramter (Systemd::JournaldSettings::Ensure) which allowed for removing keys alltogether, which is not really possible with a drop-in. Also I am not quite sure what it did when you use ensure present:
eg.
```
     $systemd::journald_settings= {
                'MaxLevelStore' => {
                  'ensure' => 'present',
                },
```

jay7x: I'd deprecate that and switch to dumping keys to the drop-in file instead. But that's breaking change from this point of view..

TheMeier: Well. the scope keeps expanding :(
I guess the same changes should be appllied to `systemd::journal_upload_settings`
And the custom type `Systemd::JournaldSettings::Ensure` should be removed then.

Any opinions on that?

manifests/journald.pp
b4ldr, line 4: i wonder if its better to create a new define which should provide a bit more flexibility and also preserve the current behaviour? e.g. 

```puppet
class systemd::journald (
  Hash $drop_ins = {}
  Boolean $purge_drop_in_dir = false,
  Boolean $purge_journald_conf = false,
) {
  $drop_in_path = '/etc/systemd/journald.conf.d'

  if $purge_journald_conf {
    file { '/etc/systemd/journald.conf':
     ensure => file,
     content => '# managed by puppet'
    }
  }

  file { $drop_in_path:
    ensure => directory
    purge  => $purge_drop_in_dir,
  }
  $drop_ins.each |$title, $params| {
    systemd::journald::drop_in { $title:
     * => $params
    }
  }
}
define systemd::journald::drop_in (
   $file_name = $name
   $settings = {}
) {
  $full_path = "${systemd::journald::drop_in_path}/${file_name}"
  $_full_path = $full_path .stdlib::end_with('.conf').bool2str($full_path, "${full_path}.conf")
  file { $_full_path: 
    ensure => file,
   content => epp('drop_in.epp', {settings => $settings}),
  }
}

TheMeier:
I have started fresh with the feedback. Closing in favor of https://github.com/voxpupuli/puppet-systemd/pull/644

PR #635, opened 2026-06-17
Add type and wrapper for purging unit files

### LINKED ISSUE #633: Add mechanism for purging unmanaged systemd timers
When using the `cron` type, it's possible to then do...

```puppet
resources { 'cron':
  purge => true,
}
```

to remove all unmanaged crons.

It'd be desirable to have some sort of equivalent for systemd timers, (and their services?).
Maybe just a mechanism to remove unit files that were _previously_ managed by puppet.  Those created with `systemd_::manage_unit` have a `# Deployed with puppet` header that could be used to identify them.

Any thoughts before I get hacking away at something??

alexjfisher:
Add a `systemd_purge_units` type, a `systemd::purge_units` wrapper, and a
`purge_units` parameter on the base class so that removing a
`systemd::manage_unit` resource from a manifest can also remove the unit
file it deployed, without users having to declare it `ensure => absent`
first.

Only files written by `systemd::manage_unit` are eligible: the type
identifies them by the `# Deployed with puppet` header on the first line
and skips anything currently managed in the catalog, hand-written files,
symlinks, and directories. By default all unit types manage_unit can
create are considered (service, timer, path, socket, mount, automount,
swap, slice); the `unit_types` parameter narrows this.

The `systemd::purge_units` wrapper wires up a `systemd::daemon_reload`
that refreshes only when something was actually purged. Hiera users can
enable the feature via `systemd::purge_units: true`, or pass an array of
paths.

Paths are normalized before comparison so a trailing slash cannot orphan
a managed file, and a missing directory warns instead of failing the run.

The type, wrapper, base-class parameter, and class tests are my own work.
Claude (Opus 4.8) reviewed the implementation and wrote the type unit
tests and the acceptance tests.

Fixes #633

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,54 @@ systemd::manage_units:
       CPUWeight: 2000
 ```
 
+#### purging previously managed unit files
+
+After removing any `systemd::manage_unit` definitions from your manifests, by default, the unit files installed will remain on your system.
+If you wish to remove them automatically, this module can help with that.
+
+Only unit files written by `systemd::manage_unit` are eligible for purging: they are identified by the `# Deployed with puppet` header that the module writes as the first line of every unit file. Hand-written unit files and symlinks (such as systemd enablement links) are never touched.
+
+You can either...
+
+Purge all previously managed units in `/etc/systemd/system`, and reload the systemd daemon...
+
+```puppet
+include systemd::purge_units
+```
+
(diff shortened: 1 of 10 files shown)


cvquesty (approved):
LGTM

bastelfreak (approved):

PR #637, opened 2026-06-28
fix: refresh .link files via udev and manage networkd reliably

vchepkov:
.link files in /etc/systemd/network are read by systemd-udevd, not
systemd-networkd, so notifying the networkd service never re-applied
changed .link files. Route them to a new udev reload exec
(udevadm control --reload && udevadm trigger --subsystem-match=net)
instead.

Also drop the systemd_internal_services fact guard on manage_networkd
(incorporates #596): networkd is an external package managed by this
module, so the fact isn't populated until a second Puppet run, leaving
the service unmanaged on the first run.


kenyon (approved):
TIL that link files are read by systemd-udevd, not systemd-networkd 🙌

vchepkov:
It would work either way, but added ordering to reduce churn

jay7x (approved):

PR #638, opened 2026-07-07
fix: don't force default_target to running state

### LINKED ISSUE #514: Puppet fails to complete run, tries to manage multi-user target during GCP node setup
## Overview

In GCP some node types don't come with cloud-init so the only thing you can do to automate Puppet startup is to use a startup script https://cloud.google.com/compute/docs/instances/startup-scripts/linux. This startup script runs prior to the full system initialization, thus multi-user target has not been reached.

In our startup script we install the Puppet repo, the puppet-agent package, and then connect to our puppet server for the remaining setup. We're finding that Puppet is hanging for an hour during the initial puppet run, apparently attempting to manage multi-user.target (likely from the changes by @jcpunk  in #270)

```
Jan 20 18:56:01 example puppet-agent[3957]: (/Stage[main]/Stdlib::Manage/User[root]/password) changed [redacted] to [redacted]
Jan 20 18:56:01 example puppet-agent[3957]: (/Stage[main]/Stdlib::Manage/Sshkey[github.com@ssh-ed25519]/ensure) created
Jan 20 18:56:01 example puppet-agent[3957]: (/Stage[main]/Stdlib::Manage/Sshkey[github.com@ecdsa-sha2-nistp256]/ensure) created
Jan 20 18:56:01 example puppet-agent[3957]: (/Stage[main]/Stdlib::Manage/Sshkey[github.com@ssh-rsa]/ensure) created
```

...30 minutes of silence from puppet...

```
Jan 20 19:26:02 example puppet-agent[87837]: Run of Puppet configuration client already in progress; skipping  (/opt/puppetlabs/puppet/cache/state/agent_catalog_run.lock exists)
```

...almost 30 more minutes of silence from puppet...

```
Jan 20 19:51:16 example puppet-agent[3957]: (/Stage[main]/Systemd/Service[multi-user.target]/ensure) change from 'stopped' to 'running' failed: execution expired
Jan 20 19:51:16 example puppet-agent[3957]: Creating state file /opt/puppetlabs/puppet/cache/state/state.yaml
Jan 20 19:51:17 example puppet-agent[3957]: Execution of Puppet configuration client did not complete within 3600 seconds and was terminated.
```

Now I can guess where you are going, and yes we do set `systemd::default_target: multi-user.target` and yes I'll remove that.  My question is more about whether Puppet should be in charge of making the system move to this target? It doesn't seem likely to be something that Puppet can or should complete on its own.  I would think that Puppet should define the target, but not be in charge of trying to take the entire system to that target 😕 

In particular, I think that this line is the problem https://github.com/voxpupuli/puppet-systemd/blob/master/manifests/init.pp#L332 and we should consider removing it?  I find it hard to imagine a negative impact from the removal, as systemd will ensure that the node reaches the default target will it not?  Even if we do need to prompt systemd, I don't think using the service mechanism to try and start this state within the puppet convergence is the appropriate action

vchepkov:
When `default_target` was set, the class declared a
`service { $target: ensure => 'running' }` alongside the
`systemctl set-default` exec. The service resource makes Puppet run
`systemctl start <target>`, which blocks until systemd reaches that
target. On hosts where Puppet runs during early boot (e.g. GCP
instances that start before multi-user.target is reached), this hangs
the entire run until the configtimeout fires (up to an hour).

Setting the default boot target is the parameter's purpose and is
handled idempotently by the `set-default` exec; systemd reaches the
default target on its own during boot. Remove the service resource so
Puppet no longer forces a target transition mid-converge.

Fixes #514

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -420,11 +420,6 @@ class systemd (
 
   if $default_target {
     $target = stdlib::shell_escape($default_target)
-    service { $target:
-      ensure => 'running',
-      enable => true,
-    }
-
     exec { "systemctl set-default ${target}":
       command => "systemctl set-default ${target}",
       unless  => "test $(systemctl get-default) = ${target}",
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -393,7 +393,7 @@ describe 'systemd' do
 
           it {
             is_expected.to contain_exec('systemctl set-default example.target')
-            is_expected.to contain_service('example.target').with_enable(true).with_ensure('running')
+            is_expected.not_to contain_service('example.target')
           }
         end
 


vchepkov:
Errors are transient. Somebody with permissions should just re-run failed jobs

kenyon (approved):

PR #646, opened 2026-08-18
Make general purpose ".d" service configs

jcpunk:
This is a follow up to https://github.com/voxpupuli/puppet-systemd/pull/644#issuecomment-5319369826

I plan to make a follow up PR to re-org and overall sort the parameters and logic inside init.pp, but this patch was massive enough...

### DIFF
--- a/manifests/resolved.pp
+++ b/manifests/resolved.pp
@@ -118,183 +118,180 @@ class systemd::resolved (
     notify_service => true,
   }
 
-  if $dns {
-    if $dns =~ String {
-      $_dns = $dns
-    } else {
-      $_dns = join($dns, ' ')
-    }
-    ini_setting { 'dns':
-      ensure  => 'present',
-      value   => $_dns,
-      setting => 'DNS',
-      section => 'Resolve',
-      path    => '/etc/systemd/resolved.conf',
-      notify  => Service['systemd-resolved'],
-    }
-  }
-
-  if $fallback_dns {
-    if $fallback_dns =~ String {
-      $_fallback_dns = $fallback_dns
-    } else {
-      $_fallback_dns = join($fallback_dns, ' ')
-    }
-    ini_setting { 'fallback_dns':
-      ensure  => 'present',
-      value   => $_fallback_dns,
-      setting => 'FallbackDNS',
-      section => 'Resolve',
-      path    => '/etc/systemd/resolved.conf',
-      notify  => Service['systemd-resolved'],
-    }
-  }
-
-  if $domains {
-    if $domains =~ String {
-      $_domains = $domains
-    } else {
-      $_domains = join($domains, ' ')
-    }
-    ini_setting { 'domains':
-      ensure  => 'present',
-      value   => $_domains,
-      setting => 'Domains',
-      section => 'Resolve',
-      path    => '/etc/systemd/resolved.conf',
-      notify  => Service['systemd-resolved'],
-    }
-  }
-
-  $_llmnr = $llmnr ? {
-    true    => 'yes',
-    false   => 'no',
-    default => $llmnr,
-  }
-
-  if $_llmnr {
-    ini_setting { 'llmnr':
-      ensure  => 'present',
-      value   => $_llmnr,
-      setting => 'LLMNR',
-      section => 'Resolve',
-      path    => '/etc/systemd/resolved.conf',
-      notify  => Service['systemd-resolved'],
-    }
-  }
-
-  $_multicast_dns = $multicast_dns ? {
-    true    => 'yes',
-    false   => 'no',
-    default => $multicast_dns,
-  }
-
-  if $_multicast_dns {
-    ini_setting { 'multicast_dns':
-      ensure  => 'present',
-      value   => $_multicast_dns,
-      setting => 'MulticastDNS',
-      section => 'Resolve',
-      path    => '/etc/systemd/resolved.conf',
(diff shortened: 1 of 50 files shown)


lib/puppet/type/systemd_purge_units.rb
```diff
@@ -30,6 +30,7 @@
   # Similar to crayfishx/purge, this ensurable block is here to make sure _this_ resource
   # enters a changed state when there are any generated resources that have been purged.
   ensurable do
+    desc 'The basic property that the resource should be in.'
```
jcpunk, line 33: Silences a warning from REFERENCE.md

manifests/resolved.pp
```diff
@@ -118,183 +118,185 @@
...
+          owner  => 'root',
+          group  => 'root',
+          mode   => '0755',
+        }
+
+        file { '/etc/systemd/resolved.conf.d/00-disable-mdns.conf':
+          ensure => link,
+          target => '/dev/null',
+          notify => Service['systemd-resolved'],
+        }
```
b4ldr, line 213: i think the directory creation could conflict with the `ensure_resource` statment in `systemd::resolved::dropin_file`  if a user had something like the following 

```puppet
class { 'systemd':
  multicast_dns => true,
}
systemd::resolved::manage_dropin { 'foo':
    resolved_entry => {'foo' => 'bar'},
   group => 'group',
}
```

could we instead do something like the following 

```puppet
# This creates an empty file in /etc/systemd/resolved.conf.d/00-disable-mdns.conf
# This overrides the systemd-resolvd provided file in /usr/lib/systemd/resolved.conf.d/
# As both files have the same name systemd preferes the one with the higher priority in /etc
systemd::resolved::dropin_file { '00-disable-mdns.conf': }
```

b4ldr: could even put the comment as the content?

templates/config_dropin.epp
```diff
@@ -0,0 +1,8 @@
+<%- | String $section, Hash $settings | -%>
```
b4ldr:
```suggestion
<%- | String $section, Hash[String[1], Variant[String[1], Numeric] $settings | -%>
```
This should provide a hard fail instead of a silent ignore if someone passes a Hash or Array and remove the need for the if on line 5

better yet may be 
```epp
<%- | String $section, Hash[String[1], Variant[String[1], Numeric, Array[String[1]] $settings | -%>
[<%= $section %>]
<% $settings.each |$key, $value| { -%>
<%- if $value =~ Array { -%>
<%= $key %>=<%= $value.join(',') %>
<%- else -%>
<%= $key %>=<%= $value %>
<% } -%>
<% } -%>
```

b4ldr (changes requested):
LGTM but some minor comments and suggestions

manifests/resolved.pp
jcpunk: I'll extend the template to make it easier to load header comments. Work started on the existing unit file dropin

b4ldr: thanks i have approved that PR and think its a good addition regardless.  however in the above im using `systemd::resolved::dropin_file` directly so we couold just set the content e.g.

```
systemd::resolved::dropin_file { '00-disable-mdns.conf': 
  content => @(COMMENT)
    # This creates an empty file in /etc/systemd/resolved.conf.d/00-disable-mdns.conf
    # This overrides the systemd-resolvd provided file in /usr/lib/systemd/resolved.conf.d/
    # As both files have the same name systemd preferes the one with the higher priority in /etc
    | COMMENT
}

Comment (jcpunk) [reply manifests/resolved.pp:213]: Excellent, I'll clean this whole commit up and chop it into better pieces (probably next week)

Comment (jcpunk) [reply manifests/resolved.pp:213]: I think this is done now.

File: templates/config_dropin.epp (hunk 3 above)
Comment (jcpunk) [reply templates/config_dropin.epp]: Since systemd uses `` as a reset string for parameters, I've implemented a version of this, but an `Array` will provide duplicate keys so you can do things like:
```ini
[Time]
NTP=
NTP=pool.ntp.org
```

templates/config_dropin.epp
```diff
@@ -0,0 +1,23 @@
+<%- |
+  String $section,
+  Hash[String[1], Variant[String[1], Numeric, Array[String[1]]]] $settings,
+  Optional[Array[String]] $comments = undef,
+| -%>
+# This file is managed with puppet
+<% if $comments { -%>
```
b4ldr, line 15: tbh this is a bit redundant as an empty array would not loop but im not fussed if you prefer to leave it

b4ldr (approved):
lgtm

templates/config_dropin.epp
TheMeier: Maybe there is also some inspiration in https://github.com/voxpupuli/puppet-systemd/blob/master/templates/unit_file.epp#L61 regarding a very flexible template

jcpunk: The unit_file.epp is very flexible. I'm not seeing any config settings for these dropins that require space separated values.

I could be missing something, but wouldn't porting over that template logic simply make more complexity for no `$_tupled_values`?

templates/config_dropin.epp
jcpunk: This ensures a single `#` between the header and the `[section]`

jcpunk:
Any further thoughts?

PR #656, opened 2026-09-17
feat: extend support for systemd.resource-control

### LINKED ISSUE #655: Missing IPAddress directives
It is not possible to set `IPAddressAllow` and `IPAddressDeny` [unit directives](https://www.freedesktop.org/software/systemd/man/latest/systemd.directives.html#IPAddressAllow=)

d1nuc0m:

* resource-control parameters can now be used in all relevant contexts
* add support for `IPAddressAllow` and `IPAddressDeny`

Fixes #655


d1nuc0m:
Reference: https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html#IPAddressAllow=

traylenator:
How about adding it also so that it can be added to `[Socket]` that seems an obvious use case as well.

d1nuc0m:
> How about adding it also so that it can be added to `[Socket]` that seems an obvious use case as well.

Would make sense. As I don't have a lot of experience with puppet-systemd (and systemd), as the resource-controls settings are shared by "Unit configuration files for services, slices, scopes, sockets, mount points, and swap devices", is there any place where I could place them shared by the templates for these or should I add in services/sockets/etc separately?

bastelfreak (approved):

kenyon:
> > How about adding it also so that it can be added to `[Socket]` that seems an obvious use case as well.
> 
> Would make sense. As I don't have a lot of experience with puppet-systemd (and systemd), as the resource-controls settings are shared by "Unit configuration files for services, slices, scopes, sockets, mount points, and swap devices", is there any place where I could place them shared by the templates for these or should I add in services/sockets/etc separately?

As far as I can tell, there is no shared type where we could place these resource control settings, they would have to be duplicated for each of the unit types.

kenyon (approved):

traylenator:
Yes would need to be duplicate in socket.

That said putting there some where common makes sense - that's well outside the scope of this MR.

d1nuc0m:
@traylenator @kenyon @bastelfreak scope of the PR has _slightly_ expanded... Existing parameters for resource control (plus IP ones) are now available in all relevant contexts, could you have a look again?

kenyon (approved):
Looks like a good deduplication of the test code 👍

